PR #5, opened 2016-05-05
added screenshot & corrected permissions.

koolhead17:


generate-Lets-Encypt-certificate-using-Concert-for-Minio.md
```diff
@@ -34,12 +34,20 @@ total 12
...
+```
+
 After this all we need to do is to start Minio server, now you have HTTPS.
 ```
 $ minio server export/
 ```
 An example server with HTTPS is running at
 ``https://churchofminio.com:9000``
+
+![Screenshot](./assets/Letsencrypt-concert-minio.png)
```
harshavardhana: This should be 'letsencypt-concert-minio.png'

PR #13, opened 2016-07-18
Recipe for using rclone with minio server

ncw:


ncw:
Thanks :-)

GarimaKapoor:
Thank you for contibuting to the cookbook :-) takes a village

PR #61, opened 2016-08-08
Freenas jail tutorial

benagricola:
This adds a quick rundown of how to install and configure Minio inside a FreeBSD jail on FreeNAS.


GarimaKapoor:
Thank you :)

PR #72, opened 2016-08-22
docs: Fix Numberings on headings.

deekoder:
Added missing fullstops.


koolhead17:
@deekoder It was removed intentionally.

Issue #134, opened 2017-02-13
docs: Add steps to use AWS .NET SDK with Minio server


koolhead17:
We need to document for using AWS .NET SDK with Minio server

Our community member ``leonidy`` requested and suggested the working configuration.
```
new AmazonS3Client("access key", "secret key", new AmazonS3Config { ServiceURL = "http://localhost:9000", ForcePathStyle = true });
```

PR #135, opened 2017-02-13
Add steps to use AWS SDK with .NET

minherz:
Handle #134 as per request from @koolhead17


minherz:
@harshavardhana , can you check why the second commit failed? It has couple of updates and highlight for the code example that make it more readable

harshavardhana:
Nothing failed the build failure is due to Travis CI issue on our end. The PR with both changes is merged properly.

koolhead17:
Thanks @minherz 👍

PR #151, opened 2017-03-23
docs: Added proxy_buffering flag to Nginx doc.

koolhead17:


harshavardhana:
Does this work @vadmeste said it does work yet?

koolhead17:
I tested it and coordinated with @vadmeste 😸

PR #187, opened 2017-10-25
Add note on how to disable nginx buffering

nitisht:


vadmeste:
Nginx documentation makes it clear that proxy_max_temp_file_size depends on proxy_buffering.


```
Syntax: | proxy_buffering on \| off;
Enables or disables buffering of responses from the proxied server.
```

```
Syntax: | proxy_max_temp_file_size size;
When buffering of responses from the proxied server is enabled, and the whole response does not fit into the buffers set by the proxy_buffer_size and proxy_buffers directives, a part of the response can be saved to a temporary file.
```

krishnasrinivas:
Tried to reproduce the problem from nginx running (default config) on localhost proxying to play.minio.io with 200MB file, TTFB was 0.053 secs, which means it is not doing buffering for 200MB. Maybe there is a threshold after which it does not buffer. Need to reproduce the issue first.

nitisht:
Since we weren't able to reproduce the behaviour, updated the PR in line with Nginx documentation. @harshavardhana @krishnasrinivas please review

harshavardhana:
Can you update the PR looks like there are conflicts? @nitisht

nitisht:
> Can you update the PR looks like there are conflicts?

Done

nitisht:
> Can you update the PR looks like there are conflicts? @nitisht

Done

PR #190, opened 2017-11-07
Simplify the nginx doc and remove "Non Root Configuration With Rewrite"

krishnasrinivas:
Usage of the following terms are non-standard and confusing:
* Standard Root Configuration
* Non Root Configuration
* Non Root Configuration With Rewrite

Changed the terms to make it more comprehensible

Removed the section "Non Root Configuration With Rewrite" as it is not a common use-case and to keep the doc focused.

### DIFF
--- a/docs/setup-nginx-proxy-with-minio.md
+++ b/docs/setup-nginx-proxy-with-minio.md
@@ -14,7 +14,7 @@ Install Nginx from [here](http://nginx.org/en/download.html).
 
 ## 3. Configuration
 
-### Standard Root Configuration
+### Proxy all requests
 Add  below content as a file ``/etc/nginx/sites-enabled``  and also remove the existing ``default`` file in same directory.
 
 ```sh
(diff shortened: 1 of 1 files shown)


docs/setup-nginx-proxy-with-minio.md
```diff
@@ -34,43 +34,24 @@ Note:
...
    proxy_buffering off;
    proxy_set_header Host $http_host;
    proxy_pass http://localhost:9000;
  }
-```
-
-Note:
-
-* Replace `http://localhost:9000` with your own server name.
-* Replace `files` with the desired path. This cannot be `~^/minio` since `minio` is a reserved word in minio.
```
nitisht: Can we keep this point from the note

krishnasrinivas: It can be `/minio` too @nitisht for web-UI access

nitisht: okay, that makes sense

PR #201, opened 2017-12-14
Remove CloudBerry docs as its documented by CloudBerry

nitisht:
fixes https://github.com/minio/cookbook/issues/200


harshavardhana:
> I am not sure why we are removing the doc.

The reason is that CBB already documents this correctly, we do not have to duplicate the effort. Another problem to notice is that our doc is stale and not updated.  Since we are not following CBB's release cycles to keep up, it makes little sense to keep stale docs. 

Let's say if we decided to keep the doc, we should have the same quality with screenshots as mentioned here - https://www.cloudberrylab.com/download/Installation-and-Configuration-Guide-CloudBerry-Backup-with-Minio.io.pdf a much more comprehensive guide. 

IMO our doc is not useful.

deekoder:
https://docs.minio.io/docs/how-to-use-cloudberry-drive-with-minio

PR #202, opened 2017-12-17
Fix path prefix issue, because other laravel storage drivers do not d…

reshadman:
…ifferentiate between '/file.txt' and 'file.txt' as a path


harshavardhana:
> Can you provide some insight what does this PR do ? Because generally on S3 /file.txt and file.txt are indeed two different object names.

ping again @reshadman ^^

reshadman:
The given reason for this pull request is not *correct* my mistake, so I close it, 
yes, they are two different objects, but in Laravel storage drivers context there is no difference between:
`/hello.txt` and `hello.txt`, and the mentioned driver normalizes the path to prevent different behavior in different drivers. Everything works as expected.

PR #205, opened 2017-12-21
Add information about browser caching

Kiina:
Minio by itself doesn't seem to provide any expire information. If you serve images for example you would want them to get cached for at least a few days.


harshavardhana:
You need to add object metadata when uploading an object set `Cache-Control` to a relevant value.

Kiina:
Hm is this documented anywhere? I have yet to see anyone implement this in plugins which upload to minio. So maybe setting the caching information for all resources can be a viable option for many users.

harshavardhana:
> Hm is this documented anywhere? I have yet to see anyone implement this in plugins which upload to minio. So maybe setting the caching information for all resources can be a viable option for many users.

This is documented as part of S3 API https://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html 

```
Expires: 
The date and time at which the object is no longer able to be cached. For more information, go to http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.21.
```

```
Cache-Control:
Can be used to specify caching behavior along the request/reply chain. For more information, go to http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.9.
```

So IMO we don't need to talk about nginx here.

kannappanr:
`Cache-Control` is already described as part of S3 API. Hence, closing this PR as not required.

PR #215, opened 2018-05-11
Add Minio monitoring with Prometheus

nitisht:
Fixes https://github.com/minio/minio/issues/5914


harshavardhana:
It would be better to fix the doc on our server?. Are you fixing that as well ?

nitisht:
> Are you fixing that as well ?

harshavardhana:
Basically here is how I am thinking .. monitoring is an integral feature of minio and cookbook is only meant to have documentation where we are talking about minio and it's usability with other applications. Since monitoring Minio might actually have multiple Solutions not just Prometheus it makes sense to actually keep it under minio documentation in my opinion. Will wait for @deekoder opinion as well.

nitisht:
>  it makes sense to actually keep it under minio documentation in my opinion.

I agree..

deekoder:
It feels like more of a recipe for me. If you can switch to any monitoring tool/product of the user's choice, its not a core minio thing. However @harshavardhana  is correct in that the section 'minio list of metrics exposed' and 'minio config.yaml' changes should be integrated into minio docs.

deekoder:
@nitisht  can you fix the doc on the server to be generic. We can called it Minio Monitoring Guide. And then link it how to use Prometheus to monitor minio, recipe.

nitisht:
okay, will do @deekoder .

nitisht:
This PR (https://github.com/minio/minio/pull/5923) is now updated with a top level doc @deekoder , PTAL

PR #247, opened 2019-04-02
remove broken links from javascript and ror docs

kanagarajkm:
Links to javascript-shopping-app and ror-resume-uploader-app
docs are removed.


kannappanr:
If `https://github.com/minio/minio-js-store-app` is the github link, can we not point that, instead of removing it

kanagarajkm:
> If `https://github.com/minio/minio-js-store-app` is the github link, can we not point that, instead of removing it

yes, will do

kanagarajkm:
@kannappanr done, please have a look

kanagarajkm:
@ebozduman done, please have a look

PR #256, opened 2019-06-04
Fix presigned put code to upload multiple files

### LINKED ISSUE #254: only one file is uploaded when multiple are selected
Running the code from https://github.com/minio/cookbook/blob/master/docs/presigned-put-upload-via-browser.md i discovered, that only the first file is uploaded when i select multiple files in the file selector and push the upload button.

kaankabalak:
Fixes #254


kaankabalak:
> Code changes looks good.
> 
> Can you please update the Minio.Client initialization part to use `useSSL` instead of `secure` (line: 23)?

Thanks for spotting it, it seems like I have overlooked it. I will be updating the code shortly.

kaankabalak:
@kanagarajkm I have made the relevant change, would appreciate if you could review once again.

PR #263, opened 2019-06-30
Update Presigned URL example to ditch jQuery (broken) in favor of ES6 (working)

Lewiscowles1986:
The jQuery example told me `$` was undefined

I switched this around so I could run locally against a golang backend to verify Minio. This is it being switched back, using modern JS ES6 with newer browser API's

## Browser compatibility

- [Promise API compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility)
- [querySelector API compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector#Browser_compatibility)
- [Fetch API compatibility](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility)
- [File API compatibility](https://developer.mozilla.org/en-US/docs/Web/API/File#Browser_compatibility)

PR #264, opened 2019-07-02
 Presigned PUT: Update Chinese example and fix English comment

kaankabalak:
The Chinese version of the doc has been updated based on the changes
made on #263.


kaankabalak:
> Line:43-45 should be updated to remove JQuery and replace XMLHttpRequest with Fetch.

Thanks for the review @kanagarajkm, seems like I have overlooked it initially. I have updated the PR, would appreciate if you could take another look.

kaankabalak:
The PR is now ready for review, would appreciate if you could take a look once again.

PR #267, opened 2019-07-15
Syntax improvement, Comment optional nginx module

Lewiscowles1986:
When I found this after having some difficulties proxying minio in a local environment; I found it did not work.

I've many years server-admin, so I knew where to look for errors and could spot points for change. I know people this wouldn't be true for, who would benefit from using minio in this setup.

- Comments out `healthcheck` and points at documentation from vendor
- adds missing semi-colon
- adds note about backend hostname in docker[-compose] environments
- small addendum about the location directive

### DIFF
--- a/docs/setup-nginx-proxy-with-minio.md
+++ b/docs/setup-nginx-proxy-with-minio.md
@@ -30,14 +30,15 @@ server {
  proxy_buffering off;
 
  location / {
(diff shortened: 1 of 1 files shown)


Lewiscowles1986:
Heh, almost forgot about this. Thanks for merging.

PR #294, opened 2020-09-04
Versioning Initial Pass

ravindk89:
@poornas @vadmeste  This is a portion of the Versioning work. Specifically:

- Top-level page on Bucket Versioning
- `mc ls` for the purpose of defining some core versioning terms/behavior
- Some random other fixes that aren't a priority.

If I have the functionality of the top level + `mc ls versions` and `mc ls rewind` correct, I will apply those throughout to all commands that support those fields. I will have a patch for copy related to `--version-id` a bit later.

### DIFF
--- a/source/includes/facts-versioning.rst
+++ b/source/includes/facts-versioning.rst
@@ -1,0 +1,36 @@
+.. start-rewind-desc
+
+Directs |command| to operate only on the object version(s) that existed at
+specified point-in-time.
+
+- To rewind to a specific date in the past, specify the date as an
+  ISO8601-formatted timestamp. For example: ``--rewind "2020.03.24T10:00"``.
+
+- To rewind a duration in time, specify the duration as a string in
+  ``#d#hh#mm#ss`` format. For example: ``--rewind "1d2hh3mm4ss"``.
+
+|rewind| requires that the specified |alias| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-rewind-desc
+
+.. start-versions-desc
+
+Directs |command| to operate on all object versions that exist in the bucket.
+
+|versions| requires that the specified |alias| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-versions-desc
+
+.. start-version-id-desc
+
+Directs |command| to operate only on the specified object version.
+
+|versionid| requires that the specified |alias| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-version-id-desc
--- a/source/includes/facts-versioning.rst
+++ b/source/includes/facts-versioning.rst
@@ -34,3 +34,98 @@ that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
 :mc-cmd:`mc version` to enable or disable bucket versioning.
 
 .. end-version-id-desc
+
+..
+
+   So this is an ugly workaround. Since you can't override substitutions
+   on a single page, those reference commands with multiple subcommands
+   get kind of messy. Instead, the following sections "duplicate"
+   the core content for supporting commands with multiple subcommands that
+   support versioning arguments.
+
+   The ideal path forward would be to extend the include directive to
+   allow for per-directive replacement, but that will take significant
+   engineering effort. So until then, kludges it is.
+
+..
+   ----------------- 2nd Argument --------------------
+
+.. start-rewind-desc-2
+
+Directs |command-2| to operate only on the object version(s) that existed at
+specified point-in-time.
+
+- To rewind to a specific date in the past, specify the date as an
+  ISO8601-formatted timestamp. For example: ``--rewind "2020.03.24T10:00"``.
+
+- To rewind a duration in time, specify the duration as a string in
+  ``#d#hh#mm#ss`` format. For example: ``--rewind "1d2hh3mm4ss"``.
+
+|rewind-2| requires that the specified |alias-2| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-rewind-desc-2
+
+.. start-versions-desc-2
+
+Directs |command-2| to operate on all object versions that exist in the bucket.
+
+|versions-2| requires that the specified |alias-2| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-versions-desc-2
+
+.. start-version-id-desc-2
+
+Directs |command-2| to operate only on the specified object version.
+
+|versionid-2| requires that the specified |alias-2| be an S3-compatible service
+that supports :ref:`minio-bucket-versioning`. For MinIO deployments, use
+:mc-cmd:`mc version` to enable or disable bucket versioning.
+
+.. end-version-id-desc-2
+
+
+
+..
+
+   -------------- 3rd Subcommand --------------
+
+.. start-rewind-desc-3
+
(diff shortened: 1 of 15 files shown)


source/minio-features/bucket-versioning.rst
```diff
@@ -0,0 +1,87 @@
...
+
+<Diagram>
+
+MinIO generates a unique immutable ID for each object. If a ``PUT`` request
+contains an object name which duplicates an existing object, MinIO does *not*
+overwrite the "older" object. Instead, MinIO retains all object versions while
+considering the most recently written "version" of the object as "latest".
+Applications retrieve the latest object version by default, but *may* retrieve
+any other version in the history of that object. To view all versions of an
+object or objects in a bucket, use the :mc-cmd-option:`mc ls versions` command.
```
poornas, line 27:
```suggestion
object or objects in a bucket, use the :mc-cmd-option:`mc ls --versions` command.
```

source/minio-features/bucket-versioning.rst
```diff
@@ -0,0 +1,87 @@
...
+
+By default, deleting an object does *not* remove all existing versions of
+that object. Instead, MinIO places a "delete" marker for the object, such that
+applications requesting only the latest object versions do not see the object.
+Applications *may* retrieve earlier versions of that object. To completely
+delete an object and its entire version history, use the
+:mc-cmd-option:`mc rm versions` command.
```
poornas, line 34:
```suggestion
:mc-cmd-option:`mc rm --versions` command. 
```

source/minio-features/bucket-versioning.rst
```diff
@@ -0,0 +1,87 @@
...
+that object. Instead, MinIO places a "delete" marker for the object, such that
+applications requesting only the latest object versions do not see the object.
+Applications *may* retrieve earlier versions of that object. To completely
+delete an object and its entire version history, use the
+:mc-cmd-option:`mc rm versions` command. 
+
+Enable Bucket Versioning
+------------------------
+
+Enabling bucket versioning on a MinIO deployemnt requires that the deployment
```
poornas, line 39:
```suggestion
Enabling bucket versioning on a MinIO deployment requires that the deployment
```

source/minio-features/bucket-versioning.rst
```diff
@@ -0,0 +1,87 @@
...
+contacting a global registration authority. 
+
+MinIO object version UUID's are *immutable* after creation. 
+
+Versioning Dependent Features
+-----------------------------
+
+The following MinIO features *require* bucket versioning for functionality:
+
+- Object Locking (:mc-cmd:`mc lock` and :mc-cmd-option:`mc mb with-lock`)
```
poornas, line 85:
```suggestion
- Object Locking (:mc-cmd:`mc lock` and :mc-cmd-option:`mc mb --with-lock`)
```

poornas (changes requested):

source/minio-features/bucket-versioning.rst
ravindk89: So the `:mc-cmd-option:` directive has a bit of odd behavior. It splits by whitespace and adds `--` as a suffix to the last argument (in this case, `versions`). 

This is mostly necessary due to some quirks with the way Sphinx parses whitespaces into anchorable links (it uses `-` to replace ` ` whitespace characters, so you end up with broken option links. 

I might be able to fix this later with optimizations to the underlying Python code, but that's for later.

This is all to say, this should work as designed - that is, in the rendered output, this shows as `mc ls --versions`. 

You can build locally to check, the build process is relatively straightforward (create a virtual environment, install from requirements, run make html)

source/minio-features/bucket-versioning.rst
ravindk89: See above. Works as designed.

source/minio-features/bucket-versioning.rst
ravindk89: See above. works as designed.

source/includes/facts-versioning.rst
```diff
@@ -0,0 +1,36 @@
+.. start-rewind-desc
```
ravindk89, line 1: @poornas @vadmeste @kannappanr 

So I tried to refactor a bit, since these options are used quite widely. 

The following file is *referenced* by the various commands that use these options. Specifically, `.. start-<option>-desc` indicates the start of the section and
`.. end-<option>-desc` indicates the end of the section. Sphinx grabs only whats in the middle and then performs additional transformations later.

`|<string>|` gets replaced in the inheriting document. It should, based on my testing, always produce appropriate results.

The descriptions are a little more generic in this form, but its much easier to manage changes here.

If we do end up with specific commands that have more unique behavior, I can expand this file to include

source/minio-cli/minio-mc/mc-rm.rst
```diff
@@ -23,6 +23,12 @@ To completely remove a bucket, use :mc:`mc rb` instead.
 Syntax
 ------
 
+.. |command| replace:: :mc-cmd:`mc rm`
```
ravindk89, line 26: This is an example of a page that uses all of the version-specific options + replacements.

Each `.. |<string>|` is replaced by the string following `replace::`

ravindk89 (commented):
@poornas @vadmeste @kannappanr 

I've made some changes to the structure that should make review much easier.

If you are having trouble building locally to validate the output, let me know and I'll create an archive of the build artifact for easier review.

source/minio-cli/minio-mc/mc-ls.rst
```diff
@@ -0,0 +1,125 @@
...
+   Lists the contents of the ``mybucket`` bucket. ``play`` corresponds to the
+   :mc-cmd:`alias <mc alias>` of a configured S3-compatible service.
+
+:mc-cmd:`mc ls --recursive play <mc ls recursive>`
+   Recursively lists all buckets and objects on the S3-compatible service.
+   ``play`` corresponds to the :mc-cmd:`alias <mc alias>` of a configured
+   S3-compatible service.
+
+:mc-cmd:`mc ls --versions play/myversionedbucket <mc ls versions>`
+   Lists the version of all objects in the ``myversionbucket`` bucket. ``play``
```
poornas, line 37:
```suggestion
   Lists the version of all objects in the ``myversionedbucket`` bucket. ``play``
```

source/minio-cli/minio-mc/mc-ls.rst
```diff
@@ -0,0 +1,125 @@
...
+   <minio-bucket-versioning>`. Use :mc:`mc versions` to enable versioning
+   on a bucket.
+
+:mc-cmd:`mc ls --rewind 7d play/myversionedbucket <mc ls versions>`
+   Lists the contents of the ``myversionedbucket`` bucket as they
+   existed 7 days prior to the current date. ``play`` corresponds to the
+   :mc-cmd:`alias <mc alias>` of a configured S3-compatible service.
+
+   :mc-cmd-option:`mc ls versions` requires :ref:`bucket versioning
+   <minio-bucket-versioning>`. Use :mc:`mc versions` to enable versioning
```
poornas, line 51:
```suggestion
   <minio-bucket-versioning>`. Use :mc:`mc version` to enable versioning
```

poornas (changes requested):

poornas:
built the docs locally and reviewed the versioning part - looks good to me.

source/minio-cli/minio-mc/mc-legalhold.rst
```diff
@@ -30,60 +30,207 @@ mode on the object.
    See :mc-cmd-option:`mc mb with-lock` for documentation on creating
    buckets with object locking enabled. 
 
+Quick Reference
```
ravindk89, line 33: @poornas @vadmeste mc legalhold was out of date (on my local), so I've updated + added versioning. Let me know if this looks right.

source/minio-cli/minio-mc/mc-tag.rst
```diff
@@ -0,0 +1,223 @@
+==========
```
ravindk89, line 1: @poornas @vadmeste added mc tag as I was missing it before, along with versioning functionality.

ravindk89 (commented):
@poornas @vadmeste This should be the last round of review. Specifically, I updated `mc legalhold` and added `mc tag` to the list of versioned commands.

If this looks good I'll merge and continue watching for other commands that support versioning. There's some ugly kludge-stuff with single-sourcing that should have no effect on the final product.

poornas (approved):
LGTM

PR #295, opened 2020-09-04
auth/authz review

ravindk89:
Hey @eco-minio , also tagging @aead as I think you're both security-minded.

This is a fake-ish PR for some of the work I did. `minio-authentication-authorization-review.rst` is a copy of an existing file, because I can't find a trick to make git think a file is new otherwise. 

The goal here is to document all of the various *things* you can set in a policy document, along with having some examples of usage. Ideally a user does *not* have to visit AWS docs to be able to create custom policies for use with MinIO. 

LMK if you have any questions.

### DIFF
--- a/source/security/minio-authentication-authorization-review.rst
+++ b/source/security/minio-authentication-authorization-review.rst
@@ -13,8 +13,8 @@ Overview
 
 *Authentication* is the process of verifying the identity of a connecting
 client. MinIO authentication requires providing user credentials in the form of
-an access key and corresponding secret key. The MinIO deployment only grants
-access *if*:
+an access key (username) and corresponding secret key (password). The MinIO
+deployment only grants access *if*:
 
 - The access key corresponds to a user on the deployment, *and*
 - The secret key corresponds to the specified access key.
@@ -36,15 +36,6 @@ referenced in a user's assigned or inherited policies.
 
 - For more information on MinIO policy creation, see
   :ref:`minio-auth-authz-pbac-policies`.
-
-MinIO also supports Secure Token Service (STS) authentication, where clients use
-a supported OIDC-compliant third-party identity provider to perform
-authentication. MinIO uses the security token returned by the provider to
-generate a random access key and secret key that the client can use for
-authenticating to the deployment. The username specified to the third-party
-identity provider *must* match an existing user on the MinIO deployment. MinIO
-uses the policies associated to that existing user for authorizing client
-operations. For more information, see :ref:`minio-sts-overview`.
 
 .. _minio-auth-authz-users:
 
@@ -569,8 +560,8 @@ The following table lists the MinIO-supported policy action keys.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 MinIO supports the following actions for use with defining policies
-for :mc-cmd:`mc admin` operations. These actions are **only** valid for
-MinIO deployments and are **not** intended for use with other S3-compatible
+for :mc-cmd:`mc admin` operations. These actions are *only* valid for
+MinIO deployments and are *not* intended for use with other S3-compatible
 services:
 
 .. policy-action:: admin:*
--- a/source/security/minio-security-security-token-service.rst
+++ b/source/security/minio-security-security-token-service.rst
@@ -1,3 +1,5 @@
+.. _minio-sts-overview:
+
 ============================
 MinIO Security Token Service
 ============================
--- a/source/security/minio-security.rst
+++ b/source/security/minio-security.rst
@@ -41,7 +41,7 @@ MinIO provides support for the following security features:
    :titlesonly:
    :hidden:
 
-   /security/minio-authentication-authorization
+   /security/minio-authentication-authorization-review
    /security/minio-security-TLS-encryption
    /security/minio-security-server-side-encryption
    /security/minio-security-security-token-service 


source/security/minio-authentication-authorization-review.rst
```diff
@@ -0,0 +1,836 @@
...
+.. contents:: On This Page
+   :local:
+   :depth: 2
+
+Overview
+--------
+
+*Authentication* is the process of verifying the identity of a connecting
+client. MinIO authentication requires providing user credentials in the form of
+an access key (username) and corresponding secret key (password). The MinIO
```
aead, line 16: Not really sure whether we should use the username / password analogy here. 
An S3 access/secret key pair is **explicitly** not a username /password - due to the S3 authentication method of computing the request authorization checksum (the value AWS S3 calls "signature"). S3 uses:
 ```
checksum := HMAC_SHA-256(serect_key, ...)
```
If an adversary observes one request checksum it can mount a (relatively cheap) brute-fore attack to recover the secret key - unless the secret key is chosen as random value of a sufficiently large domain: e.g. 256 bit random string.
HMAC-SHA256 is a function that is reasonably cheap / fast to compute and is not memory-hard. (It's a KDF, not a PBKDF).
Therefore, the access/secret key pair should be chosen at random -  not by a user as username/password.

Comment (ravindk89) [reply source/security/minio-authentication-authorization-review.rst:16]: That's fair - I used the analogy because most users should be familiar with the general idea behind it (assuming they've never approached auth/authz before outside of normal user interactions w/ apps that require it). 

It also occurs to me that this implies we *only* support that kind of auth, where we do have a number of other methods (STS, JWT Web Identity, etc.) - though I guess those *also* boil down to auth + secret key?

Comment (aead) [reply source/security/minio-authentication-authorization-review.rst:16]: Yeah, so STS or `AssumeRole*` are basically mechanisms to exchange a e.g. STS token against a (temp). S3 access/secret key pair. The S3 protocol only knows about access/secret key pairs. Any other authentication method basically acts as adapter to connect systems that use different concepts (e.g. STS).

File: source/security/minio-authentication-authorization-review.rst [hunk 2]
```diff
@@ -0,0 +1,836 @@
...
+   <API_GetObjectVersionForReplication.html>` IAM action.
+
+
+.. _minio-auth-authz-pbac-mc-admin-actions:
+
+``mc admin`` Policy Action Keys
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO supports the following actions for use with defining policies
+for :mc-cmd:`mc admin` operations. These actions are *only* valid for
```
eco-minio, line 563: might want to bold "only" here (or highlight some other way) italics only with this font is easy to miss

eco-minio (commented):
Few style comments, non crtitical

source/security/minio-authentication-authorization-review.rst
```diff
@@ -0,0 +1,836 @@
...
+   * - ``Admin``
+     - :policy-action:`admin:*`
+     - All :mc-cmd:`mc admin` commands.
+
+Users also inherit permissions from their assigned :ref:`groups
+<minio-auth-authz-groups>`. A user's total set of permissions consists of their
+explicitly assigned permissions *and* the inherited permissions from each of
+their assigned groups.
+
+.. admonition:: ``Deny`` overrides ``Allow``
```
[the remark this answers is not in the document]
ravindk89: I'm actively working on cleaning up the CSS, so this should (hopefully) not be a long-term problem :)

source/security/minio-authentication-authorization-review.rst
ravindk89: Done

source/security/minio-authentication-authorization-review.rst
ravindk89: Fixed

ravindk89:
@aead @eco-minio if this otherwise looks good I'll merge it up. Thanks again!

aead (approved):

source/security/minio-authentication-authorization.rst
```diff
@@ -37,6 +37,19 @@ referenced in a user's assigned or inherited policies.
 - For more information on MinIO policy creation, see
   :ref:`minio-auth-authz-pbac-policies`.
 
+MinIO also supports using external providers for authentication using
+the Security Token Service (STS). Clients use the external provider to
+request temporary credentials for accessing the MinIO deployment. STS 
+requires configuring each external provider as part of ``minio`` server
+startup. 
+
+With STS, clients use a supported third-party identity provider to perform
```
eco-minio, line 46:
```suggestion
With STS, clients use a supported OIDC compliant third-party identity provider to perform
```

source/security/minio-authentication-authorization.rst
```diff
@@ -560,8 +573,8 @@ The following table lists the MinIO-supported policy action keys.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 MinIO supports the following actions for use with defining policies
-for :mc-cmd:`mc admin` operations. These actions are *only* valid for
-MinIO deployments and are *not* intended for use with other S3-compatible
+for :mc-cmd:`mc admin` operations. These actions are **only** valid for
```
eco-minio, line 576: What is mc-cmd here?  Is this referencing an alias to mc or is it a formatting element?

eco-minio (changes requested):

source/security/minio-authentication-authorization.rst
ravindk89: Yep - in Sphinx, `:<string>:` is a directive, which is equivalent to a function. In this case, the `:mc-cmd:` directive searches through an index of `mc-cmd` objects for one that matches the string `mc admin` and links to the location of that object.

TL;DR if you click the `mc admin` text, it links to the page for `mc admin`. No hardlinking required.

source/security/minio-authentication-authorization.rst
ravindk89: Done

ravindk89:
@eco-minio implemented your changes with some additional refinements. LMK what you think.

eco-minio (approved):
LGTM

PR #298, opened 2020-10-09
Restructure Security Docs

ravindk89:
Restructuring Security sections to better distinguish between access controls/identity and encryption-related topics. 

TBD: 

- Split User/Groups and Policies into distinct documents
- Import KES-related docs under Encryption


source/IAM/identity-access-management.rst
```diff
@@ -1,8 +1,8 @@
 .. _minio-auth-authz-overview:
 
-================================
-Authentication and Authorization
-================================
+==============================
+Identity and Access Management
+==============================
 
 .. default-domain:: minio
```
ravindk89, line 7: @aead am now thinking of breaking this page up into two distinct child pages:

Identity Management (Covers Users + Groups)
Access Management (Covers Policies)

LMK your thoughts.

aead:
@ravindk89 My naive assumption would be a structure like:
```
 - IAM
    - Users
    - Groups
    - Policies
    - Providers
    - ....
 - Encryption & Key Management
    - Client-Side Encryption
    - Server-Side Encryption
    - KES + KMS
    - TLS + Certificates
    - ... 
```
All that could be under a section `Security` - (that would be 3 levels).
The names are just what I made up and may be adjusted to whatever fits.
Is that kind of what you have in mind as well - or is it totally different?

Also, is the idea make these sections informational concept pages explaining what is possible (and best practice)? Or should this be guides that walk you through how to configure / setup certain things?

B/c I don't think that is sufficient to just provide a `N` step guide w/o a (separate) explanation page. From my experience, user `A` wants a guide that shows how to do a specific task - e.g. setup the KMS or configure LDAP. However, there is also user `B` who wants to first understand what is possible and how the model looks like - e.g. What is the difference between client-side & server-side encryption? Is server-side encryption the same as "encryption in-flight" (TLS)? What SSE should I choose - SSE-C or SSE-S3 / SSE-KMS. What KMS vendors can I use? a.s.o.

ravindk89:
> The names are just what I made up and may be adjusted to whatever fits.
Is that kind of what you have in mind as well - or is it totally different?

These are actually fine - all that is necessary is for the names to be generally predictive of the content they contain. 

> Also, is the idea make these sections informational concept pages explaining what is possible (and best practice)? Or should this be guides that walk you through how to configure / setup certain things?

Right now I'm conceptualizing two possible structures.

- Concepts where simple guides suffice, or we can point to a reference page. User/Group creation pages might be like this. A few quick start examples + a link to `mc admin user` and `mc admin group`. 

- Concepts where you need distinct guides due to complexity. KES is an easy example - the page we discuss what KES is and does would get very long and crowded if we also included a configuration guide for each supported KMS (maybe). 

My hope is to lead with concepts and nest procedures, such that User A can always skip the concepts easily. The dual nav structure should help where we have very long pages, as someone can just click to the exact position they want. 

With the proposed structure, for example, the KES + KMS page might be "long" on guides (one for each supported KMS) - but with the right-hand in-page floating toc, its very easy to click to the one you care about. 

The only advantage of splitting IAM and EKM into first-level nav is we don't have to mess with three levels of left-hand navigation, which can get a little ugly - I'm less concerned about this in the short term as its an easy problem to solve if we get users who find `Security` to be too heavy. I guess we will see. 

I'm going to patch up using the suggested organization and add some content that provides more specific examples of both "Concept + Instruction" and "Instruction only" pages. That might help.

ravindk89:
Per discussions w/ @aead , going to patch the following skeleton:

``` - IAM
    - Users
    - Groups
    - Policies
    - Providers
    - ....
 - Encryption & Key Management
    - Client-Side Encryption (Concept + Guide)
    - Server-Side Encryption (SSE-C, SSE-S3)
    - MinIO Key Encryption Service (reference)
    - SSE with Thales (Guide)
    - SSE with AWS KMS (Guide)
    - SSE with etc. (Guide)
    - TLS + Certificates (Concept + Guide)
    - ...
```

Nest under `Security` L1 header. This should keep findability w/o search, but will require diligent cross-referencing so users understand what features require KES, KMS, etc.

ravindk89:
@aead updated with skeleton. I've done a bit of content filling, but not too much. This might be easier to go over via Zoom as the changes are significant on some pages.

I've also left a 'skeleton' of a procedure in `source/security/encryption/sse-s3-thales` . It's rough but it should provide an idea of a 'Procedure' that implements a very simple view of 'Concept' (so users have one last chance to learn before they dive in).

source/security/IAM/iam-providers.rst
```diff
@@ -0,0 +1,11 @@
...
+Providers
+=========
+
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 1
+
+Stub - might duplicate STS page?
```
aead, line 11: I guess STS should be moved here. It's part of user authorization. Eventually we need to explain what STS is (purpose, workflow etc.) However, not necessarily in this PR.

source/security/encryption/minio-kes.rst
```diff
@@ -0,0 +1,82 @@
...
+   :class: copyable
+
+   kes server --cert CERTIFICATE --key PRIVATEKEY --root ROOT_IDENTITY [OPTIONAL_FLAGS]
+
+:mc:`kes server` supports the following arguments:
+
+.. mc-cmd:: cert
+   :option:
+
+   *Required*
```
aead, line 51: KES does not have mandatory CLI flags. All are optional - and only required if the corresponding field in the config file is omitted.

aead (approved):

source/security/encryption/minio-kes.rst
ravindk89: Done

source/security/IAM/iam-providers.rst
ravindk89: Hm - yeah lets do that in another PR. We can kind of mention both in kind (External Identity Providers supported via STS) or something).

PR #315, opened 2020-12-10
Change ServerSets to Pools

ravindk89:
See https://github.com/minio/operator/pull/371 for related upstream work

@nitisht @Alevsk at your leisure can you just give this a quick headcheck? I'm pretty sure its a simple find/replace, but I could have missed something else that was done upstream that would effect the CRD.


Alevsk (approved):
LGTM

PR #319, opened 2021-01-13
DOCS-317: UI/UX fixups

ravindk89:
So this is probably *the* hackiest thing i've done HTML wise. But TL:DR:

- Cloned minio-ghost-theme and modified app.scss to include as few items as possible
- Modified the rendered css to not stomp my existing CSS wherever possible
- literally copied the HTML from the blog site for the header

Mostly I just need a headcheck to ensure I haven't done something *very very horrible*. In absence of any serious sins of HTML/CSS, I'll merge this EOW.

You can render and test yourself using the following steps:

# Build Instructions

Clone, fetch, and checkout DOCS-309 branch. 

```py
python -m pip install --user virtualenv
python3 -m venv venv

source venv/bin/activate
pip install -r requirements.txt
make html
python -m http.server --directory build/DOCS-309/html
```

# Notes

In the future, it would be nice to fully transition the CSS to SCSS from a central repo, but that will be a very big project.

Also, Sphinx supports JINJA templating natively, and i cannot make heads or tails of how to extend the templating hook to include support for handlebars. There's probably a way to make it work, but that's out of scope for this part of the project.


ravindk89:
Disregard this pull. Priority is to integrating [Rushan's work](https://github.com/rushenn/docs/commit/5646d48a14b7dee52589bd3839aad81fd2cd7da9) instead. This is for reference as part of the merge.

PR #320, opened 2021-01-13
DOCS-317 Redux

ravindk89:
@rushenn so this is the result of:

1) cherry-picking your commit over, and
2) Making some additional tweaks 

The only thing missing I can think of is I can't figure out how to get the collapse button to render. But that is *very* low priority, and in fact we can chat w/ design to find a nice way of adding a collapse/expand button.

Otherwise, this looks pretty good to me. Please review and let me know if we need to move some of the .scss definitions around. 

# Build Instructions

```py
python -m pip install --user virtualenv
python3 -m venv venv

source venv/bin/activate
pip install -r requirements.txt
make html
python -m http.server --directory build/DOCS-309/html
```


source/_static/scss/includes/_aside.scss
```diff
@@ -20,6 +20,16 @@
     }
 }
 
+.index-link {
```
ravindk89, line 23: Adding back the index link as per discussion, so users can easily find the index + know which docs property they are on.

source/_static/scss/includes/_aside.scss
```diff
@@ -86,7 +97,7 @@
             &.current {
                 & > a {
                     color: $docs-item-main-active-color;
-                    font-weight: $font-wight-medium;
+                    font-weight: $font-weight-medium;
```
ravindk89, line 100: minor typo fix.

source/_static/scss/includes/_base.scss
```diff
@@ -20,16 +20,91 @@ body {
...
+        border: none;
+        border-left: 4px solid $docs-item-admonition-note-border;
+    }
+}
+
+dl.minio {
+    margin: 10px 0 10px 0;
+}
+
+div.footer {
```
ravindk89, line 75: Temporary, the footer is minimally important and this just gets it onto the page.

source/_static/scss/includes/_base.scss
```diff
@@ -20,16 +20,91 @@ body {
...
+}
+
+.align-default {
+    text-align: left;
+}
+
+table.docutils {
+    border: none;
+    box-shadow: none;
+    -webkit-box-shadow: none;
```
ravindk89, line 87: Not sure if this belongs in _normalize or not. alabaster theme includes some CSS I would love to be rid of, but that's a bigger project.

source/_static/scss/includes/_layout.scss
```diff
@@ -31,4 +31,5 @@
     flex-shrink: 0;
     position: sticky;
     top: 0;
+    max-width: 200px;
```
ravindk89, line 34: Given a 1366 resolution, I'd like to keep at least 600 for the main body, so restricting this should guarantee that (?).

rushenn (approved):
LGTM.

PR #321, opened 2021-01-22
Fix sidebars and add responsive support

rushenn:


ravindk89:
Hey @rushenn , I did a local build and the top nav isn't rendering. Is there something specific required other than a clean build here, i.e. re-running gulp?

ravindk89:
Ah - I get a 404 on `_static/data/nav.json` when loading up using `python -m http.server`

ravindk89:
So I realised as part of this, it helps to test the build using something like `python -m http.server --directory build/<branch>/html`, which renders things as expected. 

Looking through the changes, everything here looks good.

ravindk89 (approved):

PR #328, opened 2021-03-19
Bucket Notification Import

ravindk89:
# Summary

This is a full import of all existing Bucket Notifications. There are a number of significant departures from the existing docs:

1) Each notification target gets its own dedicated page + summary of envvars and config values. Any specific support requirements are noted - in some cases there are conflicts between what we originally stated and what the underlying library states, and I will comment on those sections for visibility.

2) I've removed the sample codes for "testing" - many of them have not been updated since the original targets were added. I would prefer we not have examples of how to use third-party applications until we have the resources to update them at any given point in time. 

3) I have tried to parse through the existing code as much as possible to figure out if we are missing any envvars or config settings. There are a few that popped up - there are also a few settings I could not figure out in terms of what the user is actually expected to provide. I will note those in comments.

# Quirky Things That Might Make No Sense At First

One big thing with this PR is the `common-mc-admin-config` file. This is a single file in which all common text is shared between envvars and config files. This makes it easier (for me) to update both envvars and configs at the same time by editing within a single file. It is also an absolutely humongous file because of that. I would strongly recommend focusing on the tutorial pages and just leaving comments on the `.. include::` sections corresponding to the envvar or config for which you have a concern. That will be easier, probably.

# Review Goals

To be honest, the main thing here is just some extra eyes to catch obvious mistakes I've made, and to answer questions I've left open. Beyond that I do not need a deep-dive.

# Build Instructions

5) Run `make html`

## Key URLs

- http://127.0.0.1:8000/monitoring/bucket-notifications/bucket-notifications.html
- http://127.0.0.1:8000/reference/minio-server/minio-server.html#mqtt-bucket-notification-target
- http://127.0.0.1:8000/reference/minio-cli/minio-mc-admin/mc-admin.config.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-amqp.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-mqtt.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-nats.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-nsq.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-elasticsearch.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-kafka.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-mysql.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-postgresql.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-redis.html
- http://127.0.0.1:8000/monitoring/bucket-notifications/publish-events-to-webhook`.html

This is a *lot*.


source/includes/common-mc-admin-config.rst
```diff
@@ -0,0 +1,978 @@
...
+MinIO supports the following mechanisms:
+
+- ``PLAIN`` (Default)
+- ``SCRAM``
+
+.. end-minio-notify-kafka-sasl-mechanism
+
+.. start-minio-notify-kafka-tls-client-auth
+
+Specify the client authentication policy of the Kafka broker(s).
```
ravindk89, line 844: @poornas @vadmeste I could not find anything on what the expectation for this field actually is, nor did the kafka or go library docs shed much light here either.

source/includes/common-mc-admin-config.rst
```diff
@@ -0,0 +1,978 @@
...
+.. start-minio-notify-kafka-client-tls-key
+
+Specify the path to the client private key to use for performing
+mTLS authentication to the Kafka broker(s).
+
+.. end-minio-notify-kafka-client-tls-key
+
+.. start-minio-notify-kafka-version
+
+Specify the version of the Kafka cluster.
```
ravindk89, line 888: @vadmeste @poornas This field puzzled me - maybe it makes sense to Kafka users, but why do we need to specify the version of the Kafka cluster as part of the request?

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
```diff
@@ -0,0 +1,377 @@
...
+   :depth: 1
+
+MinIO supports publishing :ref:`bucket notification
+<minio-bucket-notifications>` events to Kafka. 
+
+MinIO relies on the :github:`Shopify/sarama` project for Kafka connectivity
+and shares that project's Kafka support. See the 
+``sarama`` :github:`Compatibility and API stability 
+<Shopify/sarama/#compatibility-and-api-stability>` section for more details. 
+MinIO requires at *minimum* Kafka 0.9.0 or 0.10.0.
```
ravindk89, line 20: @vadmeste @poornas So this is mostly taken from the existing docs, and it doesn't make much sense to me. We state we *require* 0.9 and 0.10, but the Go library referenced here is only guaranteeing support for the two most recent major releases (2.7.X I think).

source/includes/common-mc-admin-config.rst
```diff
@@ -0,0 +1,978 @@
...
+.. start-minio-notify-amqp-no-wait
+
+Specify ``on`` to enable non-blocking message delivery. Defaults to 'off'.
+
+.. end-minio-notify-amqp-no-wait
+
+
+.. start-minio-notify-amqp-internal
+
+Specify ``on`` to use the exchange only if it is bound to other exchanges.
```
ravindk89, line 68: @poornas @vadmeste this explanation is coming straight from the CLI - I think we could use a bit more clarity here, and I couldn't find much within AMQP as to what 'bound to other exchanges' signified.

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
vadmeste: > but the Go library referenced here is only guaranteeing support for the two most recent major releases (2.7.X I think).

Could you find where did you find this information ?

source/includes/common-mc-admin-config.rst
vadmeste: > @poornas @vadmeste this explanation is coming straight from the CLI - I think we could use a bit more clarity here, and I couldn't find much within AMQP as to what 'bound to other exchanges' signified.

Well as I saw, exchanges are a known term in AMQP, but I am not sure why would someone need this option. Better to ping the developer who sent AMQP support

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
ravindk89: https://github.com/Shopify/sarama#compatibility-and-api-stability

> Sarama provides a "2 releases + 2 months" compatibility guarantee: we support the two latest stable releases of Kafka and Go, and we provide a two month grace period for older releases. This means we currently officially support Go 1.14 through 1.16, and Kafka 2.6 through 2.7, although older releases are still likely to work

Not sure how tight our 0.9/.10 requirement is.

source/includes/common-mc-admin-config.rst
ravindk89: I'll dig around.

ravindk89: Ah - https://www.rabbitmq.com/e2e.html. Not sure how I missed that.

source/includes/common-mc-admin-config.rst
ravindk89: ah - @Praveenrajmani @kannappanr looks like quite a bit of the Kafka target was authored by the both of you. Any insight here?

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
ravindk89: cc @Praveenrajmani @kannappanr on the above, as you have some recent-ish commits for Kafka

source/includes/common-mc-admin-config.rst
ravindk89: cc @Praveenrajmani @kannappanr on the above

Praveenrajmani: You can point it to this https://golang.org/pkg/crypto/tls/#ClientAuthType @ravindk89 That is actually an Enum.

ravindk89: Great - I saw that, but wasn't honestly sure what it would do - but if we can point at that and call it done I've got no objections.

source/includes/common-mc-admin-config.rst
Praveenrajmani: https://github.com/shopify/sarama/blob/v1.20.1/config.go#L327

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
Praveenrajmani: Can we not assure this - `MinIO requires at *minimum* Kafka 0.9.0 or 0.10.0.` without proper testing?

source/includes/common-mc-admin-config.rst
Praveenrajmani: > The version of Kafka that Sarama will assume 

[sarama](https://github.com/Shopify/sarama) is the client pkg here that we use

source/monitoring/bucket-notifications/publish-events-to-kafka.rst
ravindk89: @Praveenrajmani no idea - this is based on what we currently have documented [here](https://docs.min.io/docs/minio-bucket-notification-guide.html)

> MinIO requires Kafka version 0.10 or 0.9. Internally MinIO uses the Shopify/sarama library and so has the same version compatibility as provided by this library.

I'm *guessing* we just never went back to update those version minimums. The last commit to that section was 2 years ago, and it's unrelated beyond linking things. So it's probably been 2+ years.

If we can safely assume MinIO supports the versions the `sarama` library supports, then we can safely drop the 0.9.0/0.10.0 suggestion.

I wasn't able to test this myself, just based on time constraints - this is unfortunately an "import" pass, focusing on formatting and chasing down areas of clarity. Fully testing each of the notification targets would significantly increase the time required to complete any one of the target pages.

Praveenrajmani (approved):

ravindk89:
https://docs.min.io/minio/baremetal/monitoring/monitoring-overview.html Published, thanks all.

PR #332, opened 2021-04-01
MinIO Implicit Prefix Trimming on Removal of Last Object 

ravindk89:
mc's removal API trims implicitly created prefixes as part of removing the last object in that prefix. This can result in deleting an entire "path" if that path was created implicitly. 

`mc` *also* applies this logic to filesystem directories, which can result in aggressive trimming of directories. This has surprised some users in the past, as per  [mc issue 3621](https://github.com/minio/mc/issues/3621)

This PR documents the trimming behavior for those mc commands which rely on the remove API for normal operations. Specifically:

- `mc rm`
- `mc mv`
- `mc ilm`
- `mc mirror`

# Build Instructions

5) Run `make html`

# Key URLS

- http://127.0.0.1:8000/reference/minio-cli/minio-mc/mc-mv.html#minio-trims-empty-prefixes-on-object-removal
- http://127.0.0.1:8000/reference/minio-cli/minio-mc/mc-rm.html#minio-trims-empty-prefixes-on-object-removal
- http://127.0.0.1:8000/reference/minio-cli/minio-mc/mc-mirror.html#minio-trims-empty-prefixes-on-object-removal
- http://127.0.0.1:8000/reference/minio-cli/minio-mc/mc-ilm.html#minio-trims-empty-prefixes-on-object-removal

### DIFF
--- a/source/includes/common-admonitions.rst
+++ b/source/includes/common-admonitions.rst
@@ -16,3 +16,51 @@
    this procedure.
 
 .. end-restart-downtime
+
+
+.. Used in the following pages:
+
+   /reference/minio-cli/minio-mc/mc-rm.rst
+   /reference/minio-cli/minio-mc/mc-mv.rst
+   /reference/minio-cli/minio-mc/mc-mirror.rst
+
+.. start-remove-api-trims-prefixes
+
+|command| relies on the :mc:`mc` removal API for deleting objects. As part of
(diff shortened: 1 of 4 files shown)


ebozduman (approved):
LGTM

Just a note: 
I am also working on a fix for mc issue #3621, to stop removing prefixes of a file in a filesystem when the file is removed.
So, we won't need this/filesystem part of the explanation anymore, when the fix is merged.

PR #333, opened 2021-04-01
Fix typo

error418:
changed "envrionment" to "environment"


ravindk89:
Thanks for catching this @error418 ! Much appreciated.

PR #334, opened 2021-04-06
DOCS-330: Active-Active Replication Tutorial

### LINKED ISSUE #330: MinIO Replication
Breaking this up into parts:

1) MinIO Bucket Replication (Overview + A -> B Replication + refresh `mc replicate`)

2) MinIO Active Active replication (A <- > B)

3) MinIO `mc mirror` replication (one-time)

4) MinIO `mc mirror` active-active replication

ravindk89:
# Summary

Add Active-Active replication tutorial. Partially resolves #330 . 

# Goals

- Create Active-Active replication tutorial

# Non-Goals

- Fully address details of how replication works
- Examples for `mc replicate`
- Additional tutorials
- Diagrams

# Build Instructions

5) Run `make html`

# Key URLS

- http://127.0.0.1:8000/replication/replication-overview.html
- http://127.0.0.1:8000/replication/enable-server-side-two-way-bucket-replication.html


source/replication/enable-server-side-two-way-bucket-replication.rst
```diff
@@ -0,0 +1,514 @@
...
+*strongly recommends* :ref:`enabling TLS <minio-TLS>` on both source and
+destination clusters to ensure the safety of objects during transmission.
+
+MinIO does *not* support replicating client-side encrypted objects 
+(SSE-C).
+
+Procedure
+---------
+
+1) Configure User Accounts and Policies for Replication
```
ravindk89, line 261: This is a new attempt at trying to ensure users don't get caught on setting up users/permissions.

source/replication/enable-server-side-two-way-bucket-replication.rst
```diff
@@ -0,0 +1,514 @@
...
+A\) Create Replication Administrators
+   The following code creates policies and users for supporting configuring
+   replication on the ``Alpha`` and ``Baker`` clusters. Replace the
+   password ``LongRandomSecretKey`` with a long, random, and secure secret key 
+   as per your organizations best practices for password generation.
+
+   .. code-block:: shell
+      :class: copyable
+
+      wget https://docs.min.io/minio/baremetal/examples/ReplicationAdminPolicy.json | \
```
ravindk89, line 287: Note: These URLS won't resolve until this is public. In the short-term you can wget the files from `source/examples/`

source/replication/enable-server-side-two-way-bucket-replication.rst
```diff
@@ -0,0 +1,514 @@
...
+         }
+
+      The ``"arn:aws:s3:::*`` resource applies the replication permissions to
+      *any* bucket on the source cluster. You can restrict the user policy to
+      specific buckets as-needed.
+
+      Use the :mc-cmd:`mc admin policy set` command to associate the policy to
+      a user on the source MinIO cluster.
+
+   .. tab:: Replication Remote Admin
```
poornas, line 135: would prefer to see this `Replication Remote Admin` referred to as a user rather than admin. w.r.t replication, the admin needs permissions to configure the remote targets, while the replication user just needs permissions to receive replicated content.

source/replication/enable-server-side-two-way-bucket-replication.rst
```diff
@@ -0,0 +1,514 @@
...
+
+As part of the replication process, MinIO *decrypts* the object on the source
+bucket and transmits the unencrypted object. The destination MinIO cluster then
+re-encrypts the object using the destination bucket SSE-S3 configuration. MinIO
+*strongly recommends* :ref:`enabling TLS <minio-TLS>` on both source and
+destination clusters to ensure the safety of objects during transmission.
+
+MinIO does *not* support replicating client-side encrypted objects 
+(SSE-C).
+
```
poornas, line 257: Might want to mention that replication between object lock enabled buckets is possible only when both sides have object-locking enabled. Also needs additional object-locking, legal hold/retention permissions in the policy statement

poornas (changes requested):
Looks much clearer and user friendly @ravindk89. Just have a couple of suggestions that might be good to touch upon

source/replication/enable-server-side-two-way-bucket-replication.rst
ravindk89: Hm. This would also be true for encryption. Do we have a list anywhere of the required permissions? I think it would require added perms for both the source replication admin and the destination replication user

ravindk89: Done!

ravindk89 (commented):
@poornas feedback addressed, a few other fixes here and there but nothing too major.

poornas (approved):
LGTM

PR #338, opened 2021-04-14
Reduce content x-padding

rushenn:


ravindk89:
LGTM - though to float an idea - it looks like `.header` inherits from `.content`. We can explicitly set the padding in `_header.scss` so that it still matches the other docs properties. Thoughts?

PR #343, opened 2021-05-19
SCSS fixups

ravindk89:
Hey Rushan,

Was doing some tweaks and accidentally simplified the left-hand toc while cleaning up some CSS weirdness. The Javascript might be a little kludgy but it works. LMK what you think.


source/_static/js/main.js
```diff
@@ -6,6 +6,23 @@ window.addEventListener("DOMContentLoaded", (event) => {
 			.appendChild(document.getElementById("table-of-contents"));
 	}
 
+   var l3 = document.getElementsByClassName("toctree-l3 current")
+   var l2 = document.getElementsByClassName("toctree-l2 current")
+   var l1 = document.getElementsByClassName("toctree-l1 current")
+
+   if (l3.length > 0) {
```
ravindk89, line 13: Since Sphinx doesn't naturally have a way of injecting the path to the "parent" `li` item, this seemed the next best thing. Based on how Sphinx does apply the `current` tag, we can use that to build the full "tree" to parent and apply some class tags to simplify the CSS application a bit.

So I'm not checking arr length here because if l3.length > 0, l2.length and l1.length must necessarily be > 0 (because of how the toc chaining works).

That said if this is just all the bad, I can extend this out to include the full chain of checking.

rushenn (approved):
LGTM

PR #365, opened 2021-07-23
Cleaning CSS, updating packages

ravindk89:
# Description

General fixups to the CSS to make things look a little cleaner, and to fix some random broken stuff.

See #minio-docs for staging URL.


requirements.txt
```diff
@@ -1,6 +1,6 @@
 sphinx == 3.5.0
```
rushenn, line 1: @ravindk89 Looks like extension `myst_parser` is missing from `requirements.txt`, leading to an error while building the HTML.

rushenn:
```
Extension error:
Could not import extension myst_parser (exception: No module named 'myst_parser')
```

ravindk89: I cleaned out my local venv, fixed requirements.txt, and tried again - error appears to be cleared out.

ravindk89: @rushenn LMK if you still run into trouble.

rushenn (approved):
LGTM

PR #373, opened 2021-08-11
Optimize PNG  images

klauspost:
Saves around 20-30% per file with lossless recompression.

[FileOptimizer](https://nikkhokkho.sourceforge.io/static.php?page=FileOptimizer) used.


ravindk89:
I *just* noticed this @klauspost , no idea why I didn't get a github notification. Thanks for pointing the utility out, I've just been dumping everything out of GIMP and hoping for the best :)

PR #374, opened 2021-08-19
Improving docs around Console config

ravindk89:
# Summary

There's an increase of users stumbling on Console issues w/ TLS enabled. LetsEncrypt (and seemingly most other providers) tend to include DNS SAN, and not IP SAN - so Console cannot validate the TLS connection and fails to connect.

We already have `MINIO_SERVER_URL`  for fixing this - this update brings `MINIO_SERVER_URL` into the deployment tutorials with a bit of guidance on when to use it.


harshavardhana (approved):

klauspost (approved):
LGTM

PR #378, opened 2021-09-08
DOCS-376: Integrate Sphinx Design, remove sphinx-tabs

ravindk89:


source/_static/js/main.js
```diff
@@ -30,3 +30,35 @@ window.addEventListener("DOMContentLoaded", (event) => {
 		$('body').toggleClass(target + '-toggled');
 	});
 });
+
+window.addEventListener('hashchange', function(){
```
ravindk89, line 34: @rushenn @kaankabalak this might be the hackiest JS I've ever used, but here we are.

I've integrated a new library https://github.com/executablebooks/sphinx-design which replaces some old contents with nice shiny new bootstrap/material design components.

The hack below allows tabs to "switch" when an anchor element is hidden (due to tab behavior).

It's kind of janky, and I have an open issue to maybe reduce the jank (https://github.com/executablebooks/sphinx-design/issues/30), but absent that - does this have any obvious "this might break a lot of things" consequences I haven't considered?

source/_static/scss/includes/_variables.scss
```diff
@@ -2,8 +2,20 @@
...
 $black: #000000;
 $theme-red: #C72C48;
+$theme-red-light: rgba($theme-red, 0.5);
+$theme-red-lighter: rgba($theme-red, 0.25);
 $theme-hover-bg: rgb(245, 245, 245);
 
+// Overrides for sphinx-design
+// https://sphinx-design.readthedocs.io/en/alabaster-theme/css_variables.html
+
+:root {
```
ravindk89, line 12: @rushenn @kaankabalak 

overrides for sphinx-design components. This seemed to work, but I'm not sure if using `:root` is the correct thing to do here.

rushenn: Adding CSS variables isnide `:root`, make them accessible from anywhere in the stylesheet. I think this is fine as we don't have any specific style config requirements.

source/_static/js/main.js
```diff
@@ -30,3 +30,35 @@ window.addEventListener("DOMContentLoaded", (event) => {
...
+   // <div class="sd-tab-content"> Some stuff </div>
+   // <radio id="foo"></radio>
+   // <label for="foo"</label>
+   // <div class="sd-tab-content"> <p id="target">Foo</p></div>
+   //
+   // Getting to the parent div isn't too hard. But identifying the correct
+   // label to modify requires "looking" back one element and hoping
+   // the ordering is correct.
+
+   var parentPanel = document.getElementById(window.location.hash.substr(1)).closest(".sd-tab-content");
```
rushenn, line 58: This will mostly return null as the radio elements are placed in the same level as `.sd-tab-content`. Please check the below screenshot.

Screenshot 2021-09-09 at 2 22 30 PM

As we already using jQuery, we can achieve the same with fewer codes as follows:

```
var hash = window.location.hash;
if (hash) {
   var target = $('.sd-tab-set ' + hash);  // .sd-tab-set is the parent wrapper of the tab group so adding it here in order to make sure the available ID is a tab
   if(target[0]) {
       target.prop('checked', true);
   }
}
```

PR #385, opened 2021-09-15
Fix region setting AWS SDK for .NET example

ferhrosa:
I tried to use MinIO with the AWS SDK for .NET and got the same error descripted on this post:
https://stackoverflow.com/questions/68678953/error-when-configuring-minio-in-c-sharp-the-authorization-header-is-malformed/69199735

So I found that the solution was to change the `RegionEndpoint` setting to `AuthenticationRegion`, as the region is used on the authentication process.
Also the AWS SDK sets the `RegionEndpoint` as null when `ServiceURL` is set.

### DIFF
--- a/docs/aws-sdk-for-dotnet-with-minio.md
+++ b/docs/aws-sdk-for-dotnet-with-minio.md
@@ -42,7 +42,7 @@ class Program
     {
         var config = new AmazonS3Config
         {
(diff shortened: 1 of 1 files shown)


ravindk89:
Hi @ferhrosa - sorry, I just noticed this pull (not sure why I didn't get a notification for this). 

Let me see if I can get someone internally to double check this, I don't have enough insight into .NET to validate the fix.

kannappanr (approved):
LGTM

PR #389, opened 2021-09-23
DOCS-386: Update replication docs for multi-site replication

### LINKED ISSUE #386: MinIO Multi-Site Replication
**Summary**

https://github.com/minio/minio/commit/c4373ef290344fd01981864cb9feca9410cf5b21 adds Multi-Site replication to master.

This should go out as part of the next release, so the current [Replication Docs](https://docs.min.io/minio/baremetal/replication/replication-overview.html) need updating to remove restrictions on the number of replication destinations per bucket.

**Goals**

- [ ] Remove graf in [Enable Two-Way Server-Side Bucket Replication](https://docs.min.io/minio/baremetal/replication/enable-server-side-two-way-bucket-replication.html) related to multi-site replication
- [ ] Update Considerations to include a blurb on setting up multi-site properly (e.g. repeat this procedure for each "pair" of sites"
- [ ] Update one-way replication docs (if necssary)

**Non-Goals**

Multi-Site replication tutorial - the active-active tutorial should be sufficient for now, with notes to repeat the procedure for each connection in the multi-site setup.

**Additional context**

Once complete, merge and publish *after* the next stable MinIO Release

ravindk89:
Closes #386 

- [ ] Document new multisite replication features
- [ ] Update previous tutorials to mention multisite capabilities
- [ ] General cleanups

Once approved merge only *AFTER* the upcoming MinIO version is published. Replace VERSION placeholders with that version *prior to merging*


poornas (changes requested):
@ravindk89 , I am fine with these docs. However, we also need to document the `mc replicate` command change to use --remote-bucket flag instead of --arn. When multiple targets are being added, the old command will not work for setting up more than one target since the --arn was representative of RoleArn field (which is unique). If users plan on using multiple destination replication, they will have to upgrade to latest mc and pass the remote target in the --remote-bucket field at least from the point of adding the second target

ravindk89:
@poornas thanks for reminding me, I'll fix that as well. 

To double check - I assume we are supporting `--arn` as a legacy field, but moving forward `--remote-bucket` will work for both single-site and multi-site?

poornas:
> @poornas thanks for reminding me, I'll fix that as well.
> 
> To double check - I assume we are supporting `--arn` as a legacy field, but moving forward `--remote-bucket` will work for both single-site and multi-site?

that's correct

PR #392, opened 2021-10-01
Add MinIO supported compression formats.

klauspost:


klauspost:
Please check formatting. I don't have it set up to test the output.

source/reference/minio-cli/minio-mc/mc-sql.rst
```diff
@@ -230,6 +230,13 @@ Syntax
    - ``BZIP2``
    - ``NONE`` (default)
 
+   Compression schemes supported by MinIO backend only:
```
ravindk89, line 233: Syntax is fine - one clarifying question. By `MinIO backend only`, do we mean that these commands only work if `mc sql` is targeting these files stored on MinIO vs another S3 storage?

klauspost: @ravindk89 Files stored on MinIO only.

ravindk89: OK just checking - this is perfect, thank you for opening the pull :)

ravindk89 (approved):

PR #401, opened 2021-10-26
DOCS-396: Update deployment/expansion guides to use systemd

### LINKED ISSUE #396: Improve Installation/Deployment Documentation
**Summary**

Our current docs tend to include examples of running MinIO in the foreground. This isn't really helpful for anyone who needs to run a distributed MinIO deployment.

We have a very nice [systemd](https://github.com/minio/minio-service/tree/master/linux-systemd) example that would be great to integrate into our docs.

Engineering has also indicated that installing via the deb/rpm packages may include installing that systemd file, which needs confirmation for RHEL8 / Ubuntu LTS.

**Goals**

- [ ] Validate RHEL/Ubuntu installation instructions and systemd default config (if any)
- [ ] Update Distributed + Expansion docs to use systemd for running MinIO

**Non-Goals**

- Guidance on non-systemd forms of running MinIO
- Guidance for Windows/OSX
- 

**Additional context**
There's a certain support burden that comes from users guessing their way through setting this up. Experienced sysadmins/techops/devops are familiar with using systemd scripts, but having better boilerplate documentation on this should reduce the amount of time spent cycling through basic troubleshooting, only to find a misconfigured systemd file.

ravindk89:
Closes #396 

# Summary

Our existing docs tend to run MinIO as a foreground process and trust the user to fill in the blanks for production.

While this has worked thus far, it would reduce our support burden and ensure a good foundation for all users to be more prescriptive in starting MinIO as a system service. This is especially easy because not only do we already publish service file examples, but our DEB/RPM packages install them by default. 

The goals here are as follows:

- [ ] - Update the Distributed MinIO and Expand MinIO tutorials to use systemd
- [ ] - Include examples of setting the environment file
- [ ] - Better call out prereqs around networking, load balancers, etc.
- [ ] - Note experimental support for Windows and direct users to appropriate channel

A few notes:

- This assumes we do some work to add a symlink for the latest stable `minio` binary for `minio.deb` and `minio.rpm`. 
- This assumes most users can run MinIO as a service. 

# Non-Goals

There are a few things out of scope here:

- A complete guide to setting up TLS (e.g. self-signed certs out of the box)
- Guidance on setting up a load balancer (Future work, likely involving Sidekick first, then HAProxy/Nginx)
- Guidance on Windows/OSX setups, as we rarely see these in production
- Compose/Swarm - our focus is Kubernetes-based deployments for orchestration.

# Local Builds

I will link a staged view in Slack, but local builds don't require too much:

- Create a python virtual environment in the project root and activate it
- Run `pip install -r requirements.txt` (requires Python 3)
- Run `make html`
- Run` python -m http.server --directory build/DOCS-396/html`
- Open your browser and go to `localhost:8080` and browse around.

### DIFF
--- a/source/includes/common-installation.rst
+++ b/source/includes/common-installation.rst
@@ -1,0 +1,181 @@
+.. start-install-minio-binary-desc
+
+The following tabs provide examples of installing MinIO onto 64-bit Linux
+operating systems using RPM, DEB, or binary. The RPM and DEB packages
+automatically install MinIO to the necessary system paths and create a
+``systemd`` service file for running MinIO automatically. MinIO strongly
+recommends using RPM or DEB installation routes.
+
+.. tab-set::
+
+   .. tab-item:: RPM (RHEL)
+      :sync: rpm
+
+      Use the following commands to download the latest stable MinIO RPM and
+      install it.
+
+      .. code-block:: shell
+         :class: copyable
+
+         wget https://dl.min.io/server/minio/release/linux-amd64/minio.rpm
+         sudo dnf install minio.rpm
+
+   .. tab-item:: DEB (Debian/Ubuntu)
+      :sync: deb
+
+      Use the following commands to download the latest stable MinIO DEB and
+      install it:
+
+      .. code-block:: shell
+         :class: copyable
+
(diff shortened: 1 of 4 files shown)


kannappanr (approved):
LGTM

PR #402, opened 2021-11-15
Rectify the typo in user creation for replication

sinhaashish:
There was a typo(lowercase was used) in user creation while replication which led to the failure of the command. 

Error 
```
mc: <ERROR> Unable to initialize admin connection. No valid configuration found for 'baker' host alias.
````


klauspost (approved):
LGTM

PR #408, opened 2021-11-25
QuickWin: Add fstab instructions for disk mounting

ravindk89:
Addressing quickwin of documenting explicitly using fstab to preserve drive mount order.

The `start-local-jbod-desc` section is included in both deployment and expansion tutorials. The `|deployment|` placeholder is replaced at build time with the appropriate subject for the tutorial.

let me know if you want a staged view.


klauspost:
LGTM, but I don't really know much about this.

harshavardhana (approved):

PR #422, opened 2022-01-31
Update header and other properties as per the main website

rushenn:


harshavardhana (approved):

ravindk89:
Looks solid. Also looks like it might have removed a bunch of security issues too.

PR #425, opened 2022-02-15
Update to sphinx 4

ravindk89:
# Summary

Python 3.9 fails on older versions of Sphinx. The update was added in Sphinx 4.x.

Sphinx 4.4 has a feature that by default throws tons of warnings due to some well-meaning pattern matching gone wrong. The fix is slated for 4.4.1.

So we're using 4.3.2

# Things to Review

- The HTML/CSS seems to have changed subtly. I've made minor modifications to aside.scss but I'm not sure if they're useful.
- adding main.min.css and main.css to gitignore since we generate those on build, there's no point in keeping them in git.
- Fix tabs css on /design.html


ravindk89:
@rushenn @kaankabalak staging can be found here:

- https://play.min.io/docs-sphinx-update/design.html
- https://play.min.io/docs-sphinx-update/design.html#tabs

As you can see the CSS still needs some work.

With this update make sure to use Python 3.10 to build. requirements.txt has been updated.

rushenn:
@ravindk89 @kaankabalak I'm working on the style issues. I will send a commit to this PR.

harshavardhana (approved):

source/_static/scss/includes/_aside.scss
```diff
@@ -1,4 +1,4 @@
-.sidebar {
+div.sidebar,aside.sidebar {
```
rushenn, line 1: @ravind It seems like if we remove the element ref from the class names, it would fix the style issues. So we can keep it simple as before `.sidebar` { ... }

source/_static/scss/includes/_aside.scss
```diff
@@ -30,7 +30,7 @@
     }
 }
 
-.sidebar-active {
+div.sidebar-active {
```
rushenn, line 33: `sidebar-active` active class is set on the body tag to toggle the sidebar. Currently the toggling isn't working due to element `div` ref. Can we keep the selector simple as `.sidebar-active`?

source/_static/scss/includes/_aside.scss
```diff
@@ -47,7 +47,7 @@
     }
 }
 
-.sidebar__title {
+div.sidebar__title {
```
rushenn, line 50: Same here as well. `sidebar__title` is an `h3` element.

ravindk89:
@rushenn removing those selectors breaks the left-hand nav for me in Firefox.

image

For some reason it pulls from `basic.css` instead.

source/_static/scss/includes/_aside.scss
ravindk89: Dropping `div.sidebar` seems to be fine, but `aside.sidebar` appears necessary.

ravindk89:
I'm not sure *why* `basics.css` is matching better now than before, but that seems to be the issue.

Changing to just `aside.sidebar` does seem to work though. Making a few other small tweaks.

Still can't figure out why the tabs design isn't respecting new root css variables.

ravindk89:
https://play.min.io/docs-sphinx-update/design.html#tabs seems fine now. Bizarre. Maybe local python server caching issue. 

@rushenn LMK if you have additional concerns on the remaining changes. On FF latest stable it seems fine.

rushenn:
@ravindk89 For me it works fine on all major browsers including Firefox. I'm seeing issues when I update the CSS class names with elements. I think the latest play link seems to be working without any issues.

rushenn (approved):
LGTM

ravindk89:
> @ravindk89 For me it works fine on all major browsers including Firefox. I'm seeing issues when I update the CSS class names with elements. I think the latest play link seems to be working without any issues.

Interesting. One thing we can do long-term is completely remove our dependency on the alabaster theme. We'll have to build out the jinja templates / underlying CSS entirely ourselves, *but* we should no longer see any weird behaviors like this. We would also have much more control over the layout of things.

Issue #430, opened 2022-03-04
[BUG] Network Encryption documentation refers to public.cert but code requires public.crt


markslater:
**Describe the bug**
The [Network Encryption (TLS)](https://docs.min.io/minio/baremetal/security/network-encryption/minio-tls.html) documentation (inconsistently) refers a number of times to creating a file called `public.cert` containing the public certificate chain, but this name is incorrect.  The correct name for the file is `public.crt`.

**To Reproduce**
The code [specifies](https://github.com/minio/minio/blob/f527c708f248b7fc9731db51fa4d815b5ae00e23/cmd/config-dir.go#L38) that the file is called `public.crt`.

**Expected behavior**
The documentation should use the correct filename.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A
**Additional context**

ravindk89:
Thanks for catching this (I have got to figure out why github doesn't fire notifications on new issues or pulls)

I've merged your PR, and am rebuilding the docs now. That should resolve the issue

harshavardhana:
> Thanks for catching this (I have got to figure out why github doesn't fire notifications on new issues or pulls)

@ravindk89 perhaps "click on watch" the repo you might have unwatched it by mistake.

PR #431, opened 2022-03-04
#430 - Correct name of certificate chain file from public.cert to public.crt

markslater:
Replace references to public.cert with public.crt.


ravindk89 (approved):

ravindk89:
Thanks for the fixup @markslater !

PR #435, opened 2022-03-16
Fix typo in mc commands

schemacs:


ravindk89:
Thanks for the PR @schemacs - Github is having a bad day so I can't merge this yet.

I did a quick walkthrough of our docs to see if this is the only typo of this kind - I am somewhat tempted to apply the style from the multi-site docs where we just use the same usernames for each site, as it works effectively the same in the end but might be less confusing. Work to do later (cc @djwfyi ) if we go that route.

Once github stabilizes I can merge and refresh the build. We appreciate your contributiou.

schemacs:
@ravindk89 The usernames/accesskeys are correct in multi-site docs, but I will give you another PR to fix some small command multi-line continuations.

PR #442, opened 2022-03-17
MC Mirror flag clarification

### LINKED ISSUE #377: Improve `mc mirror` documentation
# Summary

`mc mirror` has a number of flags whose behavioral interactions are a little vague at times. Specifically:

- `mc mirror` synchronizes content between SITE A and SITE B. It does not necessarily capture ongoing operations.
- `mc mirror --watch` synchronizes SITEA -> SITEB *and* watches for incoming **PUT** and **DELETE** operations to synchronize to SITEB. 
-  `mc mirror --watch --remove` **removes any object on SITEB which does not exist on SITEA**. 
- `mc mirror --watch --overwrite` overwrites an object on SITEB if it differs from SITE A
- `mc mirror --active-active` combines `--watch --overwrite --remove` behavior under one flag.

These behaviors are very specific in their interactions. 

This should hopefully resolve some of the confusion raised elsewhere:

- https://github.com/minio/mc/issues/3413
- https://github.com/minio/mc/pull/3787

# Goals

- Update `mc mirror` documentation with examples of how to use these flags in isolation and together
- Update the Replication documentation with `mc mirror` commands

# Non-Goals

- Exhaustive matrix of usage between these flags
-

djwfyi:
Resolves #377 .

Adds clarifications about three `mc mirror` flags:

* `--watch`
* `--overwrite`
* `--remove`

Does not add any reference to the `--active-active` flag mentioned in the CLI help.
(That flag combines the functions of the above three into one.)

### DIFF
--- a/source/reference/minio-mc/mc-mirror.rst
+++ b/source/reference/minio-mc/mc-mirror.rst
@@ -15,9 +15,8 @@ Syntax
 
 .. start-mc-mirror-desc
 
-The :mc:`mc mirror` command synchronizes content to MinIO deployment, similar to
-the ``rsync`` utility. :mc:`mc mirror` supports filesystems, MinIO deployments,
-and other S3-compatible hosts as the synchronization source.
+The :mc:`mc mirror` command synchronizes content to MinIO deployment, similar to the ``rsync`` utility.
+:mc:`mc mirror` supports filesystems, MinIO deployments, and other S3-compatible hosts as the synchronization source.
 
 .. end-mc-mirror-desc
 
@@ -25,16 +24,17 @@ and other S3-compatible hosts as the synchronization source.
 
    .. tab-item:: EXAMPLE
 
-      The following command synchronizes content from a local filesystem
-      directory to the ``mydata`` bucket on the ``myminio`` MinIO deployment.
+      The following command synchronizes content from a local filesystem directory to the ``mydata`` bucket on the ``myminio`` MinIO deployment.
 
       .. code-block:: shell
          :class: copyable
 
          mc mirror --watch ~/mydata myminio/mydata
 
-      The command "watches" the local filesystem and synchronizes any additional
-      operations to MinIO until explicitly terminated from the terminal:
+      The command "watches" for files added or removed on the local filesystem and synchronizes those operations to MinIO until explicitly terminated. 
+      
+      :mc-cmd:`mc mirror --watch` updates files changed on the local filesystem to MinIO (see :mc-cmd:`~mc mirror --overwrite`).
+      ``--watch`` does not remove other files from MinIO not present on the local filesystem (see :mc-cmd:`~mc mirror --remove`).
 
    .. tab-item:: SYNTAX
 
@@ -70,40 +70,31 @@ Parameters
 
 .. mc-cmd:: SOURCE
 
-   *REQUIRED* The file(s) or object(s) to synchronize to the 
-   :mc-cmd:`~mc mirror TARGET` S3 host.
+   *REQUIRED* The file(s) or object(s) to synchronize to the :mc-cmd:`~mc mirror TARGET` S3 host.
 
-   For objects on S3-compatible hosts, specify the path to the object as 
-   ``ALIAS/PATH``, where:
+   For objects on S3-compatible hosts, specify the path to the object as ``ALIAS/PATH``, where:
 
-   - ``ALIAS`` is the :mc:`alias <mc alias>` of a configured S3-compatible host,
-     *and*
+   - ``ALIAS`` is the :mc:`alias <mc alias>` of a configured S3-compatible host, *and*
 
-   - ``PATH`` is the path to the bucket or object. If specifying a bucket,
-     :mc:`mc mirror` synchronizes all objects in the bucket.
+   - ``PATH`` is the path to the bucket or object. If specifying a bucket, :mc:`mc mirror` synchronizes all objects in the bucket.
 
    .. code-block:: shell
 
       mc mirror [FLAGS] play/mybucket/ myminio/mybucket
 
-   For files on a filesystem, specify the full filesystem path to the file or
-   directory :
+   For files on a filesystem, specify the full filesystem path to the file or directory :
 
    .. code-block:: shell
 
       mc mirror [FLAGS] ~/data/ myminio/mybucket
 
-   If specifying a directory, :mc-cmd:`mc mirror` synchronizes all files in the
-   directory.
+   If specifying a directory, :mc-cmd:`mc mirror` synchronizes all files in the directory.
 
 .. mc-cmd:: TARGET
 
-   *REQUIRED* The full path to bucket in which :mc:`mc mirror` copies
-   synchronized SOURCE objects. Specify the ``TARGET`` as
-   ``ALIAS/PATH``, where:
+   *REQUIRED* The full path to bucket in which :mc:`mc mirror` copies synchronized SOURCE objects. Specify the ``TARGET`` as ``ALIAS/PATH``, where:
 
-   - ``ALIAS`` is the :mc:`alias <mc alias>` of a configured S3-compatible
-     host, *and*
+   - ``ALIAS`` is the :mc:`alias <mc alias>` of a configured S3-compatible host, *and*
 
    - ``PATH`` is the path to the bucket.
 
@@ -111,15 +102,13 @@ Parameters
 
       mc mirror SOURCE play/mybucket
 
-   :mc:`mc mirror` uses the object or file names from the
-   :mc-cmd:`~mc mirror SOURCE` when synchronizing to the ``TARGET`` bucket.
+   :mc:`mc mirror` uses the object or file names from the :mc-cmd:`~mc mirror SOURCE` when synchronizing to the ``TARGET`` bucket.
 
 .. mc-cmd:: --attr
    
 
-   Add custom metadata for mirrored objects. Specify key-value pairs as 
-   ``KEY=VALUE\;``. For example, 
-   ``--attr key1=value1\;key2=value2\;key3=value3``.
+   Add custom metadata for mirrored objects. Specify key-value pairs as ``KEY=VALUE\;``. 
+   For example, ``--attr key1=value1\;key2=value2\;key3=value3``.
 
 .. mc-cmd:: --disable-multipart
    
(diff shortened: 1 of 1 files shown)


harshavardhana (approved):

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -15,26 +15,26 @@ Syntax
...
       .. code-block:: shell
          :class: copyable
 
          mc mirror --watch ~/mydata myminio/mydata
 
-      The command "watches" the local filesystem and synchronizes any additional
-      operations to MinIO until explicitly terminated from the terminal:
+      The command "watches" for files added or removed on the local filesystem and synchronizes those operations to MinIO until explicitly terminated from the terminal. 
+      
+      `--watch` does not update files changed on the local filesystem to MinIO (see `--overwrite`).
```
ravindk89, line 36: If you want to make these linkable, you can use `mc-cmd` - for example:

```suggestion
      `--watch` does not update files changed on the local filesystem to MinIO (see :mc-cmd:`~mc mirror`--overwrite`).
```

The tild automatically truncates everything up to the last argument, so it *should* output a link to the `--overwrite` anchor.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -129,35 +118,27 @@ Parameters
 .. mc-cmd:: --encrypt-key
    
 
-   Encrypt or decrypt objects using server-side encryption with
-   client-specified keys. Specify key-value pairs as ``KEY=VALUE``.
+   Encrypt or decrypt objects using server-side encryption with client-specified keys.
```
ravindk89, line 121: So we don't have to answer this *now*, but it might be worth asking (@aead maybe?) whether this is a blind en/de crypt

How do we handle mixed encryption states? e.g. if the source has both encrypted and unencrypted objects (and for the sake of sanity lets assume that all encrypted objects use the same key), what happens on the destination? Do we decrypt the encrypted objects and encrypt the decrypted objects?

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,61 @@ Parameters
    
    Overwrites object(s) on the :mc-cmd:`~mc mirror TARGET`.
 
+   For example, object A exists on both the Source and the Target.
+   Object A on the source changes.
+   Running `mc mirror --overwrite` replaces the version on the target with the version from the Source.
+
```
ravindk89, line 179: We should restructure this in context of `--overwrite` already running, and then a change occurring:

```suggestion
   For example, consider an active `mc mirror --overwrite` synchronizing content from source to destination.
   
   If an object on the source changes, `mc mirror` synchronizes and overwrites any matching file on the destination.
   
   Without --overwrite, <behavior>

```

The inclusion of *version* here also might need some care. Buckets can also be unversioned, in which case I assume it's just a straight overwrite.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,61 @@ Parameters
...
    
 
-   Removes extraneous object(s) on the target. 
+   Removes object(s) on the Target that do not exist on the Source. 
+
+   For example, objects A, B, and C exist on Source.
+   Objects C, D, and E exist on Target.
+
+   When running `mc mirror --remove`, objects A and B copy to Target and objects D and E are removed from the Target.
+   C already exists on both, so does not change on either.
```
ravindk89, line 201: Might clarify here that C must be identical on source and destination. If the C on the destination differs - actually I'm not sure what would happen there - might be worth testing. With --overwrite we would overwrite, but if we are just checking for filenames and not checksums, the behavior is less clear.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,61 @@ Parameters
...
+
+   After the action, only objects A, B, and C exist on both teh Source and the Target.
```
ravindk89, line 203:
```suggestion
   After the action, only objects A, B, and C exist on both the Source and the Target.
```

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,61 @@ Parameters
...
-   Continuously monitor the :mc-cmd:`~mc mirror SOURCE` path and
-   synchronize changes.
+   Continuously monitor the :mc-cmd:`~mc mirror SOURCE` path and synchronize `PUT` and `DELETE` changes from Source to Target.
 
    Defaults to ``0`` (all objects).
 
+   For example, object A and B exist on the watched Source.
+   Objects A, B, and C exist on the watched Target.
+   
+   An action adds object D to Source and removes object B.
```
ravindk89, line 224: "An action" feels too generic. A client is going to perform a PUT write op. If it's a local filesystem it would be a write operation.

Maybe "A client writes object D to the source"

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,61 @@ Parameters
...
+
+   After the action, objects A and D exist on the Source.
+   Objects A, C, and D exist on the Target.
+   
+   Use `--watch` flag to mirror objects from Source to Target, where the Target may also have additional objects not present on the Source.
```
ravindk89, line 229: Should this be at the top? It's a good modifier to the initial graf to set the context for usage.

ravindk89 (commented):
This is a great first start.

A few things we'll have to do later:

- run a `rebase` to pull in the latest changes from upstream
- squash your commits into one blob

This is great to start though.

source/reference/minio-mc/mc-mirror.rst
djwfyi: Added https://github.com/minio/docs/issues/443 to cover this later.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -15,26 +15,26 @@ Syntax
...
+      The following command synchronizes content from a local filesystem directory to the ``mydata`` bucket on the ``myminio`` MinIO deployment.
 
       .. code-block:: shell
          :class: copyable
 
          mc mirror --watch ~/mydata myminio/mydata
 
-      The command "watches" the local filesystem and synchronizes any additional
-      operations to MinIO until explicitly terminated from the terminal:
+      The command "watches" for files added or removed on the local filesystem and synchronizes those operations to MinIO until explicitly terminated from the terminal.
```
ravindk89, line 34: Nit - terminated from the terminal is a little redundant (and if someone forked the process, that would require different steps)

It should be sufficient to state "until explicitly terminated"

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,67 @@ Parameters
    
    Overwrites object(s) on the :mc-cmd:`~mc mirror TARGET`.
 
+   For example, consider an active ``mc mirror --overwrite`` synchronizing content from Source to Destination.
+
+   If an object on Source changes, ``my mirror --overwrite`` synchronizes and overwrites any matching file on Destination.
+
+   Without ``--overwrite``, if an object already exists on the Destination, the mirror process fails for the object that exists in both locations.
```
ravindk89, line 180: Nit - can simplify

```suggestion
   Without ``--overwrite``, if an object already exists on the Destination, the mirror process fails to synchronize that object.
```

Curiosity - do we fail and continue, or just fail?

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,67 @@ Parameters
...
+   Objects C, D, and E exist on Target.
+
+   When running `mc mirror --remove`, objects A and B copy to Target and objects D and E are removed from the Target.
+   C already exists on both, so does not change on either. 
+
+   After the action, only objects A, B, and C exist on both the Source and the Target.
+
+   Use the `--remove` flag to keep the same set of objects on both Source and Target.
+
+   Use :mc-cmd:`~mc mirror --overwrite` or :mc-cmd:`~mc mirror --watch` if object C differs between Source and Target and the object on Source should be on both.
```
ravindk89, line 209: Nit: simplify

```suggestion
   Use :mc-cmd:`~mc mirror --overwrite` or :mc-cmd:`~mc mirror --watch` to overwrite objects on the Destination if they already exist.
```

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,67 @@ Parameters
...
    
 
-   Removes extraneous object(s) on the target. 
+   Removes object(s) on the Target that do not exist on the Source. 
+
+   For example, objects A, B, and C exist on Source.
+   Objects C, D, and E exist on Target.
+
+   When running `mc mirror --remove`, objects A and B copy to Target and objects D and E are removed from the Target.
+   C already exists on both, so does not change on either.
```
ravindk89, line 203: Might need to be more specific here - it's all about namespaces, when the expectation might be CRC/hash. So we're only looking to see if the set of namespaces on Source and Destination differ, and trimming the Destination to match.

An object named C exists on both source and destination. `mc mirrror --remove` only checks that "C" exists and does not verify that the contents match.

Might need to verify that is true if you haven't poked it already. I would assume so, because that would make the need for overwrite more clear

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,67 @@ Parameters
...
+   For example, object A and B exist on the watched Source.
+   Objects A, B, and C exist on the watched Target.
+   
+   A client writes object D to Source and removes object B.
+
+   After the operation, objects A and D exist on the Source.
+   Objects A, C, and D exist on the Target.
+
+   :mc-cmd:`~mc mirror --watch` implies the :mc-cmd:`~mc mirror --overwrite` flag.
+   While watching the Source, files changed on Source also update on the Target.
```
ravindk89, line 234: This feels out of place here - should it be up in 221? We could break it up into three points:

- --watch continuously synchronizes files from source to destination
- The destination may have files that do not exist on the source
- --watch overwrites objects on the Destination if they match on the Source

ravindk89 (commented):
Looking better - more nits and some additional thoughts.

source/reference/minio-mc/mc-mirror.rst
djwfyi: Fail and continue.

source/reference/minio-mc/mc-mirror.rst
djwfyi: Adding a paragraph that points to the `--overwrite` flag.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,70 @@ Parameters
...
+   For example, objects A, B, and C exist on Source.
+   Objects C, D, and E exist on Target.
+
+   When running ``mc mirror --remove``, objects A and B copy to Target and objects D and E are removed from Target.
+   Since an object C already exists on both, nothing moves from Source to Target. 
+
+   After the action, only objects A, B, and C exist on both the Source and the Target.
+
+   Note: MinIO does not verify that the contents of object C are the same on both Source and Target, only that an object called `C` exists on both.
+   To make object C on Target the same as object C on Source, use  :mc-cmd:`~mc mirror --overwrite` or :mc-cmd:`~mc mirror --watch`.
```
ravindk89, line 211:
```suggestion
   To ensure objects on the Source and Destination match both names *and* content, use  :mc-cmd:`~mc mirror --overwrite` or :mc-cmd:`~mc mirror --watch`.
```

ravindk89 (approved):
LGTM with nits. Once you take/mod/toss those let's sync to squash this down into one commit.

source/reference/minio-mc/mc-mirror.rst
```diff
@@ -191,44 +173,70 @@ Parameters
    
    Overwrites object(s) on the :mc-cmd:`~mc mirror TARGET`.
 
+   For example, consider an active ``mc mirror --overwrite`` synchronizing content from Source to Destination.
+
+   If an object on Source changes, ``my mirror --overwrite`` synchronizes and overwrites any matching file on Destination.
+
+   Without ``--overwrite``, if an object already exists on the Destination, the mirror process fails to synchronize that object.
+   MinIO logs an error and continues to synchronize other objects.
```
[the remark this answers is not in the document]
djwfyi: Well that seems obvious.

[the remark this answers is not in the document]
djwfyi: That is, it seems obvious that it should be `mc mirror` here.

source/reference/minio-mc/mc-mirror.rst
djwfyi: Using `Target` instead of `Destination` to match the rest of this section.
Otherwise, 👍

PR #451, opened 2022-03-25
Adding content for OSX quickstart

djwfyi:
Content for the MacOS Quickstart.

Use this to figure out the right structure and content for a quickstart.
Then I'll move things as appropriate to the `includes` folder for single sourcing across the quickstarts.


source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
-For instructions on deploying to production environments,
-see :ref:`deploy-minio-distributed`.
+This page deploys MinIO on |OS| for use with early development and evaluation environments.
+
+For instructions on deploying to production environments, see :ref:`deploy-minio-distributed`.
+
+The procedure described here deploys MinIO in :guilabel:`Standlone Mode`` on a directory on the |OS| harddrive.
+A standalone MinIO deployment consists of a single MinIO server process with a single drive or storage volue (what we call "filesystem mode").
+The MinIO server provides an S3 access layer to the drive or volume.
+
```
ravindk89, line 18: I think we are repeating ourselves a little bit here. I think we can compact 11, 15, and 16 into a single point, and then come in with 17. 

Spitballing:

This procedure deploys a single MinIO server in Standalone Mode using a directory on the local filesystem. Standalone mode ("filesystem mode") provides an S3 access layer to the drive or volume.

Standalone deployments are for use with...for instructions on production...

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+   MinIO cannot provide consistency guarantees if the underlying storage volumes are NFS or a similar network-attached storage volume. 
+
+   For deployments that *require* using network-attached storage, use NFSv4 for best results.
 
 Procedure
 ---------
 
-#. Step 1
+Install the MinIO Server
```
ravindk89, line 39: So you can still number these, but you might have to escape some characters. I _think_:

```
1\) Install the MinIO Server
~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```


Might work.

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Installation with `Homebrew <https://brew.sh>`_ is recommended.
+
+.. tab-set::
+
+   .. tab-item:: Homebrew
+
+      Open a Terminaal and run the following command to install the latest stable MinIO package using `Homebrew <https://brew.sh>`_.
```
ravindk89, line 48:
```suggestion
      Open a Terminal and run the following command to install the latest stable MinIO package using `Homebrew <https://brew.sh>`_.
```

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+
+         wget https://dl.min.io/server/minio/release/darwin-amd64/minio
+         chmod +x minio
+
+Test the :mc:`minio server`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Launch the :mc:`minio server` and use the :ref:`MinIO Console <minio-console>` to work with the MinIO volume.
+
+#. Launch the server
```
ravindk89, line 80: Having substeps here feels a little out of spec with the previous.

I think we should either do this as three steps:

- Download/Install
- Launch
- Connect

*or* do it as a two-stage setup

"Download and Launch"
"Connect and Explore"

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+   .. tab-item:: Binary
+
+      Download the standalone MinIO server for MacOS and make it executable.
+            
+      .. code-block:: shell
+         :class: copyable
+
+         wget https://dl.min.io/server/minio/release/darwin-amd64/minio
+         chmod +x minio
+
```
ravindk89, line 155: Include a command (outside of tabs) for creating the alias and connecting.

Since everything is default:

```sh
mc alias set local http://127.0.0.1:9000 minioadmin minioadmin
mc admin info local
```

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+Add TLS Certificates
```
ravindk89, line 156: So since this is a quickstart - in my experience users cannot generate local TLS certs quickly.

If we want to keep this section, we should include simple instructions on generating self-signed certs and what the consequences are.

We have a tool `minio/certgen` that is probably _the_ easiest thing to use here. 

_OR_ we drop this for now and come back to it later.

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+
+Issue the following command to start the :mc:`minio server` with non-default credentials. 
+The table following this command breaks down each portion of the command:
+
+.. code-block:: shell
+   :class: copyable
+
+   export MINIO_ROOT_USER=minio-admin
+   export MINIO_ROOT_PASSWORD=minio-secret-key-CHANGE-ME
+   #export MINIO_SERVER_URL=https://minio.example.net
```
ravindk89, line 176: We can drop this, it's only necessary if TLS is enabled and DNS is actually configured.

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+   * - :envvar:`MINIO_ROOT_PASSWORD`
+     - The corresponding secret key to use for the 
+       :ref:`root <minio-users-root>` user.
+
+       Replace this value with a unique, random, and long string.
+
+   * - :envvar:`MINIO_SERVER_URL`
+     - The URL hostname the MinIO Console uses for connecting to the MinIO server. 
+       This variable is *required* if specifying TLS certificates which **do not** contain the IP address of the MinIO Server host as a :rfc:`Subject Alternative Name <5280#section-4.2.1.6>`. 
+       Specify a hostname covered by one of the TLS certificate SAN entries.
```
ravindk89, line 200: Same thought on tossing this if we don't work through TLS right away.

source/quickstart/osx.rst
```diff
@@ -6,31 +6,213 @@ Quickstart for Mac OSX
...
+
+      Command-line: https://docs.min.io/docs/minio-client-quickstart-guide
+         $ mc alias set myminio http://192.0.2.10:9000 minioadmin minioadmin
+
+      Documentation: https://docs.min.io
+
+      WARNING: Detected default credentials 'minioadmin:minioadmin', we recommend that you change these values with 'MINIO_ROOT_USER' and 'MINIO_ROOT_PASSWORD' environment variables.
+
+#. Access the :ref:`minio-console` by going to a browser (such as Safari) and entering the Console address specified in the :mc:`minio server` command's output.
+   For example, http://127.0.0.1:9001.
```
ravindk89, line 110:
```suggestion
   For example, :guilabel:`Console: http://192.0.2.10:9001 http://127.0.0.1:9001` in the example output indicates two possible addresses to use for connecting to the Console
```

Worth noting that the user _can_ direct their browser to the :9000 port. We will auto-redirect.

For a quickstart it might be sufficient to tell them to open a browser to 127.0.0.1:9000 since we know that will work.

ravindk89 (commented):
Good first pass, lets chat about these after standup today.

source/quickstart/osx.rst
djwfyi: I'm trying to avoid having to manually number things.
It's just a maintenance headache when the steps required inevitably change.

source/quickstart/osx.rst
djwfyi: We do have a link in Next Steps for security.
Maybe just drop it?

PR #458, opened 2022-04-08
DOCS-457: Add mc admin user svcacct reference docs

### LINKED ISSUE #457: Add mc admin user svcaccount command
**Summary**

Add CLI reference for mc admin user svcaccount to the reference docs. Update the user-management page to call out that it's available.

ravindk89:
Closes #457 

#. Summary

- Adds the `mc admin user svcaccount` docs to the web CLI reference
- Fixes a bug in rendering output for the `mc admin` commands that did not include full output
- Slightly simplified logic for parsing options for `mc-cmd` directives

### DIFF
--- a/source/reference/minio-mc-admin.rst
+++ b/source/reference/minio-mc-admin.rst
@@ -125,6 +125,11 @@ The following table lists :mc-cmd:`mc admin` commands:
           :start-after: start-mc-admin-user-desc
           :end-before: end-mc-admin-user-desc
 
+   * - :mc:`mc admin user`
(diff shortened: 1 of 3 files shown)


vadmeste (commented):
LGTM. I just don't understand why we need to use Console to create console for openid.

harshavardhana:
> LGTM. I just don't understand why we need to use Console to create console for openid.

@vadmeste because there is no way to know the authenticated user unless - we can pass STS to mc generated externally.

And we can't generate a relevant SVC for an OIDC since we don't know the parent user.

PR #466, opened 2022-04-27
Fix several typos in mc mirror docs

jfaldanam:


ravindk89:
@jfaldanam thank you for catching these, your contribution is greatly appreciated.

PR #486, opened 2022-05-20
Match terms with MinIO Console

paseaf:
This PR updates terms to match MinIO Console.
- Relevant page:
[https://docs.min.io/minio/baremetal/console/minio-console.html](https://docs.min.io/minio/baremetal/console/minio-console.html#:~:text=You%20can%20explore,Key%3A%20zuf%2BtfteSlswRu7BJ86wekitnifILbZam1KYY3TG)

- Why:
The document uses `Access/Secret Key`
But MinIO console uses `Username` and `Password`.<br>
Maybe this PR could make the document even more intuitive.
image

### DIFF
--- a/source/console/minio-console.rst
+++ b/source/console/minio-console.rst
@@ -30,8 +30,8 @@ standalone MinIO Console using the instructions in the
 You can explore the Console using https://play.min.io:9443. Log in with
 the following credentials:
 
-- Access Key: ``Q3AM3UQ867SPQQA43P2F``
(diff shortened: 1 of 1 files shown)


ravindk89 (approved):
Good catch.

This is a fun one - from an S3 perspective, it's always an `access_key` and `secret_key`.

The Console provides a nice UI, so we can use `Username` and `Password` since those are pretty well accepted inputs.

But in the docs I tend to think access/secret key, thus this disconnect.

Thank you @paseaf for pushing this PR up for us - I'll merge it in.

Issue #465, opened 2022-04-26
Create a Sphinx Glossary for terms


djwfyi:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
Sphinx supports a glossary, as described at

- https://sublime-and-sphinx-guide.readthedocs.io/en/latest/glossary.html
- https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-glossary

Steps to implement:

1. Create a file to hold the glossary directive and contents, such as `glossary.rst`
2. Add terms to the glossary file
3. Create links to the terms by adding the `:term: <word>` directive to docs content

djwfyi:
Closed by PR #503 .

PR #503, opened 2022-07-06
Add Sphinx Glossary

djwfyi:
This PR adds a glossary to the docs site.
One scss file is changed to update the internal xref style to use the standard link color.

This PR does **not** apply `:term:` links throughout the docs to link to the glossary.
That will be done in a later pass.

The glossary is included as one of the **Reference** docs in the navigation panel.

Makes progress on #465.


djwfyi:
Merged first and then squashed.
All good.

PR #506, opened 2022-07-07
Update file on URL

cniackz:
In https://docs.min.io/minio/baremetal/installation/deploy-minio-distributed.html 
we can't download files because they does not exist:

https://dl.min.io/server/minio/release/linux-amd64/minio-20220704210254.0.0.x86_64.rpm

```
Not Found
```

We need to update these files, so we can download them.


ravindk89:
Note to @cniackz - we have a Makefile in which `make sync-minio-version` synchronizes the latest stable using the same resource the downloads page uses, so that is our preference in making this particular system work

PR #507, opened 2022-07-07
Fixing Build to use archived download link

ravindk89:
# Summary

Doing a few things here:

- dropping conf.py from git since it's generated from default-conf.py
- Adding a sed statement to ensure we use archive URLs so they don't drop off on new releases

This should also ensure that we can properly do chron jobs and just rebuild w/ make sync-minio-version

**NOTE** the DOTNET API docs are throwing tons of MYST errors. Recommend not running sync-minio-deps until those are resolved.


cniackz (approved):
LGTM

Tested locally:

```
$ cat /tmp/downloads-minio.json | jq '.Linux."MinIO Server".amd64.DEB.download' | sed "s|linux-amd64|linux-amd64/archive|g"
"https://dl.min.io/server/minio/release/linux-amd64/archive/minio_20220706202949.0.0_amd64.deb"
```
My only concern was about other vars like:

* subnet-short
* minio-tag
* minio-latest

But I see they are declared somewhere else: `./source/default-conf.py`

So hopefully by removing `source/conf.py` nothing is going to be broken.

PR #518, opened 2022-07-29
DOCS-IA: Platformize SSE Docs. Hashicorp Pass

ravindk89:
# Summary

As part of containerization efforts, this initial pass attempts to create a structure for the SSE docs.

Some of the work here will also inform how to break up other tutorials that involve deploying or modifying an existing deployment of MinIO.

# General notes

I'm standardizing on this kind of approach for platformized includes:

```
/includes/common/common-<file> # Anything generally common to a topic
/includes/<platform>/common-<file> # Anything common to a topic, but specific to the <platform>
/includes/<platform>/steps-<procedure> # Any procedures or steps associated to a tutorial for a <platform>
```

The end result should be that a given `/include/<platform>` folder is all-inclusive of platform-specific content.

We can include as- necessary from there.

The major complication is that debugging is difficult - Sphinx does not throw accurate warnings or errors about the location of a bad include.

The easiest way to debug is to look at the rendered output and check for missing includes or passages. Not ideal, but better than nothing.

To that point, frequent intermediate builds may help catch issues more easily.

# Major Concerns

As I'm working through this, it is feeling a little easier to break up the content. *But* creating a mental map of the files to modify is a challenge for new content. Since we are supporting 5 platforms total (Linux, Windows, macOS, K8s, Containers), modifying or creating a single tutorial can result in 5xN includes worth of content. But that is the trade-off here, at least as long as Sphinx treats included headers as it currently does.

More likely than not we will have to get into a habit of a two-pass system. The first is to establish a baseline of content per-platform. The second is to polish by adding more details, concepts, and steps as necessary. This way we can focus on content out first, then optimizations that might otherwise inflate a PR that already may include several dozen files.


requirements.txt
```diff
@@ -1,7 +1,7 @@
 docutils == 0.17
-sphinx == 4.3.2
-sphinx-copybutton == 0.4.0
-sphinx-design == 0.0.13
+sphinx == 5.1.1
```
ravindk89, line 2: Might drop this back.

Sphinx 5.1 wants DocUtils 0.20.0, but MyST wants 0.19.0

I'm not seeing anything we can use in newer versions of sphinx, it mostly seems autodoc related. Might get us up to 4.4.0 and DocUtils 0.19.0 and freeze it there.

Also need to potentially fork DocUtils 0.19.0 and patch the code for the Include directive to return better errors around missing keys.

source/includes/macos/steps-configure-minio-kes-hashicorp.rst
```diff
@@ -0,0 +1,135 @@
...
+- Deploy a |KES| server configured to use `Hashicorp Vault <https://www.vaultproject.io/>`__ as the root |KMS|.
+
+- Create a new |EK| on Vault for use with |SSE|.
+
+- Deploy a MinIO server configured to use the |KES| container for supporting |SSE|.
+
+- Configure automatic bucket-default :ref:`SSE-KMS <minio-encryption-sse-kms>`.
+
+For production environments, MinIO recommends using Linux hosts. 
+See the MinIO on Linux documentation for configuring MinIO with KES and Hashicorp Vault.
```
djwfyi, line 13: Can we add a link here and on line 15?

source/includes/macos/steps-configure-minio-kes-hashicorp.rst
```diff
@@ -0,0 +1,135 @@
...
+
+For production orchestrated environments, use the MinIO Kubernetes Operator to deploy a tenant with |SSE| enabled and configured for use with Hashicorp Vault.
+
+Prerequisites
+-------------
+
+.. _minio-sse-vault-prereq-vault:
+
+Hashicorp Vault
+~~~~~~~~~~~~~~~
```
djwfyi, line 23: Add verbs to subheadings, as mentioned for the other platform pages.

source/operations/server-side-encryption/configure-minio-kes-hashicorp.rst
```diff
@@ -16,279 +16,24 @@ Server-Side Object Encryption with Hashicorp Vault Root KMS
...
-  :ref:`SSE-KMS <minio-encryption-sse-kms>` and 
-  :ref:`SSE-S3 <minio-encryption-sse-s3>`.
+.. cond:: linux
 
-Prerequisites
--------------
+   .. include:: /includes/linux/steps-configure-minio-kes-hashicorp.rst
 
-.. _minio-sse-vault-prereq-vault:
+.. cond:: macos
```
djwfyi, line 33: Are you doing a Windows slice?

djwfyi (changes requested):
Lots of great work went into this.
I think the way you sliced docs is fine.
We will need to be careful about how we build out the includes/ directory as we go forward. Either the individual files or the whole folder will become unwieldy if we don't figure out a good IA for it sooner rather than later. I think this is fine as is structure wise for now, though.

I have a number of grammar/style suggestions or corrections to make and can take another look after you apply those.

source/operations/server-side-encryption/configure-minio-kes-hashicorp.rst
ravindk89: image

source/includes/macos/steps-configure-minio-kes-hashicorp.rst
ravindk89: Done throughout.

Also removing the prereqs that are actually just behaviors.

djwfyi (approved):
LGTM

PR #521, opened 2022-08-05
Clarify Upgrade Procedure

ravindk89:
We got user feedback that there was confusion on our existing upgrade procedure. Specifically around the checklist and its implications.

On discussion and testing, we validated that:

- `mc admin service restart` works with `systemctl` managed systems
- The restart can be coordinated via this command
- The restart is *fast*

This further reinforces that MinIO upgrades are non-disruptive. Amazon SDKs [have built in retry tolerance](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) as do MinIO SDKs (to my knowledge), so the actual user impact during these few seconds is nil. 

This rewrite does some general cleanups of the intro. Some discussion is happening offline w/ AB directly as well.

### DIFF
--- a/source/includes/common-installation.rst
+++ b/source/includes/common-installation.rst
@@ -51,7 +51,7 @@ recommends using RPM or DEB installation routes.
 
 .. start-upgrade-minio-binary-desc
 
-The following tabs provide examples of updating MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary.
+The following tabs provide examples of updating MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary executable.
 
 For infrastructure managed by tools such as Ansible or Terraform, defer to your internal procedures for updating packages or binaries across multiple managed hosts.
 
--- a/source/includes/common-installation.rst
+++ b/source/includes/common-installation.rst
@@ -51,8 +51,9 @@ recommends using RPM or DEB installation routes.
 
 .. start-upgrade-minio-binary-desc
 
-The following tabs provide examples of updating MinIO onto 64-bit Linux
-operating systems using RPM, DEB, or binary:
+The following tabs provide examples of updating MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary.
+
+For infrastructure managed by tools such as Ansible or Terraform, defer to your internal procedures for updating packages or binaries across multiple managed hosts.
 
 .. tab-set::
 
@@ -66,7 +67,7 @@ operating systems using RPM, DEB, or binary:
          :class: copyable
          :substitutions:
 
-         wget |minio-rpm| -O minio.deb
+         curl |minio-rpm| --output minio.rpm
          sudo dnf update minio.rpm
 
    .. tab-item:: DEB (Debian/Ubuntu)
@@ -79,7 +80,7 @@ operating systems using RPM, DEB, or binary:
          :class: copyable
          :substitutions:
 
-         wget |minio-deb| -O minio.deb
+         curl |minio-deb| --output minio.deb
          sudo dpkg -i minio.deb
 
    .. tab-item:: Binary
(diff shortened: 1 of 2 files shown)


harshavardhana (approved):

source/installation/upgrade-minio.rst
```diff
@@ -10,67 +10,58 @@ Upgrade a MinIO Deployment
...
-   MinIO **strongly discourages** performing blind updates to production
-   clusters.  You should *always* test upgrades in a lower environment
-   (dev/QA/staging) *before* applying upgrades to production deployments.
+This procedure does **not** require taking downtime and is non-disruptive to ongoing operations.
 
-   Exercise particular caution if upgrading to a :minio-git:`release
-   <minio/releases>` that has backwards breaking changes. MinIO includes
-   warnings in release notes for any version known to not support
-   downgrades.
+This page documents methods for upgrading using the update-than-restart method for both ``systemctl`` and user-managed MinIO deployments.
```
djwfyi, line 20: > update-th**e**n-restart

It needs to be "then", not "than"

source/installation/upgrade-minio.rst
```diff
@@ -10,67 +10,58 @@ Upgrade a MinIO Deployment
...
-   * - Test in Lower Environments
-     - Test all upgrades in a lower environment such as a dedicated 
-       testing, development, or QA deployment.
+MinIO's upgrade-then-restart procedure does *not* require taking downtime or scheduling a maintenance period.
+MinIO restarts are fast, such that restarting all server processes in parallel typically completes in a few seconds. 
+MinIO operations are atomic and strictly consistent, such that applications using MinIO or S3 SDKs can rely on the built-in :aws-docs:`transparent retry <general/latest/gr/api-retries.html>` without further client-side logic.
+This ensures upgrades are non-disruptive to ongoing operations.
 
-       **Never** perform blind upgrades on production deployments.
+"Rolling" or serial "one-at-a-time" upgrade methods do not provide any advantage over the recommended "parallel" procedure, and can introduce unnecessary complexity to the upgrade procedure.
```
djwfyi, line 42: I'd suggest making this it's own consideration and then having the procedure below the standard procedure, with a link to it here.

source/installation/upgrade-minio.rst
```diff
@@ -10,67 +10,58 @@ Upgrade a MinIO Deployment
...
 Update ``systemctl``-Managed MinIO Deployments
 ----------------------------------------------
 
-Deployments managed using ``systemctl``, such as those created
-using the MinIO :ref:`DEB/RPM packages <deploy-minio-distributed-baremetal>`,
-require manual update and simultaneous restart of all nodes in the
-MinIO deployment.
+Use these steps to upgrade a MinIO deployment where the MinIO server process is managed by ``systemctl``, such as those created using the MinIO :ref:`DEB/RPM packages <deploy-minio-distributed-baremetal>`.
 
 1. **Update the MinIO Binary on Each Node**
```
djwfyi, line 66: We don't typically bold procedure steps. I wouldn't change the normal style just for this procedure.
Let's let our CSS handle that.

djwfyi (changes requested):
- Content is fine.
- There are a couple of grammar/spelling things.
- We shouldn't be applying formatting to procedure steps when we have CSS in place.
- And I think the rolling update procedure in the considerations section should move to its own procedure at the end of the doc.

source/includes/common-installation.rst
```diff
@@ -51,8 +51,9 @@ recommends using RPM or DEB installation routes.
 
 .. start-upgrade-minio-binary-desc
 
-The following tabs provide examples of updating MinIO onto 64-bit Linux
-operating systems using RPM, DEB, or binary:
+The following tabs provide examples of updating MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary.
```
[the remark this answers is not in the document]
ravindk89: Added, but I think binary executable is the term vs file.

source/installation/upgrade-minio.rst
ravindk89: Based on discussions w/ AB, we don't want to draw more attention to this than absolutely necessary. Plus once I pull this into the platformization docs, this section only applies to the Container platform - for all other platforms, this is not a required block.

For now leaving as-is.

source/installation/upgrade-minio.rst
ravindk89: Done

djwfyi (approved):
LGTM

PR #522, opened 2022-08-09
Update theme as per new design specs

rushenn:
Fixes https://github.com/miniohq/min.io/issues/1552

**Preview**

Screenshot 2022-08-09 at 20 42 13
Screenshot 2022-08-09 at 20 42 19


ravindk89:
Will review @rushenn - it looks like a handful of commits from `master` got pulled in here as well. I think if you do:

```
git rebase -i HEAD~15
```

And drop all but your and kaans fixes, it should be OK?

ravindk89:
Alternatively I can resync master and docs-ia, and that might also help here. Will try that today.

rushenn:
@ravindk89 I'm rebasing the changes currently. Please do not review now. Once rebased, I will let you know. Thank you.

ravindk89:
Sure thing!

rushenn:
@ravindk89 The PR is ready for review. Please take a look when you get a chance.

ravindk89:
Sure - very excited to try this out

harshavardhana (approved):

djwfyi:
@rushenn We're getting duplicate copy anchor location icons when hovering over a heading.
image

djwfyi:
@rushenn We need to be able to distinguish clickable monospace from what's not clickable.
So in this screenshot, `mc rb` is a clickable link, but `rm --rf` is not. We need some way to indicate that the `mc ...` are links. Maybe the same font color as other `<a>` tags? Perhaps a different background color for those?

image

djwfyi:
1. On smaller breakpoints, portions of the doc navigation becomes inaccessible. It doesn't wrap, and there's no scroll bar to move horizontally. I'm not sure the solution.

2. We are losing the ability to navigate to other parts of the min.io site on the medium and small breakpoints. Since docs is moving to min.io/docs, it seems like we should keep the top nav accessible on mobile devices. The hamburger menu gives the docs table of contents. At that location on the top left, I would expect the navigation to be for the min.io top nav.

   Doc TOC should be in a lower hamburger, above where the page level drop down "Table of Contents" appears in this screenshot. 

3. Also, the magnifying glass icon is missing on the button to access the search function at the smallest breakpoint. (To the right of the table of contents dropdown in the screenshot.)

image

djwfyi (changes requested):
Left comments separately on PR for requested changes.

rushenn:
@djwfyi Thank you for your detailed review and feedback. I have addressed most of the review comments except the doc horizontal scrolling. PTAL. 

> On smaller breakpoints, portions of the doc navigation becomes inaccessible. It doesn't wrap, and there's no scroll bar to move horizontally. I'm not sure the solution.

As per the current design, I believe we'll have to use the horizontal mouse/touchpad scroller in order to navigate in a resized window on the desktop. Normally, we can consider anything below ~1190px x-resolution as a mobile device and we can use the touch gestures to navigate the overflowing scrollable contents. 

We can also display chevron icons on each end and use them as click points to scroll horizontally as a fix for this scenario. 

Let us know your thoughts as well @oscarocastellanos

oscarocastellanos:
@rushenn @djwfyi Correct, Rushan. In desktop, we would prefer to use the horizontal scroller native to the browser. On mobile, however, the idea was to use gestures exactly as you described. We would need to implement some kind of fade out effect to convey this interaction accordingly: 
Captura de Pantalla 2022-08-10 a la(s) 15 43 52

harshavardhana (approved):

rushenn:
@oscarocastellanos Added fade ends to the nav bars. 

Screenshot 2022-08-11 at 20 28 32
Screenshot 2022-08-11 at 20 28 47

oscarocastellanos:
Fantastic @rushenn !

djwfyi (approved):
Thanks for the fixes.
This looks great!

rushenn:
@ravindk89 @djwfyi Copy button is not available on most of the code blocks. Is this behaviour expected?

harshavardhana:
> @ravindk89 @djwfyi Copy button is not available on most of the code blocks. Is this behaviour expected?

Should we take this live and work on this right after?

rushenn:
> > @ravindk89 @djwfyi Copy button is not available on most of the code blocks. Is this behaviour expected?
> 
> Should we take this live and work on this right after?

Yes, we can address this with a separate PR.

djwfyi:
> @ravindk89 @djwfyi Copy button is not available on most of the code blocks. Is this behaviour expected?

Yes, this is expected. We have to declare  a code block as copyable when we write it.
As long as it is appearing on some (and it is), then we are good.

ravindk89:
Agreed - @rushenn can you squash down to a single commit, and then we will merge? I'd use git's tool here but since this is your branch it might be safer that way.

rushenn:
@ravindk89 Done, PTAL.

PR #528, opened 2022-08-18
Integrating .. only logic into .. cond

ravindk89:
So after some exploration and guidance from Sphinx-familiar folks, i was able to come up with a modified Conditional extension that takes the `.. only` behavior from Sphinx and combines it with the proper discard-if-false logic gate of the `.. cond` directive, which is derived from work done by the MongoDB Docs Tools team.

The end result is, from what I can tell, correct parsing of nested headings and insertion into the TOC tree at the correct header levels.

If the tag does not match, the entire nested node is simply ignored, which saves on build cycles and avoids potentially weird behavior like errors due to missing files (that are excluded entirely from the build). 

If the tag does match, and if the included content has header decorations, the logic matches those decorations against the "parent" file and attempts to figure out how far "down" it needs to be. 

For this reason, our method of header descriptions *must* be consistent. Based on what we have so far:

```
========
Title
========

H1 Header
--------------

Content

H2 Header
~~~~~~~~

Content

H3 Header
++++++++

Content

H4 Header
^^^^^^^^^^
Content
```

@balamurugana Most of this code is inherited from the Sphinx project, but if you see something clearly wrong from a Python-y point of view please let me know.

### DIFF
--- a/sphinxext/cond.py
+++ b/sphinxext/cond.py
@@ -1,8 +1,22 @@
-from docutils import nodes
-from docutils.parsers.rst import Directive, directives
-from docutils.utils import SafeString
-from sphinx.util.nodes import nested_parse_with_titles, set_source_info
+"""
 
+   Original Cond logic derived from https://github.com/mongodb/docs-tools/blob/master/sphinxext/fixed_only.py
+
+   Original Only logic derived from https://github.com/sphinx-doc/sphinx/blob/5.x/sphinx/directives/other.py#L289, under the BSD license
+
+   
+
+"""
+
+from docutils import nodes
+from docutils.nodes import Element, Node
+from docutils.parsers.rst import Directive, directives
+from sphinx.util.docutils import SphinxDirective
+from docutils.utils import SafeString
+from sphinx.util.nodes import nested_parse_with_titles, set_source_info, Node
+from typing import TYPE_CHECKING, Any, Dict, List, cast
+
+# Keeping this in case we need it later, but assuming everything works as expected, we can probably drop this entirely
 
 class Cond(Directive):
     """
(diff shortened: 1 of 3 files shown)


sphinxext/cond.py
```diff
@@ -29,9 +43,88 @@ def run(self):
...
+    option_spec = {}
+
+    def run(self):
+        config = self.state.document.settings.env.config
+        try:
+            result = config._raw_config['tags'].eval_condition(self.arguments[0])
+        except ValueError as err:
+            raise self.severe(u'Error parsing conditional parsing expression: {}'.format(SafeString(str(err))))
+
+        if result:
```
balamurugana, line 63: You could return quickly here and reduce indentation level like
```py
if not result:
    return []
```

sphinxext/cond.py
```diff
@@ -29,9 +43,88 @@ def run(self):
...
+
+               current_depth = 0
+               parent = self.state.parent
+               while parent:
+                  current_depth +=1
+                  parent = parent.parent
+                  # This should stop at the "Title" element
+
+               # Accounts for the "Title" element?
+               current_depth -= 2
```
balamurugana, line 94: This value may go negative and will lead to unexpected behavior. It seems we could simplify here.

ravindk89: I'm inclined to agree, but also am only sort of following how the depth-pop logic works.

If I have something nested under an H1, I would expect to get up to `current_depth=1`. Then we do -2 (making it -1), subtract another 1 later (making it -2) and add 1 (making it -1). So yeah that seems...wrong.

Will test a little more and see what happens. I am guessing we can simply state that if `current_depth` is <2, do nothing?

Or if it works _in practice_ there might be something about how Sphinx/Docutils treats the node tree where this value always ends up positive (somehow).

ravindk89: After some testing I can't seem to trigger the case where this happens.

I've added a simple print statement if the depth ever drops to be negative - that way if it *does* happen we can identify the circumstances in which it happens and correct for it. As it stands I don't know enough about why the `-=2` logic is included to remove or correct for a negative value.

PR #539, opened 2022-08-24
Completion of SSE-KMS doc rewrite

ravindk89:
# Summary

With the addition of the updated Conditional logic, we can significantly streamline the conversion process.

Initial PR commit is to show how the conditionals work w/ both in-page conditional headers and included conditional headers.

Further work in this PR will be on the remaining KMS solutions.


djwfyi (approved):
LGTM

source/includes/container/common-minio-kes.rst
```diff
@@ -69,33 +69,27 @@ The following commands create two TLS certificates that expire within 30 days of
     :class: copyable
     :substitutions:
 
+    # These commands output keys to |path|/certs on the host operating system
```
djwfyi, line 72: `|path|` needs to be corrected. It is not replacing correctly in the rendered file.

djwfyi (approved):
One replacement that needs to be corrected.

Consider updating code blocks that have `\` in multi-line to just two spaces and then the `\`. The replacements are causing havoc with trying to line them up.

djwfyi:
Approved / LGTM with one fix and an optional suggestion in my review.

source/includes/container/common-minio-kes.rst
ravindk89: Done

Also: `\s*\\` <- regex to replace all of those kinds of code-block style things. Fun.

Issue #492, opened 2022-05-31
Post-IA Update Additions and Cleanup


djwfyi:
# Summary

These are a list of tasks that need to be completed post-IA conversion to fill out empty or blank topics or sections

## Goals

List the in-scope goals

### Operations Topics

- [x] Operations/Installation
- - [x] Need to update the Standalone Single-Drive section for the new Single Drive features
- [x] (Daryl) Operations/Concepts
- - [x] (Daryl) Contains a heading for Erasure Coding that is incomplete. Note, there is a separate page on Erasure Coding. 
- - [x] (Daryl) Needs per-platform chunking
- [x] Operations/Installation/Decommission Pool
- - [x]  First sentence has "a server pools"
- - [x] Link to the calculation that shows calculation of free space
- [x] (Ravind) All the Data Encryption tutorials needs to point to the Operator Console for setup
- [x] (Ravind) Operations/Network-Encryption
- - [x] Page shows Linux, OSX, and Windows instructions in tabs. Need to conditionalize these for the per platform IA.
- [x] Operations/Deployment Checklists
- - [x] /Hardware Checklists is a stub
- - [x] /Security Checklist is a stub
- - [x] /Software Checklist is a stub
- [x] Operations/Data-recovery/recover-after-site-failure is a stub
- [x] Operations/Troubleshooting is a stub
- [x] Operations/* needs K8s platform conditionals
- - [x] operations/tls needs to focus on Operator Console TLS management
- - [x] external identity management should focus on Operator Console, K8s specifics
- - [x] Monitoring needs Operator Console + Tenant Prometheus mentions 
- - [x] (Daryl) Add CLI and GUI instructions for creating a K8s MinIO tenant

### Administration Topics

- [x] (Daryl) Consider having a "Core Administration Concepts" to parallel the "Core Operational Concepts"
- [x] Administration/Identify Access Management/Minio Identity Management - unhide the dropdown at the bottom, turn into a standard note
- [x] Modify `MinIO Console` page for K8s to be `MinIO Tenant Console`
- [x] Modify OpenID and Active Directory pages to reference K8s lingo for K8s build
- [x] Modify tutorials to be Console first, CLI second
- [x] Add K8s-appropriate language where necessary

## Developers

- [x] developers/minio-drivers - need download links
- [x] General tidy-up of origin markdown files (handled via SDK Git Repos

## Reference

- [x] Operator CLI reference docs need to be viewable
- [x] Reorganize mc reference to be less flat

## Non-Goals

[...]

PR #540, opened 2022-08-26
Adding site recovery content

djwfyi:
Fills out content for the site recovery stub.
Partially addresses #492 .

Also minor fixes for spelling and formatting in the site-replication doc.


source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
 
 .. default-domain:: minio
 
 .. contents:: Table of Contents
    :local:
    :depth: 1
 
-ToDo- site replication recovery procedure
+MinIO can make the loss of an entire site, while significant, a relatively minor incident.
+Site recovery depends on the backup option you use for the site.
```
ravindk89, line 14: I understand the intent here, but the second sentence loses me since we're talking about 'backup options' - which many users will assume is traditional backup, and not replication setup.

Since we will have around 3 distinct types of replication a user could configure, we should focus on that and briefly summarize (maybe as a table) what kinds of site failures each replication type can work with.

For example:

- Site Replication : total restoration from healthy peer site
- Bucket Replication : data restoration (objects only) per configured bucket from healthy remote site
- Other REplication; Whatever limitations it might have

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
+
+Site Replication
+----------------
+
+:ref:`Site replication <minio-site-replication-overview>` keeps two or more MinIO deployments in sync.
```
ravindk89, line 19: Can we be more specific? It keeps IAM, buckets, bucket configurations, and objects synchronized between sites.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
+If you suffer complete site loss with site replication enabled, you can restore IAM policies, objects, metadata, and other :ref:`replicable data <minio-site-replication-what-replicates>` from the remaining site(s).
```
ravindk89, line 20: 'site loss with site replication enabled' is a little clunky. Spitballing:

"If a peer site completely fails, such as due to a major disaster or long power outage, you can use the remaining healthy peer sites to restore..."

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
+
+1. Deploy a new MinIO site using the same ``root`` credentials as used on other deployments in the site replication configuration
```
ravindk89, line 22: We could mention here that the new site can use the original hardware, but will require wiping all data (I assume). Maybe @poornas can comment on what would be required for something like a peer site having partial data loss (e.g. a span of drives failed simultaneously and were repaired, so only X% of data is actually lost).

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
+3. On the site with data, create a new ``remote target`` using the :mc-cmd:`mc admin bucket remote add` command and record the ARN from the output
+4. From the site with the data, use the :mc-cmd:`mc replicate resync` command with the ARN from the previous command to rebuild the on the new site
+5. Wait for re-synchronization to complete (us :mc-cmd:`mc replicate resync status` to check)
+6. Set up bucket replication rule(s) from the new MinIO site to the existing target bucket(s)
+7. `(Optional)` Delete the bucket replication rules from the target deployment(s) to restore an active-passive replication scenario
+
+Bucket Replication
+------------------
+
+Alternatively, use :ref:`bucket replication <minio-bucket-replication>` directly to restore the site contents by performing a replication from the target bucket(s) to a new MinIO site.
```
ravindk89, line 46: Weird to start with Alternatively since this is a section header. Alternatively to what? We lose the context here.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,76 @@
...
+
+   If you prefer to keep an active-active replication between the buckets, do not delete the rules on the target bucket.
+   In active-active replication, changes to the objects at either location affect the objects at the other location.
+
+Mirroring
+---------
+
+MinIO's mirroring copies an object from any S3 compatible storage system.
+
+Mirroring does not provide an object's version ID, versions, or other metadata, even if copying from another MinIO deployment.
```
ravindk89, line 69:
```suggestion
Mirroring only copies the latest version of each object and does not include versioning metadata, regardless of the source.
```

ravindk89 (changes requested):
A few things I think we can tighten up.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: That's a really good call out. Fixed.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: I did try to cover this with the replicable data link in the next sentence, but I moved things around.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: Agreed.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍 Added.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: It's an alternative to the resync process above. Removed.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,89 @@
...
+.. list-table:: 
+   :widths: 25 75
+   :width: 100%
+
+   * - Site Replication
+     - Total restoration from the healthy peer site(s)
+   * - Bucket Replication
+     - Data restoration of objects and metadata from a healthy remote location for each bucket configured for replication
+   * - :mc-cmd:`mc mirror`
+     - Data restoration of objects only (no metadata) from a healthy remote location
```
poornas, line 25: it is inaccurate to say `no metadata` is restored in `mc mirror` - it will restore other metadata except VersionID, ModTime. Difference b/w bucket replication and `mc mirror` is that replication preserves version history and all metadata whereas mirror is more useful for unversioned buckets OR where they care only to restore the latest copy

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,89 @@
...
+
+1. Deploy a new MinIO site using the same ``root`` credentials as used on other deployments in the site replication configuration
+
+   You can use the original hardware, if still available, but you must first wipe any remaining data before creating the new site.
+2. Configure the new site with the same Identity Provider (IDp) as the other sites
+3. :ref:`Expand the existing site replication <minio-expand-site-replication>` by adding the newly deployed site
+4. Remove the lost site from the site replication configuration
+
+Site replication healing automatically adds the buckets and objects from the existing site(s) to the new site with no further action.
+
```
poornas, line 41: Above section is correct if site replication is in place prior to the disaster. However, there are situations where bucket replication may have been in place but customer wants to switch to site replication after disaster to take advantage of the automatic healing in site replication.

If previously bucket replication was in place on the sites, and after the disaster site replication is being brought in to recover the unhealthy site, all previous bucket replication configuration needs to be removed first on the healthy site - otherwise references to stale targets in the replication config can interfere with site replication.

poornas (changes requested):

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: I appreciate the clarification and correction. 
Removed the `(no metadata)` from the sentence. I think the Mirroring section at the bottom of the page adequately addresses the use cases.

source/operations/data-recovery/recover-after-site-failure.rst
ravindk89: Sounds like we might need a dedicated page for migrating from bucket to site replication.

For here would it be enough (I think, if I'm reading this correctly) to note that:

- You cannot configure site replication for recovery if any bucket replication rules remain in place

- Bucket replication is mutually exclusive with site replication

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
+   You can use the original hardware, if still available, but you must first wipe any remaining data before creating the new site.
+2. Configure the new site with the same Identity Provider (IDp) as the other sites
+3. :ref:`Expand the existing site replication <minio-expand-site-replication>` by adding the newly deployed site
+4. Remove the lost site from the site replication configuration
+
+Site replication healing automatically adds IAM settings, buckets, bucket configurations, and objects from the existing site(s) to the new site with no further action required.
+
+If you are switching from using bucket replication to using site replication, you must first remove all bucket replication rules from the healthy site prior to setting up site replication.
+
+Replication Resync
```
ravindk89, line 46: Rename to 'Bucket Replication' to make it findable

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
+------------------
+
+For scenarios where :ref:`bucket replication <minio-bucket-replication>` was in place prior to the failure, you can use :mc-cmd:`mc replicate resync` to restore data to a new site.
+Create a new site to replace the failed deployment, then synchronize the data from an existing bucket to the new site.
```
ravindk89, line 50: Tweak: from an existing replication-enabled to the new site

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
+
+1. Deploy a new MinIO site
+2. Set up IAM and users as needed
+3. On the site with data, create a new ``remote target`` using the :mc-cmd:`mc admin bucket remote add` command and record the ARN from the output
+4. From the site with the data, use the :mc-cmd:`mc replicate resync` command with the ARN from the previous command to rebuild the on the new site
+5. Wait for re-synchronization to complete (us :mc-cmd:`mc replicate resync status` to check)
+6. Set up bucket replication rule(s) from the new MinIO site to the existing target bucket(s)
+7. `(Optional)` Delete the bucket replication rules from the target deployment(s) to restore an active-passive replication scenario
+
+Bucket Replication
```
ravindk89, line 60: Maybe rename this to "Passive Bucket Replication Resynchronization" , while the previous one is "Active". 

I'm still not sure of the value-add here given the speed consideration, but there might indeed be a BC/DR scenario where the downed site is non-critical and simply needs to recover at some point.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
+------------------
+
+:ref:`Bucket replication <minio-bucket-replication>` can directly restore the site contents by performing a replication from the target bucket(s) to a new MinIO site.
+
+As a passive process, bucket replication may not perform as quickly as desired for a site recovery scenario.
+In such cases where performance is needed, use :mc-cmd:`mc replicate resync` described above instead.
```
ravindk89, line 66: I think we're backing into the point here a little bit.

Using Bucket Replication in this way (post-disaster recovery) uses the standard replication scanner queue, which does not take priority over any other process (iirc). For recovery procedures with stricter SLA/SLO, use the active resync procedure.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
+MinIO can restore the object with all of these attributes to a new MinIO site if bucket replication had already been in use prior to the site loss.
+
+1. Deploy a new MinIO site
+2. Set up IAM and users as needed
+3. On the remaining target bucket deployment(s), create bucket replication rule(s) for each bucket to the new MinIO site
+4. Wait for replication to complete
+5. Set up bucket replication rule(s) from the new MinIO site to the existing target bucket(s)
+6. `(Optional)` Delete the bucket replication rules from the target deployment(s) to restore an active-passive replication scenario
+
+   If you prefer to keep an active-active replication between the buckets, do not delete the rules on the target bucket.
```
ravindk89, line 78: If it's a new site, these wont exist anyways. It's more I think that if they previously had an active-active configuration, they will need to re-create those rules on the new site.

ravindk89 (changes requested):
A few more nits but I think we're close here.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,95 @@
...
-ToDo- site replication recovery procedure
+MinIO can make the loss of an entire site, while significant, a relatively minor incident.
+Site recovery depends on the replication option you use for the site.
+
+.. list-table:: 
+   :widths: 25 75
+   :width: 100%
+
+   * - Site Replication
+     - Total restoration from the healthy peer site(s)
```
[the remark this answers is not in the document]
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: Updated heading per the suggestion below.
Added additional context about switching from bucket replication to site replication.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: This is a second paragraph under step 6. I reworked the sentence to hopefully make that more clear.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: I'm adding a new section on the site replication page about switching from bucket replication.

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,100 @@
...
+   * - Bucket Replication
+     - Data restoration of objects and metadata from a healthy remote location for each bucket configured for replication
+   * - :mc-cmd:`mc mirror`
+     - Data restoration of objects only from a healthy remote location with no versioning
+
+Site Replication
+----------------
+
+:ref:`Site replication <minio-site-replication-overview>` keeps two or more MinIO deployments in sync with IAM policies, buckets, bucket configurations, objects, and object metadata.
+If a peer site fails, such as due to a major disaster or long power outage, you can use the remaining healthy site(s) restore the :ref:`replicable data <minio-site-replication-what-replicates>`.
```
poornas, line 31:
```suggestion
If a peer site fails, such as due to a major disaster or long power outage, you can use the remaining healthy site(s) to restore the :ref:`replicable data   @@<minio-site-replication-what-replicates>`.
```

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -1,13 +1,100 @@
...
+Active Bucket Replication Resynchronization
+-------------------------------------------
+
+For scenarios where :ref:`bucket replication <minio-bucket-replication>` was in place prior to the failure, you can use :mc-cmd:`mc replicate resync` to restore data to a new site.
+Create a new site to replace the failed deployment, then synchronize the data from an existing, healthy, bucket replication-enabled deployment to the new site.
+
+1. Deploy a new MinIO site
+2. Set up IAM and users as needed
+3. On the site with data, create a new ``remote target`` using the :mc-cmd:`mc admin bucket remote add` command and record the ARN from the output
+4. From the site with the data, use the :mc-cmd:`mc replicate resync` command with the ARN from the previous command to rebuild the on the new site
```
poornas, line 58:
```suggestion
4. From the site with the data, use the :mc-cmd:`mc replicate resync start` command with the ARN from the previous command to rebuild the bucket on the new site
```

poornas (changes requested):

ravindk89 (approved):
I've got nothing else to add pending resolving Poorna's comments.

we can always polish later with user feedback.

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

source/operations/data-recovery/recover-after-site-failure.rst
djwfyi: 👍

PR #543, opened 2022-09-02
Nginx Config

ravindk89:
This is the initial work for what should be a git-managed set of nginx configs we can include into the Nginx Docs server.

The idea here is we manage redirects in git, and either chron-copy them to a location we can include via the sites-available/<file>, *or* we just include them from the git path (if nginx tolerates that)

I can't really _test_ these right now since they are using live URLs. if we move forward with the soft-launch, we'll have a few opportunities to maybe test a subset of these URLs and see what happens.

Please let me know if there is anything *obviously* wrong here.

ToDo: The Legacy Docs redirects

### DIFF
--- a/redirects/k8s.conf
+++ b/redirects/k8s.conf
@@ -1,0 +1,70 @@
+# This file contains all redirects for the Kubernetes Documentation for MinIO
+# This includes redirecting links from legacy docs and the Docs V1 projects
+
+       # Use this as a template
+    #  location <path> {
+    #        return 301 /docs/k8s/<filepath>.html;
+    #  }
+
+
+        location /minio/baremetal/quickstart/k8s.html {
+             return 301 /docs/k8s/operations/index.html;
+        }
+
+       location /minio/k8s/deployment/deploy-minio-operator.html {
+             return 301 /docs/k8s/operations/installation.html;
+       }
+
+
+       location /minio/k8s/tenant-management/deploy-minio-tenant.html {
+             return 301 /docs/k8s/operations/install-deploy-manage/deploy-minio-tenant.html;
+       }
+
+       location /minio/k8s/core-concepts/core-concepts.html {
+             return /docs/k8s/operations/concepts.html;
+       }
+
+       location /minio/k8s/tenant-management/manage-minio-tenant.html {
+             return /docs/k8s/operations/deploy-manage-tenants.html;
+       }
+
+       location /minio/k8s/tutorials/user-management.html {
+             return /docs/k8s/administration/identity-access-management/minio-user-management.html;
+       }
+
+       location /minio/k8s/tutorials/group-management.html {
+             return /docs/k8s/administration/identity-access-management/minio-group-management.html;
+       }
+
+       location /minio/k8s/tutorials/policy-management.html {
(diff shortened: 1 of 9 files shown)


redirects/linux.conf
```diff
@@ -0,0 +1,344 @@
...
+        # =============================================
+
+
+        location /minio/baremetal/reference/minio-mc.html {
+             return 301 /docs/linux/reference/minio-mc.html;
+        }
+
+        location /minio/baremetal/reference/minio-mc/ {
+             rewrite /minio/baremetal/reference/minio-mc/(.*) /docs/linux/reference/minio-mc/$1 permanent;
+        }
```
ravindk89, line 293: Can't be 100% this works until we have this all in prod, which is risky

redirects/linux.conf
```diff
@@ -0,0 +1,344 @@
...
+
+        location /minio/baremetal/reference/minio-mc-admin.html {
+             return 301 /docs/linux/reference/minio-mc.html;
+        }
+
+        location /minio/baremetal/reference/minio-mc/ {
+             rewrite /minio/baremetal/reference/minio-mc-admin/(.*) /docs/linux/reference/minio-mc-admin/$1 permanent;
+        }
```
ravindk89, line 301: Same issue here.

redirects/k8s.conf
```diff
@@ -0,0 +1,70 @@
...
+# This includes redirecting links from legacy docs and the Docs V1 projects
+
+       # Use this as a template
+    #  location <path> {
+    #        return 301 /docs/k8s/<filepath>.html;
+    #  }
+
+
+        location /minio/baremetal/quickstart/k8s.html {
+             return 301 /docs/k8s/operations/index.html;
```
djwfyi, line 11: Suggest that we go with `/docs/kubernetes/...` for our URL path rather than `/docs/k8s/...`

redirects/linux.conf
```diff
@@ -0,0 +1,344 @@
...
+        location /minio/baremetal/installation/decommission-pool.html {
+             return 301 /docs/linux/operations/install-deploy-manage/decommission-server-pool.html;
+        }        
+
+
+        # Security Redirects from NuDocs 1 to NuDocs 2
+        # ============================================
+
+        location /minio/baremetal/security/IAM/identity-access-management.html {
+             return 301 /docs/linux/administration/identity-access-management.html;
```
djwfyi, line 53: Why redirecting to linux and not kubernetes?
Are we not kubernetes first?

djwfyi (approved):
Redirects look fine overall, but two general questions:

1. I think we should use `kubernetes` in the URLs instead of `k8s`.
2. We are linking into the Linux platform, but I thought we were trying to be Kubernetes forward?

redirects/linux.conf
ravindk89: baremetal is Linux-specific, so those pages should all continue pointing at Linux platform as it's most relevant to that page.

*But* min.io/docs will default to Kubernetes.

redirects/k8s.conf
ravindk89: Agreed! Will make that fixup.

PR #547, opened 2022-09-08
Add extension to create sitemaps

djwfyi:
This PR adds the `sphinx_sitemap` to the list of extensions in default-conf.py.
The extension creates a sitemap.xml file at the root of the generated HTML output at the end of an HTML build.

This allows us to create the sitemap.xml file for each product/platform.
We'll need the website team to add a sitemap-index.xml file at the root of min.io that links to each of the product/platform doc sitemaps.

### DIFF
--- a/sitemap_index.xml
+++ b/sitemap_index.xml
@@ -2,18 +2,18 @@
 
 <sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
    <sitemap>
-      <loc>https://www.min.io/docs/linux/sitemap.xml</loc>
+      <loc>https://www.min.io/docs/minio/linux/sitemap.xml</loc>
    </sitemap>
    <sitemap>
-      <loc>https://www.min.io/docs/windows/sitemap.xml</loc>
+      <loc>https://www.min.io/docs/minio/windows/sitemap.xml</loc>
    </sitemap>
    <sitemap>
-      <loc>https://www.min.io/docs/macos/sitemap.xml</loc>
+      <loc>https://www.min.io/docs/minio/macos/sitemap.xml</loc>
    </sitemap>
    <sitemap>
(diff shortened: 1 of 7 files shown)


sitemap_index.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+   <sitemap>
+      <loc>https://www.min.io/docs/linux/sitemap.xml</loc>
```
ravindk89, line 5: We might as well update these to what we have in #minio-docs pending AB's final sign-off

source/default-conf.py
```diff
@@ -101,8 +101,11 @@
 exclude_patterns = ['includes/*', '*-template.rst']
 
 # template for adding custom exclude paths if necessary for a given tag
+# html_baseurl is used by sphinx_sitemap extension to generate a sitemap.xml for each platform.
+# The sitemaps are combined in a sitemapindex.xml file at the root level.
 
 if tags.has("linux"):
+    html_baseurl = 'https://www.min.io/docs/linux/'
```
ravindk89, line 108: same w.r.t. the expected base URLs.

ravindk89 (approved):
LGTM mod nits

ravindk89 (approved):
LGTM - one thing to double check is that we need the trailing slash on `html_baseurl`. If the sitemap URLs all look fine then you are good to go

source/default-conf.py
djwfyi: Fixed

sitemap_index.xml
djwfyi: Updated

djwfyi:
The example on https://github.com/jdillard/sphinx-sitemap has a trailing slash, so I'm leaving it.

ravindk89:
:shipit:

PR #554, opened 2022-09-12
cleanup makefile and avoid github API requests

harshavardhana:


build-docs.sh
```diff
@@ -0,0 +1,24 @@
...
+make container
+make windows
+make macos
+
+
+sudo mkdir -p /var/www/docs/minio/kubernetes/upstream
+sudo mkdir -p /var/www/docs/minio/linux
+sudo mkdir -p /var/www/docs/minio/windows
+sudo mkdir -p /var/www/docs/minio/macos
+sudo mkdir -p /var/www/docs/minio/container
```
ravindk89, line 16: Request we clean the directory contents, unless we have some other spot where we do that cleanup.

Sphinx has a behavior where if FILE.rst includes CONTENT.rst, and only CONTENT.rst changes, it does not update FILE.rst on build.

Only cleaning the contents of the build directory fixes this.

ravindk89: Maybe adding --remove-destination would do the trick here as it would re-write all files on collision

ravindk89: or -u to update, since source will always be newer by virtue of a newer build.

source/_templates/content-navigation.html
```diff
@@ -1,31 +1,31 @@
 <div class="content__nav">
     <div class="container">
         <nav class="content__nav__inner">
-            <a target="_blank" rel="noreferrer" href="http://borkserv.ddns.net:9000/staging/docs-ia/k8s/html/index.html" class="{{ 'active' if doc_platform == 'kubernetes' }}">
+            <a target="_blank" rel="noreferrer" href="/docs/minio/kubernetes/upstream/index.html" class="{{ 'active' if doc_platform == 'kubernetes' }}">
```
ravindk89, line 4: didn't realize we could do rellinks like this. Nice.

ravindk89 (approved):
General requests, but maybe not necessary depending on what is happening within the server.

PR #555, opened 2022-09-12
Final pass on platformization

ravindk89:


ravindk89:
@djwfyi can you pull this and just confirm that on a fast pass things look OK? these are largely non-material changes

Makefile
```diff
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line, and also
 # from the environment for the first two.
-SPHINXOPTS    ?= -n -j "auto" -w "build.log"
+SPHINXOPTS    ?= -n -w "build.log"
```
harshavardhana, line 6: we should keep this

harshavardhana (changes requested):

Makefile
ravindk89: SO I get a warning from the sitemap builder with this flag enabled, *but* it doesn't seem to impact the actual contents of the file. Not sure - I am reaching out to the maintainer to understand what the concern is.

harshavardhana: Without this it is not parallelized and slow.

PR #557, opened 2022-09-13
cleanup README.md build instructions

harshavardhana:
- remove Windows, macOS instructions just mandate GNU/Linux for all. GNU/Linux instructions work macOS as well.

- introduce 'make all' to build all targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,8 +15,9 @@ help:
 
 .PHONY: help Makefile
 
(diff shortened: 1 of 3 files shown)


harshavardhana:
Taking this into build requirements for the webhook, README.md changes are more simplification to avoid too many instructions for building many different platforms.

PR #559, opened 2022-09-15
UI enhancements: headerlinks and code tags

rushenn:
- Fix headerlinks on non-heading elements
- Style sidebar code tags:
   Before/After
   code


djwfyi (approved):
LGTM

PR #561, opened 2022-09-15
Updating base URLs for Algolia crawler and intersphinx links

djwfyi:
This PR:
- Updates the base URL links to remove the `www` subdomain. This sphinx-sitemap extension adds a canonical URL on each page with the base URL, but min.io redirects all www.min.io traffic to min.io (without www subdomain). This was causing a conflict with the Algolia crawler as the resulting URL did not match the canonical URL defined on each page.
- Updates the intersphinx links to use the linux and kubernetes/upstream folders to clear intersphinx errors happening during builds.

### DIFF
--- a/source/default-conf.py
+++ b/source/default-conf.py
@@ -105,7 +105,7 @@ exclude_patterns = ['includes/*', '*-template.rst']
 # The sitemaps are combined in a sitemapindex.xml file at the root level.
 
 if tags.has("linux"):
-    html_baseurl = 'https://www.min.io/docs/minio/linux/'
+    html_baseurl = 'https://min.io/docs/minio/linux/'
     excludes = [
         'operations/install-deploy-manage/deploy-minio-tenant.rst',
         'operations/install-deploy-manage/modify-minio-tenant.rst',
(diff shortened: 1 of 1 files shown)


source/default-conf.py
```diff
@@ -284,7 +284,8 @@
 # k8s is temporary until integrating the references here
 
 intersphinx_mapping = {
-    'baremetal': ('https://www.min.io/docs/minio/', None),
+    'baremetal': ('https://www.min.io/docs/minio/linux/', None),
```
ravindk89, line 287: baremetal -> linux

we're mostly leaving 'baremetal' behind.

Where this comes into play is, I think, you can prefix a reference to let Sphinx know where to look.

so

:linux:ref:`foo` should specifically look up the `foo` reference for the `linux` intersphinx inventory.

I think.

ravindk89 (approved):
One fix, otherwise LGTM

source/default-conf.py
djwfyi: Ah, that makes sense. Fixing.

PR #572, opened 2022-09-20
General cleanups for driver synchronization

ravindk89:
A few minor cleanups:

- Removing the SDK markdown docs from git tracking. They're already tracked in the SDK directories, we don't need to do that twice
- Cleaning up the builds, as only Linux needs to build out the SDK docs at this time
- Fixing build-docs.sh to build the SDK docs with Linux (so we get proper version numbers). Means we have to build Linux with SYNC_SDK=TRUE on the server to pull in the SDK docs (otherwise they will all be empty). 

Tested locally and everything seems to work just fine. Will see what happens when it goes live, but I figured I'd ask for @harshavardhana 's thoughts since they know more about the build server script process than I do.


djwfyi (approved):
LGTM

PR #578, opened 2022-09-22
Adds file to list steps needed after adding a new platform

djwfyi:
General cleanups I've noticed after new docs launched.

- Creates a markdown doc for steps needed after creating a new platform for the doc site. 
- Adds min.io sitemap to the sitemap_index file.
- Fixes two links causing some crawler errors in glossary and troubleshooting files.

### DIFF
--- a/new-platform.md
+++ b/new-platform.md
@@ -1,0 +1,49 @@
+# Launching a New Platform
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


ravindk89 (approved):
LGTM. 

I'll add some examples of conditional includes on the design.rst page later as a reference.

PR #579, opened 2022-09-23
Fix doc menu and search being hidden on mobile

rushenn:
We might need to enable TOC on all pages to maintain the original mobile menu design.

### DIFF
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -107,7 +107,7 @@ window.addEventListener("DOMContentLoaded", (event) => {
       tocTargetEl.parentNode.replaceChild(tocOriginalEl, tocTargetEl);
     }
     else {
-      tocAsideEL.style.display = "none";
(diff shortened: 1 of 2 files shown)


djwfyi (approved):
Looks good.

We're still missing the page toc that should be in a right nav on the platform's landing index.html page. But the search and doc site left nav are back.

ravindk89:
I just noticed the missing index toc. That's odd, I wonder why it's just happening on that page.

rushenn:
@ravindk89 @djwfyi There is no TOC available on the index and some other pages. Have we set it disabled in the configs?

ravindk89:
Ah, I see the issue with index.rst

The included conditional file header levels do not match properly with the actual page being included, and that's causing issues.

I can handle that in a separate PR - I would benture if other pages are also missing a TOC, its due to a mismatch in header levels within a conditional include.

PR #582, opened 2022-09-26
UI Enhancements: scroll, component naming, and more

rushenn:
- Fixed the sidebar scroll - with this fix, you'll be able to scroll to the end of the left sidebar at any given screen height. 
- Renamed content-navigation component and CSS class names for better consistency. 
- Changed the footer style by including it as part of the main content area. This change is required to have the scroll fix mentioned at point no. 1.
- Reduced the hero height by a small margin, in order to have more sidebar scroll room on smaller devices such as 13 inch laptop and all.
- Used higher contrast color on search modal's bg and scrollbar thumb to have better color consistency.
- Made scrollbar thumbs visible by default.
- Set sidebar scroll position to match the active doc item.
- Added custom scrollbars for code blocks.

### DIFF
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -1,5 +1,6 @@
 window.addEventListener("DOMContentLoaded", (event) => {
   const tocMenuEl = document.querySelector("#table-of-contents > ul.simple");
+  const root = document.documentElement;
   var readModeLs = localStorage.getItem("read-mode");
 
   // --------------------------------------------------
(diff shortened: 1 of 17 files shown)


djwfyi (changes requested):
The functionality of this works great.
I can now scroll either the left nav or the main content separately, and it's great.

My only request would be that we have some kind of indication that the left nav is scrollable. One suggestion would be just leaving the left nav's scroll bar always visible. But there might be another design choice that could accomplish the same effect.

I'm concerned that users would not realize that they can hover over the left nav to see more content.

While doing that, if we could get the left nav to scroll on load to whatever the current active page is, that would also be really, really nice.

djwfyi:
To clarify what I mean by scrolling on load, right now when opening the MC Admin Client page, it looks like this:
image

The Admin Client is way far down in the left nav, so you can't see where you are.

Instead, it would be great if the left nav loaded and scrolled to the active page so that the user sees something like this when the page loads, with the MC Admin content visible in the left nav:

image

rushenn:
@djwfyi I have addressed your review comments, PTAL.

PR #597, opened 2022-10-04
rename 'mc policy' -> 'mc anonymous' deprecated command

harshavardhana:


source/reference/minio-mc/mc-anonymous-get-json.rst
```diff
@@ -1,7 +1,7 @@
 .. _minio-mc-policy-get-json:
 
 ======================
-``mc policy get-json``
+``mc anonymous get-json``
 ======================
```
ravindk89, line 6: Apparently I can't put suggestions on "deleted" lines. Thanks git.

The top and bottom decorators need to expand enough to cover the full length of the new string. Need to do this across all of these pages.

ravindk89 (approved):
One requested change to prevent build warnings related to title decorations.

PR #600, opened 2022-10-05
Adds mc replicate diff command

### LINKED ISSUE #599: Add information about mc replicate diff
**Summary**

Under Minio docs, information about mc replicate diff command seems to be missing. It  will be good to have that with some examples.

djwfyi:
Closes #599
mc replicate diff was missing. This adds it.

### DIFF
--- a/source/reference/minio-mc/mc-replicate-diff.rst
+++ b/source/reference/minio-mc/mc-replicate-diff.rst
@@ -1,0 +1,145 @@
+.. _minio-mc-replicate-diff:
+
+=====================
+``mc replicate diff``
+=====================
+
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+.. mc:: mc replicate diff
+
+Description
+-----------
+
+.. start-mc-replicate-diff-desc
+
+The :mc:`mc replicate diff` shows a list of unreplicated new or deleted objects.
+
+.. end-mc-replicate-diff-desc
+
+You can list the replication status of objects for a particular remote target.
+To do so, you must have the ARN of the remote target.
+You can use :ref:`retrieve the remote targets configured for a bucket <minio-retrieve-remote-bucket-targets>` to find the ARN.
+
+Syntax
+------
+
+.. tab-set::
+
+   .. tab-item:: EXAMPLE
+
+      The following command shows new or deleted objects in the ``notes`` bucket of the ``teamorange/projects`` prefix on the ``myminio`` alias that have not yet replicated to a specific remote target bucket.
+      The remote target's ARN is ``arn:minio:replication::3bb8c736-4014-42c5-b3cb-d64e3ebaa75e:notes``.
+
+      .. code-block:: shell
+         :class: copyable
+
+         mc replicate diff myminio/notes/teamorange/projects --arn arn:minio:replication::3bb8c736-4014-42c5-b3cb-d64e3ebaa75e:notes
+
+      If any new or deleted objects have not yet replicated, the command outputs something similar to the following:
+
+      .. code-block:: shell
+
+         [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:59 UTC]          478efe49-aa9d-46ab-8268-45b70cc4c341 PUT agenda.docx
+         [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:15 UTC]          b283bf43-319f-455a-a779-3c2e669fad88 PUT budget-meeting.docx
+
+      In the output, ``PUT`` corresponds to a new object.
+      Deleted objects or versions would show ``DEL``.
+
+   .. tab-item:: SYNTAX
+
+      The command has the following syntax:
+
+      .. code-block:: shell
+         :class: copyable
+
+         mc [GLOBALFLAGS] replicate diff   \
+                          [--arn "string"] \
+                          TARGET
+
+      .. include:: /includes/common-minio-mc.rst
+         :start-after: start-minio-syntax
+         :end-before: end-minio-syntax
+
+Parameters
+~~~~~~~~~~
+
+.. mc-cmd:: TARGET
+   :required:
+
+   The path to the alias, prefix, or object.
+
+.. mc-cmd:: arn
+   :optional:
+
+   The ARN of the remote bucket to check for new or deleted objects that have not yet replicated.
+
+   When specified, the command returns a list of any new or deleted objects that have not replicated to the remote target.
+   If not specified, the command returns a list of new or deleted objects on the source deployment that have not replicated to any remote target.
+
+Global Flags
+~~~~~~~~~~~~
+
+.. include:: /includes/common-minio-mc.rst
+   :start-after: start-minio-mc-globals
+   :end-before: end-minio-mc-globals
+
+Examples
+--------
+
+View Unreplicated Versions of Objects at a Prefix
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Display unreplicated ``PUT`` and ``DELETE`` actions for a prefix:
+
+.. code-block:: shell
+   :class: copyable
+
+   mc replicate diff myminio/mybucket/path/to/prefix
+
+- Replace ``myminio/mybucket`` with the :mc-cmd:`~mc replicate add ALIAS` and
+  full bucket path for which to create the replication configuration.
+
+- Replace ``path/to/prefix`` with the prefix or object to use for the request.
+
+If unreplicated objects exist, the output returns a list of the actions that created or removed objects at the prefix that have not replicated to a remote target:
+
+.. code-block:: shell
+
+   [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:59 UTC]          478efe49-aa9d-46ab-8268-45b70cc4c341 PUT agenda.docx
+   [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:15 UTC]          b283bf43-319f-455a-a779-3c2e669fad88 PUT budget-meeting.docx
+
+View Unreplicated Objects at a Specific Remote Target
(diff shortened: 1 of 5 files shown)


source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,113 @@
...
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+.. mc:: mc replicate diff
+
+Description
+------
```
ravindk89, line 16: need to fill this decorator the rest of the way

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,113 @@
...
+
+.. start-mc-replicate-diff-desc
+
+The :mc:`mc replicate diff` shows object versions that are not replicated.
```
ravindk89, line 20:
```suggestion
The :mc:`mc replicate diff` shows object versions that are not replicated to the configured remote target.
```

We can configure multiple replication rules on a bucket right? If we have multiple remote arns, how does this work? Which remote do we check against, since prior to replication we have not yet determined which rule would hit (based on priority) 

Or does this only really work with site replication?

ravindk89 (commented):
A few questions here - I can also test to confirm but I have some confusion on how this works when you have multiple remotes in a bucket.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: fixed

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: `mc replicate` is bucket replication only, not site replication.
When the ARN is specified, the return is the replication status of the target ARN.
When not specified, the return is a general replication status of the bucket.

Added to the `arn` flag.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,116 @@
...
+.. end-mc-replicate-diff-desc
+
+Syntax
+------
+
+.. tab-set::
+
+   .. tab-item:: EXAMPLE
+
+      The following command shows unreplicated objects for the ``teamorange/projects`` prefix in the ``notes`` bucket for the ``myminio`` ALIAS:
```
ravindk89, line 31: Just to confirm - should we just specify `--arn` here so users from the outset understand how to use this?

Chances are they do not intend to get the default status, given that they can get that with `mc replicate info` anyways.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,116 @@
...
+
+.. mc-cmd:: TARGET
+   :required:
+
+   The path to the alias, prefix, or object.
+
+.. mc-cmd:: arn
+   :optional:
+
+   The Amazon Resource Number (ARN) to use for the request.
```
ravindk89, line 64:
```suggestion
   The Amazon Resource Number (ARN) of the remote replication target to use for the diff.
   
   Use :mc-cmd:`mc admin bucket remote ls` to retrieve the configured remote ARNs for the bucket.
```

ravindk89: Logic here is that ARN isn't helpful to the user - what we want them to provide is the remote ARN for their preferred target. We have to lead the user a bit here since we diverge from AWS in terms of how/where to get this .

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,116 @@
...
+
+   When specified, the command returns the replication status of the ARN.
```
ravindk89, line 66: It should return the diff of source against that remote ARN right? vs the diff against any other configured remote on this bucket.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,116 @@
...
+--------
+
+View Unreplicated Versions of Objects at a Prefix
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Display unreplicated versions of objects for a prefix:
+
+.. code-block:: shell
+   :class: copyable
+
```
ravindk89, line 86: Still a little confused here - without --arn , does this not just return the status?

ravindk89: Or does this behavior only matter if you have multiple remotes for a bucket?

ravindk89 (changes requested):
A few more clarifying questions - overall this is fine, I'm just drilling in on functionality here.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: I added a graph at the top of the page with a link to the `mc admin bucket remote` heading that explains how to find lists of remote bucket ARNs.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: Yeah, this was just wrong. It's a list of `PUT` or `DELETE` actions that have not yet replicated. Corrected.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: Corrected throughout. I was misunderstanding what this output.
It's a list of `PUT` and `DELETE` actions that have not yet replicated.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+   :depth: 2
+
+.. mc:: mc replicate diff
+
+Description
+-----------
+
+.. start-mc-replicate-diff-desc
+
+The :mc:`mc replicate diff` shows a list of ``PUT`` or ``DELETE`` actions to objects and object versions that have not replicated to a remote bucket.
```
ravindk89, line 20: This is technically accurate, but I think a little hard to parse. Maybe...

shows a list of unreplicated new or deleted objects for a given remote bucket ?

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+
+.. end-mc-replicate-diff-desc
+
+You can specify a configured remote target ARN to shows ``PUT`` and ``DELETE`` actions that have not yet replicated to the specific remote target.
```
ravindk89, line 24: Clarifying question - this only applies for buckets with multiple ARNs? Do we default to something if unspecified?

If so maybe modify to

For buckets with multiple remote targets, you can specify the target ARN to show only objects not replicated to that remote.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+
+To find the replication status of objects for a particular remote target, you must have the ARN of the remote target.
```
ravindk89, line 26: Actually I'd drop the previous paragraph for just this one.

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+You can use :ref:`retrieve the remote targets configured for a bucket <minio-retrieve-remote-bucket-targets>`.
+
+Syntax
+------
+
+.. tab-set::
+
+   .. tab-item:: EXAMPLE
+
+      The following command shows unreplicated ``PUT`` or ``DELETE`` actions to objects and object versions to the remote target at ARN ``arn:minio:replication::3bb8c736-4014-42c5-b3cb-d64e3ebaa75e:notes`` for the ``teamorange/projects`` prefix in the ``notes`` bucket for the ``myminio`` ALIAS:
```
ravindk89, line 36: Same thought as above

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+
+      .. code-block:: shell
+         :class: copyable
+
+         mc replicate diff myminio/notes/teamorange/projects --arn arn:minio:replication::3bb8c736-4014-42c5-b3cb-d64e3ebaa75e:notes
+
+      If any unreplicated actions exist, the command outputs something similar to the following:
```
ravindk89, line 43: Same thought on actions -> new or deleted objects, somehow

source/reference/minio-mc/mc-replicate-diff.rst
```diff
@@ -0,0 +1,142 @@
...
+
+      .. code-block:: shell
+
+         [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:59 UTC]          478efe49-aa9d-46ab-8268-45b70cc4c341 PUT agenda.docx
+         [0001-01-01 00:00:00 UTC] [2022-10-06 17:18:15 UTC]          b283bf43-319f-455a-a779-3c2e669fad88 PUT budget-meeting.docx
+
```
ravindk89, line 49: Maybe here we can note that `PUT` corresponds to a new object, `DEL` corresponds to a deleted object.

ravindk89 (approved):
This makes a lot more sense. One general thought on actions vs objects, but otherwise LGTM pending however you want to work w/ that.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: Deleted this per the next comment.

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: 👍

source/reference/minio-mc/mc-replicate-diff.rst
djwfyi: Added.

PR #613, opened 2022-10-17
Changes to get Algolia crawler to pull OpenShift docs

djwfyi:
Algolia's crawler is not picking up the OpenShift docs.
This has fixes to get the sitemap to build correctly so the OpenShift docs can be crawled.


ravindk89 (approved):
Nicely resolved.

PR #622, opened 2022-10-27
Adding load balancing content for site replication

### LINKED ISSUE #618: Site Replication Requires Using Load Balancer
**Summary**

https://github.com/minio/minio/issues/15944

As per this discussion:

>  if you configure a node endpoint in site replication, what you are seeing is expected - for proper failover, you should be using a loadbalancer in front of each site and configuring site replication with LB urls.

We need to update the Site Replication docs to explicitly state that our expectation is the configured alias is a load balancer, reverse proxy, or similar network control plane component which automatically routes requests to nodes in the deployment. 

Specifying a single node as the alias results in a single point of failure for replication. Since our docs do call out using a load balancer as standard practice, this should not be entirely unexpected - but the user's experience and expectations is understandable.

Longer term we can try to integrate sidekick into our docs as a quick-and-simple LB for these purposes.

**Goals**

List the in-scope goals

- [x] Update site replication docs to include explicit guidance around setting the alias
- [x] Add warnings that using an alias that binds to a single host risks creating a single point of failure for replication

**Non-Goals**

At least for now, documenting sidekick or a similar LB

**Additional context**

djwfyi:
Closes #618

Adds recommendation for pointing to a load balancer, reverse proxy, or similar for multi-node environments. Explains that this recommendation avoids a single point of failure risk.

### DIFF
--- a/source/operations/install-deploy-manage/multi-site-replication.rst
+++ b/source/operations/install-deploy-manage/multi-site-replication.rst
@@ -110,9 +110,9 @@ You can achieve this with a central KES server or multiple KES servers (say one 
 Load Balancers Installed on Each Multi-Node Site
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When replication to multi-node sites, use the URL or IP address of the site's load balancer, reverse proxy, or similar network control plane component which automatically routes requests to nodes in the deployment.
-
-For such environments, pointing to a specific node within the deployment creates a single point of failure risk for replication.
+.. include:: /includes/common-replication.rst
+   :start-after: start-mc-admin-replicate-load-balancing
+   :end-before: end-mc-admin-replicate-load-balancing
 
 Switch to Site Replication from Bucket Replication
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -137,6 +137,7 @@ Configure Site Replication
 .. tab-set::
 
    .. tab-item:: Console
+      :sync: gui
 
       #. :ref:`Deploy <deploy-minio-distributed>` two or more separate MinIO sites, using the same Identity Provider for each site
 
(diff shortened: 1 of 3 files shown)


source/operations/install-deploy-manage/multi-site-replication.rst
```diff
@@ -108,6 +107,13 @@ For :ref:`SSE-S3 <minio-encryption-sse-s3>` or :ref:`SSE-KMS <minio-encryption-s
 
 You can achieve this with a central KES server or multiple KES servers (say one per site) connected via a central supported :ref:`key vault server <minio-sse>`.
 
+Load Balancers Installed on Each Multi-Node Site
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When replication to multi-node sites, use the URL or IP address of the site's load balancer, reverse proxy, or similar network control plane component which automatically routes requests to nodes in the deployment.
+
+For such environments, pointing to a specific node within the deployment creates a single point of failure risk for replication.
```
ravindk89, line 115: "For such environments" I *think* is referencing the load balancer env which is already optimal.

It's enough to state:

"Using a single node for configuring site replication creates a single point of failure, where that node being offline results in replication failure."

ravindk89: Same throughout. You could pop this into `common-replication.rst` for single sourcing.

ravindk89 (changes requested):
Thought - https://min.io/docs/minio/linux/operations/install-deploy-manage/multi-site-replication.html#configure-site-replication - wherever we talk about endpoints/aliases we should modify to specify "The hostname or IP of the load balancer managing connections to the site" or something to that effect. "

That way we're covered on all angles.

source/operations/install-deploy-manage/multi-site-replication.rst
djwfyi: Good rewording. Thanks.
Also, moved to common-replication as suggested.

ravindk89 (approved):

PR #623, opened 2022-10-28
Attempting to enable github rendering of markdown

ravindk89:
# Summary

The default Markdown parser MyST provides doesn't seem to like the way we've designed our SDK Markdown docs.

This change *should* enable github-strict parsing. It also adds auto-headers to all H1s, which might resolve some of the TOC issues.

This gives us some breathing room to fixup the SDKs where we can still allow for the files to look good in GitHub, but know they will render properly on our Sphinx-built docs.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,3 +8,5 @@ sphinx-sitemap == 2.2.0
 sphinx-togglebutton === 0.3.2
 sphinxcontrib-images === 0.9.4
 myst-parser === 0.18.0
+linkify === 1.4
+linkify-it-py === 2.0.0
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -7,4 +7,6 @@ Sphinx-Substitution-Extensions == 2020.9.30.0
 sphinx-sitemap == 2.2.0
 sphinx-togglebutton === 0.3.2
 sphinxcontrib-images === 0.9.4
-myst-parser === 0.18.0
\ No newline at end of file
+myst-parser === 0.18.0
+linkify === 1.4
+linkify-it-py === 2.0.0
```
ravindk89, line 12: I'll just note - I do not know why `linkify` did not work. `linkify-it-py` *did* seem to work. I'll remove `linkify` in another pass with a bit more testing. But wanted to confirm this works on other machines first.

djwfyi (approved):
LGTM

PR #628, opened 2022-10-31
Add standalone migration procedure

### LINKED ISSUE #624: [FEATURE] Document recommended migration strategy from fs backend
Now that the fs backend has been removed in `RELEASE.2022-10-29T06-21-33Z`, there should be some documented strategy on how users might migrate off the fs backend to a supported backend.

Based on the discussion at https://github.com/minio/minio/discussions/15967, "don't upgrade" is the current documented recommendation, but this isn't a viable strategy unless there will be an automated migration path introduced in the future.

Perhaps the only solution is for users to create a fresh parallel deployment and explicitly copy all the objects over.  Whatever the recommendation is, it should be documented to give users a path to continue to receive updates and security fixes.

djwfyi:
Closes #624

Creates a new page under install-deploy-manage Operator docs. This page summarizes the changes that led to the deprecation of the Gateway/Filesystem. It overviews the steps required to create a new Single-Node, Single-Drive deployment with the contents, settings, and IAM policies from the old deployment.

### DIFF
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -1,0 +1,107 @@
+.. _minio-gateway-migration:
+
+=======================================
+Migrate from Gateway or Filesystem Mode
+=======================================
+
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 1
+
+Background
+----------
+
+The MinIO Gateway entered a feature freeze in July 2020.
+In February 2022, MinIO announced the `deprecation of the MinIO Gateway <https://blog.min.io/deprecation-of-the-minio-gateway/?ref=docs>`__.
+Along with the deprecation announcement, MinIO also announced that the feature would be removed in six months time.
+
+As of :minio-release:`RELEASE.2022-10-29T06-21-33Z`, the MinIO Gateway and the related filesystem mode code have been removed.
+Deployments still using the `standalone` or `filesystem` MinIO modes that upgrade to :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later receive an error when attempting to start MinIO.
+
+Overview
+--------
+
+To upgrade to a :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later, those who were using the `standalone` or `filesystem` deployment modes must create a new :ref:`Single-Node Single-Drive <minio-snsd>` deployment and migrate settings and content to the new deployment.
+
+This document outlines the steps required to successfully launch and migrate to a new deployment.
+
+Procedure
+---------
+
+#. Retrieve the existing deployment's **environment variables**
+
+   Use :mc-cmd:`mc admin config get` with the ``--json`` flag to retrieve a list of the environment variables defined on the existing standalone MinIO deployment.
+
+#. :ref:`Create a new Single-Node Single-Drive MinIO deployment <minio-snsd>`
+
+   The location of the deployment can be any empty folder on the storage medium of your choice.
+   As long as the existing deployment is not to the root of a drive, a new folder on the same drive can work for the new deployment.
+   If the existing standalone system points to the root of the drive, you must use a separate drive for the new deployment.
+
+   Use the environment variable values retrieved from the standalone deployment to establish the same values for the new deployment.
+
+   Set the port to a custom point different than the existing standalone deployment.
+
+#. Duplicate **configurations** from existing standalone deployment to new deployment
+
+
+
+#. Duplicate **buckets** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc ls` with the ``--json`` flag to retrieve a list of the buckets that exist on the standalone deployment.
+   Use the list to recreate the buckets on the new deployment.
+
+#. Duplicate **tiers** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin tier ls` with the ``--json`` flag to retrieve a list of the tiers that exist on the standalone deployment.
+   Use the list to recreate the tiers on the new deployment.
+
+#. Duplicate **policies** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin policy list` with the ``--json`` flag  to retrieve a list of policies that exist on the standalone deployment.
+   Use the list to recreate the policies on the new deployment.
+
+#. Duplicate **groups** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin group list` with the ``--json`` flag to retrieve a list of groups that exist on the standalone deployment.
+   Use the list to recreate the groups on the new deployment.
+
+#. Duplicate **users** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin user list` with the ``--json`` flag to retrieve a list of users with access key, policy name, and status.
+   Use the list to recreate the users on the new deployment.
+
+   Note: You will need to define each user's ``SECRETKEY`` on the new deployment.
+   Make note of the ``SECRETKEY`` for each user to let them know their new credentials.
+
+#. Duplicate **service accounts** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin user svcacct ls` with the ``--json`` flag to list existing service accounts on the standalone deployment.
+   Use the list to recreate the service accounts on the new deployment.
+
+#. Use :mc:`mc mirror` with the ``--watch`` flag on the standalone deployment to move objects to the new |SNSD| deployment
+
+   Do this for each bucket.
+
+#. After the initial mirror process completes, convert standalone deployment to be read only
+
+   Some options for doing this include:
+
+   - ``mc admin policy remove`` for each active policy
+   - ``mc admin policy unset`` to remove policy/policies for each group
+
+#. Wait for ``mc mirror`` to complete for all buckets.
+
+#. Shut down the server for the existing standalone deployment.
+
+   
+
+#. Restart the new MinIO deployment with the ports used for the previous standalone deployment.
+
+
+
+#. Confirm the new deployment works as expected.
+
+   Verify that users and service accounts have access to the buckets and objects as usual.
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -13,7 +13,7 @@ Migrate from Gateway or Filesystem Mode
 Background
 ----------
 
-The MinIO Gateway entered a feature freeze in July 2020.
+The MinIO Gateway and the related filesystem mode entered a feature freeze in July 2020.
 In February 2022, MinIO announced the `deprecation of the MinIO Gateway <https://blog.min.io/deprecation-of-the-minio-gateway/?ref=docs>`__.
 Along with the deprecation announcement, MinIO also announced that the feature would be removed in six months time.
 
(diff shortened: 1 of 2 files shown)


source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,107 @@
...
+To upgrade to a :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later, those who were using the `standalone` or `filesystem` deployment modes must create a new :ref:`Single-Node Single-Drive <minio-snsd>` deployment and migrate settings and content to the new deployment.
+
+This document outlines the steps required to successfully launch and migrate to a new deployment.
+
+Procedure
+---------
+
+#. Retrieve the existing deployment's **environment variables**
+
+   Use :mc-cmd:`mc admin config get` with the ``--json`` flag to retrieve a list of the environment variables defined on the existing standalone MinIO deployment.
```
eco-minio, line 35: This is incorrect as this will only retrieve values that were set by mc, but would not retrieve values from /etc/default/minio or the command line that were not explicitly set via `mc admin config set`. Not sure the best path forward here but it is something like a hybrid of "if you used mc, do this, if you used env vars, do that, if you used a mix, do both"

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,107 @@
...
+#. Duplicate **service accounts** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin user svcacct ls` with the ``--json`` flag to list existing service accounts on the standalone deployment.
+   Use the list to recreate the service accounts on the new deployment.
+
+#. Use :mc:`mc mirror` with the ``--watch`` flag on the standalone deployment to move objects to the new |SNSD| deployment
+
+   Do this for each bucket.
+
+#. After the initial mirror process completes, convert standalone deployment to be read only
```
eco-minio, line 88: Not sure what this provides? Seems like they can just decom at will at that point

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,107 @@
...
+
+   Some options for doing this include:
+
+   - ``mc admin policy remove`` for each active policy
+   - ``mc admin policy unset`` to remove policy/policies for each group
+
+#. Wait for ``mc mirror`` to complete for all buckets.
```
eco-minio, line 95: I think this will fail if you remove policy as above correct? (For all subsequent requests)

eco-minio (changes requested):
Few comments

source/operations/install-deploy-manage/migrate-fs-gateway.rst
harshavardhana: It should be config export and import

bootc:
These instructions are missing steps for copying across ILM policies, which could always be set for `fs` buckets but stopped working some time ago.

harshavardhana:
> These instructions are missing steps for copying across ILM policies, which could always be set for `fs` buckets but stopped working some time ago.

ILM was never supported on FS mode, they will be copied if you perform `mc mirror -a`

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: 👍

harshavardhana (approved):

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+To upgrade to a :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later, those who were using the `standalone` or `filesystem` deployment modes must create a new :ref:`Single-Node Single-Drive <minio-snsd>` deployment and migrate settings and content to the new deployment.
+
+This document outlines the steps required to successfully launch and migrate to a new deployment.
+
+.. important:: 
+
+   Standalone/file system mode continues to work on any release up to and including `RELEASE.2022-10-24T18-35-07Z <https://github.com/minio/minio/releases/tag/RELEASE.2022-10-24T18-35-07Z>`__.
+   To continue using a standalone deployment, install that MinIO release or any `earlier release <https://github.com/minio/minio/releases>`__.
+
+   Do not upgrade to any release later than ``RELEASE.2022-10-24T18-35-07Z``.
```
ravindk89, line 35: I think we can drop this, we get the point across in the first two paragraphs.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+
+Procedure
+---------
+
+#. Retrieve the existing deployment's **environment variables**
+
+   Use the :mc-cmd:`mc admin config` export command to retrieve the environment variables defined on the existing standalone MinIO deployment.
```
ravindk89, line 42:
```suggestion
   Use the :mc-cmd:`mc admin config export <mc admin config>`  command to retrieve the environment variables defined on the existing standalone MinIO deployment.
```

ravindk89: I see the command is missing - I would say just add a stub for now on the mc admin config page that just says

```
mc admin config export
- exports any configuration settings created using `mc admin config set`
```

Since as @eco-minio noted previously anything envvar set does not necessarily end up in this data (iirc)

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+
+   Standalone/file system mode continues to work on any release up to and including `RELEASE.2022-10-24T18-35-07Z <https://github.com/minio/minio/releases/tag/RELEASE.2022-10-24T18-35-07Z>`__.
+   To continue using a standalone deployment, install that MinIO release or any `earlier release <https://github.com/minio/minio/releases>`__.
+
+   Do not upgrade to any release later than ``RELEASE.2022-10-24T18-35-07Z``.
+
+Procedure
+---------
+
+#. Retrieve the existing deployment's **environment variables**
```
ravindk89, line 40: I think we can tab this out and note as a pre-graph something like

"You can set MinIO configuration settings in environment variables and using :mc-cmd:`mc admin config set <mc admin config>`. 
Depending on your current deployment setup, you may need to retrieve the values for both:"

For envvars they can use something like `env | grep MINIO_" or check the `/etc/default/minio` file for systemd setups. Ultimately users can do whatever so general guidance here is sufficient.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+
+      mc admin tier ls ALIAS --json
+
+   - Replace ``ALIAS`` with the alias for the existing standalone deployment.
+   
+   Use the list to recreate the tiers on the new deployment.
+
+#. Duplicate **policies** from existing standalone deployment to new deployment
+
+   Use :mc-cmd:`mc admin policy list` with the ``--json`` flag  to retrieve a list of policies that exist on the standalone deployment.
```
ravindk89, line 116: Unfortunately this doesn't work. You need to list first, then use `mc admin policy info ALIAS POLICYNAME`. No `--json` required.

I thought we at one point had a mass export function but I don't see it anymore.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+   .. code-block:: shell
+      :class: copyable
+
+      mc admin group list ALIAS --json
+
+   - Replace ``ALIAS`` with the alias for the existing standalone deployment.
+
+   Use the list to recreate the groups on the new deployment.
+
+#. Duplicate **users** from existing standalone deployment to new deployment
```
ravindk89, line 140: We should note that if the original deployment was using external IAM, they need to recreate those settings in the SNSD deployment along with all associated policies.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+   Use :mc-cmd:`mc admin user svcacct list` with the ``--json`` flag to list existing service accounts on the standalone deployment.
+
+   .. code-block:: shell
+      :class: copyable
+
+      mc admin user svcacct list ALIAS --json
+
+   - Replace ``ALIAS`` with the alias for the existing standalone deployment.
+
+   Use the list to recreate the service accounts on the new deployment.
```
ravindk89, line 167: Should link to `mc admin user svcacct create` I think

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+   .. code-block:: shell
+      :class: copyable
+
+      mc mirror --preserve --watch SOURCE/BUCKET TARGET/BUCKET
+
+   - Replace ``SOURCE/BUCKET`` with the alias and a bucket for the existing standalone deployment.
+   - Replace ``TARGET`` with the alias and corresponding bucket for the new deployment.
+
+#. After the initial mirror process completes, convert standalone deployment to be read only
+
```
ravindk89, line 180: We could address Eco's point by simply noting "If the legacy deployment does not have any active write operations, you can skip this step and let `mc mirror` complete as per the next step."

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+   .. code-block:: shell
+      :class: copyable
+
+      mc mirror --preserve --watch SOURCE/BUCKET TARGET/BUCKET
+
+   - Replace ``SOURCE/BUCKET`` with the alias and a bucket for the existing standalone deployment.
+   - Replace ``TARGET`` with the alias and corresponding bucket for the new deployment.
+
+#. After the initial mirror process completes, convert standalone deployment to be read only
```
ravindk89, line 179: More to the point is just to stop all write operations to the legacy deployment, from any S3 or POSIX client.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+
+   .. code-block:: shell
+
+      mc admin policy unset ALIAS POLICYNAME [user=USERNAME | group=GROUPNAME]
+
+   - Replace ``ALIAS`` with the alias for the existing standalone deployment 
+   - Replace ``POLICYNAME`` with the name of a policy on the existing standalone deployment.
+   - Replace either ``USERNAME`` or ``GROUPNAME`` with either the user or group assigned the policy.
+
+   Use care to not remove permissions for the user running the ``mc mirror`` command.
```
ravindk89, line 191: IMO this is too risky. It's enough to ask users to stop clients from writing to this deployment any further.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -0,0 +1,203 @@
...
+
+#. Wait for ``mc mirror`` to complete for all buckets for any remaining operations.
+
+#. Stop the server for both deployments.
+
+#. Restart the new MinIO deployment with the ports used for the previous standalone deployment.
```
ravindk89, line 197: Ensure you apply all environment variables and runtime configuration settings, and validate the behavior.

ravindk89 (changes requested):

harshavardhana (commented):
One more thing we can mention is `mc admin cluster iam export` and `import` ? Daryl @ravind89

ravindk89:
I didn't see that option in help. Maybe it's hidden? Will check again but
that's what I was hoping we could learn on

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: dropped

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: 👍

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: 👍

djwfyi:
Updating both bucket and IAM sections to use the `mc admin cluster` import/export functionality.
I'll have a separate PR with the updates for these changes.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: Including this and cutting the last step, as this covers that.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: Cut.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: 👍

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: I simplified the step to just stopping writes from S3 or POSIX clients.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: This is cut and covered by new steps to use IAM export/import instead. Svcaccts are included in that process.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: Moved to a note.

jtackaberry:
@djwfyi I followed this procedure today and it went well.  Very nicely documented, thanks for putting it together.

The only slight surprise for me step 13 "Wait for mc mirror to complete for all buckets for any remaining operations" at first made concerned that it stalled out, because I was expecting the `mc mirror` command to terminate.  It didn't take long to realize this is (reasonably) expected with `--watch` but it gave me a momentary pause of concern.

Otherwise the migration went swimmingly, and I was pleasantly surprised by the import/export convenience for the bucket metadata and IAM resources.

djwfyi:
Great to hear @jtackaberry!

PR #631, opened 2022-11-04
Updates the Gateway migration doc

djwfyi:
- Utilizes mc admin cluster commands to export/import bucket and IAM settings.
- Adds export and import to the mc admin config command reference doc.
- Applies additional feedback from #628  that was added after that PR was merged.

### DIFF
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -32,21 +32,18 @@ This document outlines the steps required to successfully launch and migrate to 
    Standalone/file system mode continues to work on any release up to and including `RELEASE.2022-10-24T18-35-07Z <https://github.com/minio/minio/releases/tag/RELEASE.2022-10-24T18-35-07Z>`__.
    To continue using a standalone deployment, install that MinIO release or any `earlier release <https://github.com/minio/minio/releases>`__.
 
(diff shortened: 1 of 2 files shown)


source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -32,21 +32,19 @@ This document outlines the steps required to successfully launch and migrate to
...
-
-   .. code-block:: shell
-      :class: copyable
+.. note:: 
+   
+   You can set MinIO configuration settings in environment variables and using :mc-cmd:`mc admin config set <mc admin config>`.
+   Depending on your current deployment setup, you may need to retrieve the values for both.
 
-      mc admin config export ALIAS > config.txt
+   The procedure does not cover this, as there are many options you may define and multiple ways for retrieving the information.
```
ravindk89, line 45:
```suggestion
   This procedure does not cover migrating environment variables due to the variety of configuration methods.
   You can examine any runtime settings using ``env | grep MINIO_`` or, for deployments using MinIO's systemd service, check the contents of ``/etc/default/minio``
```

ravindk89 (approved):
one nit, otherwise LGTM. We can iterate further if we *really* need to, but this is a lot of guidance as is.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: 👍

PR #637, opened 2022-11-09
Fix typo in installation

reivaj05:
Found this typo while reading the documentation. 
Not sure if this everything that is needed since i'm not familiar with how docs are generated but hope this helps :)

Please close this PR if i'm missing something


djwfyi (approved):
Thanks for catching and fixing that!

PR #649, opened 2022-11-21
Adding bucket limit information

### LINKED ISSUE #548: Limit to the number of buckets (MinIO RELEASE.2022-09-17T00-09-45Z)
## Description
limit number of buckets to 500k

# Summary

- Add limit to the reference pages for `mc mb`
- We need to import the Limits page from the legacy docs, should include this here
- Need to update training materials as well.

djwfyi:
- Imports the limits doc from legacy into the Checklists section
- Adds 500K limit to buckets in several places

Closes #548

### DIFF
--- a/source/administration/minio-console.rst
+++ b/source/administration/minio-console.rst
@@ -147,6 +147,7 @@ Use the :guilabel:`Search` bar to search for specific buckets or objects.
 Select the row for the bucket or object to browse. 
 
 Select :guilabel:`Create Bucket` to create a new bucket on the deployment.
+The S3 API allows for a maximum of 500,000 buckets per deployment.
 
 Each bucket has :guilabel:`Manage` and :guilabel:`Browse` buttons.
 
--- a/source/operations/checklists/thresholds.rst
+++ b/source/operations/checklists/thresholds.rst
@@ -1,0 +1,139 @@
+.. _minio-server-limits:
+
+=====================
+Thresholds and Limits
+=====================
+
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+This page reflects limits and thresholds that apply to MinIO.
+
+Refer to the :ref:`hardware <minio-hardware-checklist>` and :ref:`software <minio-software-checklists>` for related recommendations and requirements.
+
+S3 API Limits
+-------------
+
+.. list-table::
+   :header-rows: 1
+   :widths: 60 40
+   :width: 90%
+
+   * - Item
+     - Specification 
+
+   * - Maximum object size
+     - 50 TiB
+
+   * - Minimum object size
+     - 0 B
+
+   * - Maximum object size per PUT operation
+     - | 5 TiB for non-multipart upload
+       | 50 TiB for multipart upload
+
+   * - Maximum number of parts per upload
+     - 10,000
+
+   * - Part size range
+     - 5 MiB to 5 GiB. Last part can be 0 B to 5 GiB
+
+   * - Maximum number of parts returned per list parts request
+     - 10,000
+
+   * - Maximum number of objects returned per list objects request
+     - 1,000
+
+   * - Maximum number of multipart uploads returned per list multipart uploads request
+     - 1,000
+
+   * - Maximum length for bucket names
+     - 63
+
+   * - Maximum length for object names
+     - 1024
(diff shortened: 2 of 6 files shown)


source/operations/checklists/thresholds.rst
```diff
@@ -0,0 +1,144 @@
...
+     - Specification 
+
+   * - Maximum number of servers per cluster
+     - no limit
+
+   * - Minimum number of servers
+     - 1
+
+   * - Minimum number of drives per server when server count is 1
+     - 2
```
ravindk89, line 86: This should be 1 now right, since *technically* we support SNSD w/ 0-parity erasure coding?

Or we can state that explicitly here with a note that "SNSD Deployments support 1x1 setups in a 0-parity mode that provides no failover tolerance."

source/operations/checklists/thresholds.rst
```diff
@@ -0,0 +1,144 @@
...
+The above list is not exhaustive and may not apply to your operating system and filesystem combination.
+
+Consult your operating system vendor or filesystem documentation for a comprehensive list for your situation.
+
+MinIO recommends using LInux operating system with an XFS based filesystem for production workloads.
+
+Conflicting Objects
+~~~~~~~~~~~~~~~~~~~
+
+Objects cannot have a conflicting object as its parent.
```
ravindk89, line 131: I cannot understand this sequence.

I assume the intent here from the examples is that you cannot have an object name that matches a prefix for nested objects

e.g. for the two examples, the problem is the object `/a/b` matches the prefix `/a/b/`

source/reference/minio-mc/mc-mb.rst
```diff
@@ -22,6 +22,7 @@ specified path.
 
 You can also use :mc:`mc mb` against the local filesystem to produce
 similar results to the ``mkdir -p`` commandline tool.
+The S3 API allows for up to 500,000 buckets on a deployment or tenant.
```
ravindk89, line 25: MinIO supports up to 500K buckets per deployment

since it's not an S3 limit, per se

Also you can do

```
.. cond:: k8s
   
   MinIO supports up to 500,000 buckets per tenant.

.. cond:: not k8s

  MinIO supports up to 500,000 buckets per deployment.
```

djwfyi: 👍

source/operations/checklists/thresholds.rst
djwfyi: Fixed.

harshavardhana (changes requested):
We are planning again remove this limit from our docs and code to make sure we only keep this as a recommendation instead of a hard-limit.

So calling this as hard limit is not needed in these docs IMO

ravindk89:
> We are planning again remove this limit from our docs and code to make sure we only keep this as a recommendation instead of a hard-limit.
> 
> So calling this as hard limit is not needed in these docs IMO

Which line harsha? I think that got lost in the review.

source/operations/checklists/thresholds.rst
```diff
@@ -0,0 +1,139 @@
...
+Applications must assign non-conflicting, unique keys.
+
+For example, MinIO does not support a situation where an object's name is also the name of the prefix for a child object. 
+MinIO does not allow the following two examples, as the object names conflict:
```
ravindk89, line 129:
```suggestion
For the following example operations, the second PUT operation fails because of a naming conflict with the object created by the first. 
```

ravindk89 (approved):
One nit, one general thought/suggestion.

source/operations/checklists/thresholds.rst
djwfyi: Good wording. Added.

source/administration/minio-console.rst
```diff
@@ -147,6 +147,7 @@ Use the :guilabel:`Search` bar to search for specific buckets or objects.
 Select the row for the bucket or object to browse. 
 
 Select :guilabel:`Create Bucket` to create a new bucket on the deployment.
+The S3 API allows for a maximum of 500,000 buckets per deployment.
```
harshavardhana, line 150: We seem to still mention this!?

djwfyi: I'll fix it. Thanks for catching it!

PR #657, opened 2022-11-30
Fix IDP info command

reivaj05:


djwfyi (approved):
Nice catch. Thanks for fixing it!

PR #661, opened 2022-12-05
UI Enhancements: URLs, Web-font path and CSS fixes

rushenn:
- Only the external links are set to open in new windows/tab
- Fixed incorrect web-font paths being called for preload. 
- Minor CSS transition fixes.
- Conditionally called the search modal's keyboard events.

### DIFF
--- a/source/_static/js/instantSearch.js
+++ b/source/_static/js/instantSearch.js
@@ -271,7 +271,7 @@ window.addEventListener("DOMContentLoaded", () => {
             return false;
           }
 
-          return `<a target="_blank" href="${docUrl}">
+          return `<a href="${docUrl}">
                     ${returnString}
                   </a>`;
         },
@@ -331,7 +331,10 @@ window.addEventListener("DOMContentLoaded", () => {
     (e) => {
       // Close the search on esc key press
       if (e.key === "Escape") {
-        closeSearchModal();
+        if(searchModalEl.classList.contains("search--focused") 
+        || searchModalEl.classList.contains("search--active")) {
+          closeSearchModal();
+        }
       }
 
       // Focus the search input on "Meta + K" key press
--- a/source/_static/js/instantSearch.js
+++ b/source/_static/js/instantSearch.js
@@ -123,13 +123,6 @@ window.addEventListener("DOMContentLoaded", () => {
           clearRefinements();
         }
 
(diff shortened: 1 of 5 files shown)


djwfyi:
External vs internal links are working great! Thank you for working on this.

Couple of thoughts:

1. Since search results don't persist when leaving the page, it might be better to have any search result click open in a new window.
Either that, or we could have a search results page instead. Maybe the results modal could have a link somewhere to "Open search results in a separate window".
I'm just thinking when you Google something, clicking on a result opens in the same window. But you can press back to go back to the results. That doesn't work with our search, since the results and the search do not persist in any way.

2. When I did a search, my first impulse was to just click outside the search result modal to close the search and stay on the page. With this iteration, I have to either press `Esc` or click the modal's `x` button. This may seem unexpected based on users experiences with other search modals on other sites.
But also, that might just be me. Thoughts?

rushenn:
Make sense @djwfyi. These are good feature to have. Point no. 2 is doable as part of this PR. I might need some time for the point no. 1 as I need to check the InstantSearch API. I can address this with a new PR.

rushenn:
@djwfyi @ravindk89 Added a commit to address the [point no.2 ](https://github.com/minio/docs/pull/661#issuecomment-1337693006). Please take another look.

I will explore the possibilities for the "Search result page" and address it with the new PR.

djwfyi (commented):
The modal closing when clicking outside of the results works great.
I just want to make sure, though. This also reverted the outside links to opening in the same window. Basically, the behavior we had before. 

Links from the search modal open in a new window.
All other links (whether to MinIO docs or 3rd party websites) open in the same window.

Is that change on hold for your exploration of a search results page?

rushenn:
@djwfyi I haven't reverted any of the changes we had prior to the "click outside to close" commit. It works like the following for me:

- Outside links are opened in new window/tab.
- Internal links are opened in the same window. 

The only I change I made was, I have set the search result links to open in new window/tab. Please let me know if I'm missing anything?

djwfyi (approved):
Looks like the outside link behavior I was seeing was fixed when I did a clean build.
All is good.

Internal links open in the same window.
External links open in a new window.
Search modal closes when clicking outside of the modal.

PR #669, opened 2022-12-08
Fix anchor scrolling issue on Webkit browsers

rushenn:


kaankabalak (approved):
LGTM & tested

djwfyi (changes requested):
This fixes the scrolling issue, but we have a new issue with this PR that breaks retention of the reading mode and dark mode on Firefox.

To reproduce:

- Build this PR
- Open the site in Firefox
- activate dark mode and or reading mode
- Click on any link in the navigation

Page returns to the standard page view in light mode.

djwfyi:
> This fixes the scrolling issue, but we have a new issue with this PR that breaks retention of the reading mode and dark mode on Firefox.
> 
> To reproduce:
> 
> * Build this PR
> * Open the site in Firefox
> * activate dark mode and or reading mode
> * Click on any link in the navigation
> 
> Page returns to the standard page view in light mode.

Seems to be due to viewing local files directly rather than through a web server.
Problem resolved after using a python server to view the pages instead.

djwfyi (approved):
LGTM

PR #674, opened 2022-12-09
Adds a recommendation to limit prefix contents to 10k objects.

### LINKED ISSUE #666: [SUBNET] Add recommendations for prefix object limits
**Describe the bug**
Per request from @klauspost in Slack, the docs do not currently have recommendations for how many objects can be reasonably stored under a prefix.
Add information in the docs that for best performance, any specific prefix should not have more than 10k objects as a general rule.

**Additional context**
- [S3 docs on optimizing performance.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html)
- [S3 docs on organizing prefixes](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-prefixes.html)

djwfyi:
Per @klauspost we need to add info to the docs about MinIO's recommendation to limit prefixes to no more than 10k objects.
This updates the glossary entry, object management doc, and admin concepts to add that recommendation.

Closes #666

### DIFF
--- a/source/administration/concepts.rst
+++ b/source/administration/concepts.rst
@@ -103,6 +103,16 @@ MinIO automatically generates two folders in the ``articles`` bucket based on th
          2022-01-02-MinIO-Advanced-Deployment-comments.json
          2022-01-04-MinIO-Interview.md
 
+MinIO itself does not limit the number of objects that any specific prefix can contain.
+However, hardware and network conditions may show performance impacts with large prefixes.
+
+- Deployments with modest or budget-focused hardware should architect their workloads to target 10,000 objects per prefix as a baseline. 
+  Increase this target based on benchmarking and monitoring of real world workloads up to what the hardware can meaningfully handle. 
+- Deployments with high-performance or enterprise-grade :ref:`hardware <deploy-minio-distributed-recommendations>` can typically handle prefixes with millions of objects or more.
+
+|SUBNET| Enterprise accounts can utilize yearly architecture reviews as part of the deployment and maintenance strategy to ensure long-term performance and success of your MinIO-dependent projects.
+
+For a deeper discussion on the benefits of limiting prefix contents, see the article on :s3-docs:`optimizing S3 performance <optimizing-performance.html>`.
 
 How can I backup and restore objects on MinIO?
 ----------------------------------------------
--- a/source/administration/object-management.rst
+++ b/source/administration/object-management.rst
@@ -45,8 +45,18 @@ following:
          2020-01-02-MinIO-Advanced-Deployment-comments.json
          2020-01-04-MinIO-Interview.md
 
-MinIO supports multiple levels of nested directories and objects to support 
-even the most dynamic object storage workloads.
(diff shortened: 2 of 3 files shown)


source/administration/object-management.rst
```diff
@@ -45,8 +45,8 @@ following:
          2020-01-02-MinIO-Advanced-Deployment-comments.json
          2020-01-04-MinIO-Interview.md
 
-MinIO supports multiple levels of nested directories and objects to support 
-even the most dynamic object storage workloads.
+MinIO supports multiple levels of nested directories and objects using a :term:`prefix` structure to support even the most dynamic object storage workloads.
+For the best performance, MinIO recommends that any single prefix contain no more than 10,000 objects.
```
harshavardhana, line 49: I think 10,000 is very small @djwfyi it can be easily 1million without issues on a large enough cluster. It is only a problem of smaller deployments that do not have a lot of disks.

harshavardhana (changes requested):

source/administration/concepts.rst
```diff
@@ -103,6 +103,9 @@ MinIO automatically generates two folders in the ``articles`` bucket based on th
          2022-01-02-MinIO-Advanced-Deployment-comments.json
          2022-01-04-MinIO-Interview.md
 
+MinIO does not technically limit the number of objects that any specific prefix can contain.
```
eco-minio, line 106:
```suggestion
MinIO itself does not limit the number of objects that any specific prefix can contain, rather it is a function of the hardware provided.
```

eco-minio (commented):
One potential change if we want to point out that it is a physics limitation technically

klauspost (approved):
LGTM

ravindk89:
So I *think* we still have a subjective issue here - what is "smaller" and what is "larger".

Someone could deploy a "large" cluster using atom processors and old slow HDD disks. They would run afoul of this issue.

Someone else could deploy a "small" cluster that uses high end xeon processors and optane disks. They might not have any issues.

Rather than "small" or "large", I think we would instead focus on the quality of the hardware. Spitballing:

- MinIO performance is directly associated to the quality of the underlying hardware
- Budget setups using low-end proccessors, storage mediums, and networking may exhibit performance issues as the number of objects in a single prefix grows
- Use tools like warp, dperf, nperf, etc. to better understand the performance characteristics of the hardware
- Budget setups can target ~10K objects per prefix as a starting point in absence of performance testing

Thoughts?

klauspost:
@ravindk89 They can contact us on subnet for a more precise estimate.

ravindk89:
@klauspost main issue is users (and customers) who do read the docs tend to read them without wiggle room (unfortunately). 

I think this is resolved by slightly tweaking the wording to something closer to "small deployments should target no more than 10K objects per prefix and evaluate cluster performance at that level. MinIO provides architecture reviews through SUBNET..."

That way it's more guidance vs "If you hit 10K+ your MinIO deployment will always melt"

ravindk89 (approved):

PR #682, opened 2022-12-20
Proof of Concept on right-nav video links via RST + JS

ravindk89:
As per @petehnath and @abperiasamy we want to have some curated links in the right nav bar for select pages.

This is an initial proof of concept on how this might function, both in terms of RST and JavaScript to make the work happen.

In this phase it would require placing a container on each target page from which we insert into the right sidebar the links.

This has not had any styling around how this would work on mobile if at all - probably on mobile we would want to disable the JS call and just have a styled box on the page for mobile users to look at.

This needs review by the webdesign team and @djwfyi as far as how this all feels.

Longer term it would be better to design something - possibly as part of Sphinx via a custom directive, but perhaps more easily done as a python pre-processor that uses a CSV to generate the necessary block on each page listed in the CSV. This would be easier to manage overall and create a single place for the education team to manage.

Still, this will do for short term.

### DIFF
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -157,19 +157,52 @@ window.addEventListener("DOMContentLoaded", (event) => {
 
 
   // --------------------------------------------------
-  // TOC
+  // TOC, External Links
   // --------------------------------------------------
   (function () {
-    // Move the TOC to the right side of the page
-    const tocOriginalEl = document.getElementById("table-of-contents");
-    const tocTargetEl = document.getElementById("content-toc");
-    const tocAsideEL = document.querySelector(".content__toc");
+    // Move the TOC, External Links, etc. to the right side of the page
+    const extVideoLinks = document.querySelector(".extlinks-video")
 
-    if (tocOriginalEl) {
-      tocTargetEl.parentNode.replaceChild(tocOriginalEl, tocTargetEl);
+    const tocPreferredEl = document.getElementById("table-of-contents");
+    const tocLegacyEl = document.getElementById("content-toc");
+
+    // This is the target element for the "aside" or right nav bar
+    const tocAsideEl = document.querySelector(".content__toc");
+
+    // Don't need this element so remove it
+    tocLegacyEl.remove();
+
+    // Build an array of elements to add, in order
+    // Then iterate the array and append it, one by one, to the aside
+    const asideElements = new Array();
+
+    if (tocPreferredEl) {
+      asideElements.push(tocPreferredEl);
+    }
+    if (extVideoLinks) {
+      // Minor cleanups to the CSS classes
+      extVideoLinks.classList.remove("docutils", "container")
+      extVideoLinks.classList.add("topic")
+
+      // Inject the header text
+
+      const extVideoLinkHeader = document.createElement("div");
+      extVideoLinkHeader.classList.add("extVideoLink-header");
+      extVideoLinkHeader.innerHTML += "<p>Recommended Videos</p>";
+      extVideoLinks.prepend(extVideoLinkHeader);
+
+
+      asideElements.push(extVideoLinks)
+    }
+
+    // Sets the empty CSS class if nothing on the page goes in the sidebar
+    if (asideElements.length == 0) {
+      tocAsideEl.classList.add("content__toc--empty");
     }
     else {
-      tocAsideEL.classList.add("content__toc--empty");
+      for (i = 0; i < asideElements.length; i++) {
+         tocAsideEl.appendChild(asideElements[i]);
+       }
     }
     
     // Treat the TOC as a dropdown in mobile
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -193,6 +193,15 @@ window.addEventListener("DOMContentLoaded", (event) => {
 
 
       asideElements.push(extVideoLinks)
+
+      // Need to force-add extlinks, they don't seem to get added as-is for some reason
+
+      const extVideoItems = document.querySelectorAll(".extlinks-video a");
+      extVideoItems.forEach(item => {
+         item.setAttribute("target","_blank");
+         item.setAttribute("rel", "noopener");
+         item.setAttribute("rel", "noreferrer");
+      });
     }
 
     // Sets the empty CSS class if nothing on the page goes in the sidebar
--- a/source/includes/common/installation.rst
+++ b/source/includes/common/installation.rst
@@ -10,6 +10,14 @@ Install and Deploy MinIO
    :local:
    :depth: 1
 
+.. container:: extlinks-video
+
+   - `Installing and Running MinIO on Linux <https://www.youtube.com/watch?v=74usXkZpNt8&list=PLFOIsHSSYIK1BnzVY66pCL-iJ30Ht9t1o>`__
+
+   - `Object Storage Essentials <https://www.youtube.com/playlist?list=PLFOIsHSSYIK3WitnqhqfpeZ6fRFKHxIr7>`__
+   
+   - `How to Connect to MinIO with JavaScript <https://www.youtube.com/watch?v=yUR4Fvx0D3E&list=PLFOIsHSSYIK3Dd3Y_x7itJT1NUKT5SxDh&index=5>`__
+
 MinIO is a software-defined high performance distributed object storage server.
 You can run MinIO on consumer or enterprise-grade hardware and a variety
 of operating systems and architectures.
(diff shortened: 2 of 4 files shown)


source/_static/js/main.js
```diff
@@ -157,19 +157,52 @@ window.addEventListener("DOMContentLoaded", (event) => {
...
+      // Inject the header text
+
+      const extVideoLinkHeader = document.createElement("div");
+      extVideoLinkHeader.classList.add("extVideoLink-header");
+      extVideoLinkHeader.innerHTML += "<p>Recommended Videos</p>";
+      extVideoLinks.prepend(extVideoLinkHeader);
+
+
+      asideElements.push(extVideoLinks)
+    }
```
ravindk89, line 196: Couldn't think of a better way to do this - and at least this way we don't need to hardcode it within the RST.

source/_static/js/main.js
```diff
@@ -157,19 +157,52 @@ window.addEventListener("DOMContentLoaded", (event) => {
...
+
+    // Sets the empty CSS class if nothing on the page goes in the sidebar
+    if (asideElements.length == 0) {
+      tocAsideEl.classList.add("content__toc--empty");
     }
     else {
-      tocAsideEL.classList.add("content__toc--empty");
+      for (i = 0; i < asideElements.length; i++) {
+         tocAsideEl.appendChild(asideElements[i]);
+       }
```
ravindk89, line 205: Idea is we can eventually also add something like "Recommended Blog" or "Related Resources" or whatever and just define it up above, and for any given page it just happens on load.

source/includes/common/installation.rst
```diff
@@ -10,6 +10,14 @@ Install and Deploy MinIO
    :local:
    :depth: 1
 
+.. container:: extlinks-video
```
ravindk89, line 13: @djwfyi this ended up working well enough, and it's dirt simple. We could change the classname if we wanted to.

Admonitions come with a lot of other classes attached, which made it a little harder to script / style without accidentally munging something else.

kaankabalak:
Thanks @ravindk89, the proof of concepts is mostly looking good to me. As a small improvement, we might want to open the links on an external window. We can apply a change like the following to implement this:

```diff
diff --git a/source/_static/js/main.js b/source/_static/js/main.js
index b790160..33f6b07 100644
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -184,6 +184,14 @@ window.addEventListener("DOMContentLoaded", (event) => {
       extVideoLinks.classList.remove("docutils", "container")
       extVideoLinks.classList.add("topic")
 
+      // Add attributes to individual links
+      const extVideoItems = document.querySelectorAll(".extlinks-video a");
+      extVideoItems.forEach(item => {
+        item.setAttribute("target", "_blank");
+        item.setAttribute("rel", "noopener");
+        item.setAttribute("rel", "noreferrer");
+      })
+
       // Inject the header text
 
       const extVideoLinkHeader = document.createElement("div");
```

ravindk89:
Good idea - odd it didn't do that by default. Maybe something about how I'm generating the links, Sphinx is usually pretty good about opening external links in a new window. Will poke and implement.

kaankabalak:
As another small improvement, we can align the recommended links to the right hand side of the page for viewport widths less than 992px too:

Screen Shot 2022-12-20 at 2 29 32 PM


We can apply the following change to do so:

```diff
diff --git a/source/_static/scss/includes/_toc.scss b/source/_static/scss/includes/_toc.scss
index 3b9a406..abf7a13 100644
--- a/source/_static/scss/includes/_toc.scss
+++ b/source/_static/scss/includes/_toc.scss
@@ -39,8 +39,12 @@ div.topic {
                  margin-left: 0.75rem;
            }
         }
-      border-bottom: 1px $dark-500 solid;
-      padding: .5rem 0 .5rem 0;
+        border-bottom: 1px $dark-500 solid;
+        padding: .5rem 0 .5rem 0;
+
+        @include breakpoint-max(breakpoints(lg)) {
+            margin-left: auto;
+        }
     }
 
     .extVideoLink-header {
```

rushenn:
> As another small improvement, we can align the recommended links to the right hand side of the page for viewport widths less than 992px too:
> 
> Screen Shot 2022-12-20 at 2 29 32 PM
> 
> We can apply the following change to do so:
> 
> ```diff
> diff --git a/source/_static/scss/includes/_toc.scss b/source/_static/scss/includes/_toc.scss
> index 3b9a406..abf7a13 100644
> --- a/source/_static/scss/includes/_toc.scss
> +++ b/source/_static/scss/includes/_toc.scss
> @@ -39,8 +39,12 @@ div.topic {
>                   margin-left: 0.75rem;
>             }
>          }
> -      border-bottom: 1px $dark-500 solid;
> -      padding: .5rem 0 .5rem 0;
> +        border-bottom: 1px $dark-500 solid;
> +        padding: .5rem 0 .5rem 0;
> +
> +        @include breakpoint-max(breakpoints(lg)) {
> +            margin-left: auto;
> +        }
>      }
>  
>      .extVideoLink-header {
> ```

@kaankabalak @ravindk89 I think it's better to move the recommended links to the bottom of the page (above the footer) or remove it completely from the mobile devices so the reading experience wouldn't be compromised. Current design would cause issues when there is more/lengthier link titles and, on compact mobile devises such as smaller iPhones and all, side by side display wouldn't possible at all.

ravindk89:
@rushenn no objection to either - for moving it to the bottom of the page...I suppose I need something in the javascript to detect the screen size and basically pop/push the element.

TBH it's probably easier to just remove it or have some CSS To make the element hidden if it's inside of the main content block. Let me see if I can figure out how we're doing that for the current TOC< and if not I kindly request some help :)

ravindk89:
Ah I see the issue - @rushenn our current logic in main.js has us set:

```js
  const tocMenuEl = document.querySelector("#table-of-contents > ul.simple");
```

We then do as part of hiding the aside on mobile:

```js
      if(tocMenuEl) {
        tocMenuEl.closest(".content__toc").classList.remove("active");
      }
```

I'm not entirely clear how the first line helps us determine if we are on mobile - is it that you can reliably assume that on mobile, certain elements get rendered differently @rushenn @kaankabalak ?

So on pages with a recommended video section, it seems to break setting the 'active' var on the sidebar - that is, the entire action of hiding the toc menu on mobile basically fails.

rushenn:
@ravindk89 

```
 if(tocMenuEl) {
        tocMenuEl.closest(".content__toc").classList.remove("active");
}
```

Above code is used to show/hide the TOC as a dropdown in mobile and the styles are applied based on the CSS media query. 

> So on pages with a recommended video section, it seems to break setting the 'active' var on the sidebar - that is, the entire action of hiding the toc menu on mobile basically fails.

I do not see any issues other than the mobile display with your implementation @ravindk89. Again, I would recommend it to hide it completely on the mobile at least for now.

ravindk89:
Hm. OK - so if I can figure out how to apply the CSS (maybe based on breakpoints?) to just set the entire thing as hidden, then we've solved this for now.

kaankabalak:
@ravindk89 you can apply the CSS up until a certain breakpoint as follows. The following example applies the style until the large breakpoint (hence the query `breakpoint-max(breakpoints(lg))`, where `max` indicates that the rule should be applied until the dimension specified in the argument is reached, and `breakpoints(lg)` indicates the large breakpoint as per our definition [here](https://github.com/minio/docs/blob/main/source/_static/scss/includes/_variables.scss#L26).)

```css
@include breakpoint-max(breakpoints(lg)) {
    /* your CSS here */
}
```

Meanwhile, the following example applies the style starting from the small breakpoint:

```css
```

djwfyi:
I like the way this is implemented as a container at the top of the rst file.
Not too difficult to maintain. Have you tested it with conditionals yet?

I agree that hiding the container on mobile is the best at this point. We can revisit later.

I'm not a huge fan of the visual on a page that has both a TOC and the video list. I'd like to see some more styling go on to separate the two. Maybe just adjusting the background of the video links to match the left nav?

Also, how about `Related Videos` instead of `Recommended`?

djwfyi:
It would be snazzy if we can make the bullet point a video camera for the links.

ravindk89:
@kaankabalak @rushenn changes implemented, including hiding the navlinks for small screens.

@djwfyi I'm going to ask the design team to take a pass on this, as I generally agree the styling right now is pretty rough. But it's just a PoC, and it's enough to get the basics out of the door.

kaankabalak (approved):

PR #688, opened 2022-12-30
Fix k8s search result page's pathname

rushenn:


source/_static/js/instantSearch.js
```diff
@@ -382,6 +381,8 @@ window.addEventListener("DOMContentLoaded", () => {
             platform = activePlatform;
           }
 
+          var pathname = environment === "dev" ? "/search.html" : `/docs/minio/${platform}/search.html`;
```
djwfyi, line 384: Another concern here is that @ravindk89 is in the process of adding a bunch of additional k8s platforms, which means the 377-383 `if` block is going to have to greatly expand. And each time we add a new k8s platform, we'll have to remember to add it to this.

ravindk89: might not be another way around that for now. Thanks for the reminder as well.

djwfyi (approved):
Going to merge and see what happens.

source/_static/js/instantSearch.js
rushenn: Not sure how to handle this the right way. Do you think this can be handled programatically by defining both the main platform and sub somewhere in the `default.conf.py`?

djwfyi:
@rushenn @ravindk89 
Looks like this fixed it in prod. 🎉

PR #714, opened 2023-01-25
Add clickable buttons to navigate between images in the carousel

rushenn:
Preview:
Screenshot 2023-01-25 at 20 03 00

Limitations:
- Arrows are only present on the carousel and not in the Lightbox preview. In the future, I'll investigate the options for including arrows on the Light box. For this, we might need to replace the existing Lightbox plugin with a modern one. 
- Loop navigation is not available; instead, arrows will only be shown when they correspond to the first and last slides.

### DIFF
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -426,4 +426,70 @@ window.addEventListener("DOMContentLoaded", (event) => {
       });
     }
   })();
+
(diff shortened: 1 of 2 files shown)


djwfyi:
This looks great, @rushenn!
Thank you!

djwfyi (approved):
Great improvement!

PR #722, opened 2023-02-03
Port-Forwarding cleanups

ravindk89:
As per https://github.com/minio/console/issues/2539 , the websocket behavior integrated as part of Console 0.22.1 (https://github.com/minio/console/pull/2419) seems to break port-forwarding behavior.

There's no easy fix for this. NodePorts are a workaround, but slightly kludgy. Ingress is the better long-term solution, but requires more work.

This is a stopgap:

- For Operator, point users towards NodePorts if port-forwarding fails
- For Tenant Console, simply drop port-forwarding entirely and point only at Ingress/LB

Out of scope but in progress is Ingress guidance for Nginx and Traefik so we can close the loop on this.


djwfyi (changes requested):
These two includes for connecting via CLI also reference `port-forward`.

```
.. cond:: openshift

   .. include:: /includes/openshift/steps-deploy-minio-tenant.rst

.. cond:: k8s and not (openshift or eks)

   .. include:: /includes/k8s/steps-deploy-tenant-cli.rst
```

Rest of the content is LGTM.

PR #726, opened 2023-02-08
UI Enhancements: Recommended video section, k8s active class, mobile nav links

rushenn:
- Give recommended video section a makeover 
  Screenshot 2023-02-08 at 14 05 44

- Fix k8s's active nav state
- Fix nav links wrapping on mobile (current behaviour on production looks likes below)
  Screenshot 2023-02-08 at 11 50 10


source/_static/scss/includes/_theme.scss
```diff
@@ -111,6 +111,9 @@ $theme-properties: (
     --alert-warning-code-bg: #f3dfe1 #3d1014,
     --alert-warning-link-decoration-color: #e3c3c6 #54141a,
 
+    // Recommended videos
+    --recommended-videos-marker: url("../img/icons/play.svg") url("../img/icons/play-dark.svg"),
```
djwfyi, line 115: I like the concept of the icons for each video.
But this particular icon looks pretty pixelated on my screen.

Can we use a different icon image set?

source/_static/scss/includes/_nav.scss
```diff
@@ -479,6 +489,10 @@
 		background-image: linear-gradient(270deg, var(--content-nav-sub-bg) 0%, var(--content-nav-sub-bg) 10%, transparent 100%);
 	}
 
+	.container {
+		padding: 0;
+	}
+
```
djwfyi, line 495: We need some kind of scroll bar or button to access other K8s platforms on a mobile device.
Right now if you're on Upstream on a mobile device, only Upstream, Openshift, and EKS are accessible platforms. There is no way to move the nav bar to scroll to more options to the right.

djwfyi: image

djwfyi (changes requested):
- Kubernetes active state is great!
- Play button as bullet point for videos is good, but the specific image used looks pixelated.
- There does not seem to be a way to scroll to other platforms on mobile devices where the nav bar fades to indicate there are more options. (Maybe it works with tapping on a device, but dragging with my mouse on a desktop at a smaller windows doesn't work.)

source/_static/scss/includes/_theme.scss
rushenn: @djwfyi Can you share a screenshot? Its an SVG, so not its sure why it's getting pixelated.

source/_static/scss/includes/_nav.scss
rushenn: @djwfyi This is the original design and as per it, we'll have to use the horizontal mouse-wheel incase if we encounter this behaviour on desktop when the window is resized to < 991px. This experience will be mostly on mobile devices, and we can use the native gestures to scroll horizontally.

We can also add conditional arrows on both end at least on the desktop. Let us know your thought as well @oscarocastellanos

oscarocastellanos: @rushenn @djwfyi I don't think in mobile, the use of gestures comes as restrictive. Gestures are very intuitive for the majority of mobile users and are widely used on a lot of websites and platforms. Nonetheless, we might want to reduce or avoid any possible conflict or friction that any user might have. Adding arrows for navigation should solve this. Here's a mockup of what I think might work:

Captura de Pantalla 2023-02-09 a la(s) 10 00 00

Thanks for the heads up, Daryl!

source/_static/scss/includes/_theme.scss
djwfyi: Here's a screenshot.
image

djwfyi: It may just be my monitor. It's better on the MacBook's screen than on the external.

source/_static/scss/includes/_nav.scss
djwfyi: I like the arrows a lot. Thanks for adding them.

djwfyi (approved):
This looks good to me. Thanks for your work on it!
If no one else mentions the arrow, I'd chalk it up to my monitor being funky.

source/_static/scss/includes/_theme.scss
rushenn: Thanks @djwfyi, I will try to reproduce the issue.

PR #729, opened 2023-02-10
DOCS-728: Dropping Windows cruft

### LINKED ISSUE #728: Clean up Windows Docs
Simple one - we have a bunch of cruft on deployment topologies for Windows.

We really only support single-node single-drive. There are kludgy workarounds for multi-drive that we don't want to touch, because they are necessarily kludgy. 

They also do nothing to deal with inherent issues w/ NTFS for our workload type/pattern.

So we drop those for now, and in a later pass point users properly back to Linux.

ravindk89:
Closes #728

### DIFF
--- a/source/includes/windows/quickstart.rst
+++ b/source/includes/windows/quickstart.rst
@@ -15,7 +15,7 @@ This procedure deploys a :ref:`Single-Node Single-Drive <minio-installation-comp
    MinIO has limited support for multi-drive Windows deployments due to behaviors and limitations of the NTFS filesystem.
 
    This documentation only covers Single-Node Single-Drive deployment topologies on Windows.
-   Use Linux hosts for deploying Single-Node Multi-Drive or Multi-Node Multi-Drive MinIO topologies.
+   Use :minio-docs:`Linux hosts <minio/linux/operations/installation.html>` for deploying Single-Node Multi-Drive or Multi-Node Multi-Drive MinIO topologies.
 
 Use Windows-based MinIO deployments for early development and evaluation.
 MinIO strongly recommends Linux (RHEL, Ubuntu) systems for long-term development and production environments.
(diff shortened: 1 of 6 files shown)


source/includes/windows/quickstart.rst
```diff
@@ -10,7 +10,17 @@ Quickstart for Windows
...
+   This documentation only covers Single-Node Single-Drive deployment topologies on Windows.
+   Use Linux hosts for deploying Single-Node Multi-Drive or Multi-Node Multi-Drive MinIO topologies.
+
+Use Windows-based MinIO deployments for early development and evaluation.
+MinIO strongly recommends Linux (RHEL, Ubuntu) systems for long-term development and production environments.
+
+MinIO supports non-EOL Windows versions (Windows 10, Windows Server 2016+). 
 
 Prerequisites
 -------------
```
djwfyi, line 26: - Line 83, should update the port to the Console port (`9090`).
- Line 84 is confusing when rendered. I suggest making the :guilabel: section a code block instead and splitting up the paragraph.
   Or, just making it monotype instead of guilabel might be sufficient.

source/includes/windows/quickstart.rst
```diff
@@ -10,7 +10,17 @@ Quickstart for Windows
 
 This procedure deploys a :ref:`Single-Node Single-Drive <minio-installation-comparison>` MinIO server onto |OS| for early development and evaluation of MinIO Object Storage and its S3-compatible API layer. 
 
-For instructions on deploying to production environments, see :ref:`deploy-minio-distributed`.
+.. note::
+
+   MinIO has limited support for multi-drive Windows deployments due to behaviors and limitations of the NTFS filesystem.
+
+   This documentation only covers Single-Node Single-Drive deployment topologies on Windows.
+   Use Linux hosts for deploying Single-Node Multi-Drive or Multi-Node Multi-Drive MinIO topologies.
```
djwfyi, line 18: I suggest doing ref links on these for quick access to the Linux docs. Sphinx interlinking should Just Work™, right?

djwfyi (approved):
Couple of suggestions.
Don't need to see it again.

Otherwise, LGTM.

source/includes/windows/quickstart.rst
ravindk89: So I have _not_ been able to get this to work when there are overlapping refs.

The sphinx docs indicate it *should* work. In theory the syntax is correct. In practice it doesn't fire.

My guess is something in my custom code isn't completely correct for intersphinx. So external links will have to do for now.

Issue #757, opened 2023-03-15
Update docs build instructions


feorlen:
Update the docs build instructions in README.md. 

- [ ] Clarify individual build targets vs all targets.
- [ ] Add detail on dependencies for Ubuntu 22.04.

PR #758, opened 2023-03-15
Update docs build instructions

feorlen:
The docs build instructions in README.md should not recommend the outdated command `make all`. Instead, recommend building/viewing one docs target at a time.

Also add `venv` and `npm` to the prerequisites, they may not be installed by default on some OSes. (Specifically, Ubuntu. Tested on 22.04 LTS.)

Addresses https://github.com/minio/docs/issues/757

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,13 +34,13 @@ python3 -m venv venv && source venv/bin/activate
 pip install -r requirements.txt && npm install && npm run build
 ```
 
-4. Build all targets at once:
+4. Build your desired platform target.
 
 ```
-make all
+make k8s
 ```
 
-5. View the generated documentation visit http://localhost:8000:
+5. View the generated documentation at http://localhost:8000.
 
 ```
 python -m http.server --directory build/YOUR_BRANCH/<PLATFORM>/html
--- a/README.md
+++ b/README.md
@@ -8,8 +8,10 @@ MinIO uses [Sphinx](https://www.sphinx-doc.org/en/master/index.html) to generate
 
 - Any GNU/Linux Operating System.
 - python 3.10.x and python-pip
+- python3.10-venv
 - sphinx 4.3.2
 - nodejs 14.5.0 or later
+- npm
 - `git` or a git-compatible client
 
 ### Build


README.md
```diff
@@ -8,8 +8,10 @@ MinIO uses [Sphinx](https://www.sphinx-doc.org/en/master/index.html) to generate
 
 - Any GNU/Linux Operating System.
 - python 3.10.x and python-pip
+- python3.10-venv
 - sphinx 4.3.2
 - nodejs 14.5.0 or later
+- npm
```
feorlen, line 14: I don't know which versions of npm this works with, suggestions?

ravindk89: So locally at least 14.5.0 seems to be our base. That said I think we recently tested with Hugo...16.19.1 and it worked fine.

Since hugo seems to require 16.19 right now, and that's Node stable anyways, we could bump to that.

README.md
```diff
@@ -8,8 +8,10 @@ MinIO uses [Sphinx](https://www.sphinx-doc.org/en/master/index.html) to generate
 
 - Any GNU/Linux Operating System.
```
djwfyi, line 9: Probably worth dropping this line, or at least adding MacOS to it.

README.md
```diff
@@ -34,13 +36,13 @@ python3 -m venv venv && source venv/bin/activate
 pip install -r requirements.txt && npm install && npm run build
 ```
 
-4. Build all targets at once:
+4. Build your desired platform target.
 
 ```
-make all
+make k8s
```
djwfyi, line 42: k8s is going to drop a lot of sphinx warnings because of files not referenced in it.
I suggest `make linux` instead.

djwfyi (approved):

README.md
feorlen: macOS 12.3 (Monterey) finally stopped shipping py 2.7, so that might be a reasonable lower bound?

djwfyi: Seems reasonable to me.

README.md
```diff
@@ -6,10 +6,12 @@ MinIO uses [Sphinx](https://www.sphinx-doc.org/en/master/index.html) to generate
 
 ### Prerequisites
 
-- Any GNU/Linux Operating System.
+- Any GNU/Linux Operating System, or macOS 12.3 or later.
 - python 3.10.x and python-pip
+- python3.10-venv
 - sphinx 4.3.2
 - nodejs 14.5.0 or later
+- npm 16.19.1 or later
```
ravindk89, line 14: This is also what our current Hugo build for kes-docs require so we might as well bump it.

djwfyi (approved):
LGTM

README.md
```diff
@@ -34,13 +36,13 @@ python3 -m venv venv && source venv/bin/activate
 pip install -r requirements.txt && npm install && npm run build
 ```
 
-4. Build all targets at once:
+4. Build your desired platform target.
 
 ```
-make all
+make linux
 ```
```
ravindk89, line 44: Could add multi-builds. The output gets a bit hard to parse vs serially building, but

```
make linux k8s container
```

Is my personal standard batch when adding new tutorials.

harshavardhana (approved):

ravindk89 (approved):
LGTM mod nits

README.md
feorlen: This started working for me, unclear why but seems solid now. Adding multiple target example.

PR #759, opened 2023-03-15
Adding staging logic

ravindk89:
# Summary

The actual logic is pretty straightforward - run `mc cp -r` to recursively copy the contents of the correct directory out to a MinIO alias and bucket.

For now, the alias and bucket are hardcoded. We could make that more variable, but this is easier.

This will require you to install `mc` locally and ensure you have the `docs-staging` access setup . check slack for the details - if `mc ls docs-staging/staging` shows you a list of already staged docs, you're in good shape.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
 GITDIR        = $(shell git rev-parse --abbrev-ref HEAD)
+STAGINGURL    = http://192.241.195.202:9000/staging
 
 # Put it first so that "make" without argument is like "make help".
 help:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -8,6 +8,7 @@ SPHINXBUILD   ?= sphinx-build
 SOURCEDIR     = source
 BUILDDIR      = build
 GITDIR        = $(shell git rev-parse --abbrev-ref HEAD)
+STAGINGURL    = http://192.241.195.202:9000/staging
```
feorlen, line 11: Is this a globally shared URL? How does it work when multiple people do this at the same time?

feorlen: Differentiated by git branch, got it!

feorlen (approved):
LGTM

PR #782, opened 2023-03-29
Adds info on certificates with the MinIO Client

### LINKED ISSUE #781: Document the path for trusted certificates
Internal request to add information to document certificates with `mc` and any related configurations:

- [x] Add a new heading to the MinIO Client page for certificates.
- [x] Document that trusted certificates are stored at `~/.mc/certs/CAs/`
- [x] Recent versions prompt user about trusting new certificates (find version info)

djwfyi:
- Adds a new section on certificates to the MinIO Client page
- Adds a new section in Behaviors to the mc alias set page
- Minor fix to a --json flag command in version doc

Closes #781

Staged:
- http://192.241.195.202:9000/staging/certificates/reference/minio-mc.html#certificates
- http://192.241.195.202:9000/staging/certificates/reference/minio-mc/mc-alias-set.html#server-side-encryption

### DIFF
--- a/source/reference/minio-mc.rst
+++ b/source/reference/minio-mc.rst
@@ -418,6 +418,34 @@ The following list describes each possible file path location in the order
 
 You can use the ``--config-dir``
 
+.. _minio-mc-certificates:
+
+Certificates
+------------
+
+The MinIO Client stores certificates and CAs for deployments with :ref:`server side encryption <minio-sse>` to the following paths:
+
+Linux, MacOS, and other Unix-like systems:
+
+.. code-block:: shell
+
+   ~/.mc/certs/ # certificates
+   ~/.mc/certs/CAs/ # Certificate Authorities
+
+Windows systems:
+
+.. code-block:: shell
+
+   C:\Users\[username]\mc\certs\ # certificates
+   C:\Users\[username]\mc\certs\CAs\ # Certificate Authorities
+
+When creating a new :ref:`alias <minio-mc-alias>`, the MinIO Client prompts the user whether to accept the deployment's certificates.
+If you decide to trust the certificate, the MinIO Client adds the certificate to the above paths.
+
+.. note::
+
+   In testing environments, you can bypass the certificate check for selected mc commands by passing the ``--insecure`` flag.
+
 .. _minio-mc-global-options:
 
 Global Options
--- a/source/reference/minio-mc.rst
+++ b/source/reference/minio-mc.rst
@@ -423,7 +423,7 @@ You can use the ``--config-dir``
 Certificates
 ------------
 
-The MinIO Client stores certificates and CAs for deployments with :ref:`server side encryption <minio-sse>` to the following paths:
+The MinIO Client stores certificates and CAs for deployments to the following paths:
 
 Linux, MacOS, and other Unix-like systems:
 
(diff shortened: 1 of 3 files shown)


source/reference/minio-mc.rst
```diff
@@ -418,6 +418,34 @@ The following list describes each possible file path location in the order
 
 You can use the ``--config-dir``
 
+.. _minio-mc-certificates:
+
+Certificates
+------------
+
+The MinIO Client stores certificates and CAs for deployments with :ref:`server side encryption <minio-sse>` to the following paths:
```
ravindk89, line 426: I don't *think* sse matters here.

The `mc` client has to have a CA for the MinIO deployment it's connecting to - so that just requires the MinIO Server to enforce TLS.

If the MinIO Server TLS cert were self signed, you would need to put the public key into the mc CA directory (or something like that) to connect without `--insecure`.

For SSE I don't _think_ mc directly interfaces with KES. It would do so through MinIO's internal kes-go client, so the CA there wont matter as much.

source/reference/minio-mc/mc-alias-set.rst
```diff
@@ -198,4 +198,14 @@ For more complete documentation on S3 Access Control, see
...
+
+Server-Side Encryption
+~~~~~~~~~~~~~~~~~~~~~~
+
+For deployments with :ref:`server side encryption <minio-sse>` enabled, the MinIO Client asks the user whether to trust the certificates for the deployment when creating the alias.
+
+If trusted, the MinIO Client automatically adds the certificate and certificate authority to:
+
+-  ``~/.mc/certs/`` on Linux and other Unix-like systems.
+-  ``C:\Users\[username]\mc\certs`` on Windows systems.
```
ravindk89, line 211: Ah - this is the SSE-specific bit. So this only works for when we have SSE enabled?

What CA are we pulling down here? I assume the MinIO Deployment CA, not the KES or External KMS CA?

I'm a bit surprised we can download the MinIO CA at all - that sounds like new behavior?

ravindk89 (changes requested):
A few changes and follow-ups, possibly due to my own lack of knowledge on this subject matter.

source/reference/minio-mc.rst
djwfyi: Cut the SSE bit.

source/reference/minio-mc.rst
```diff
@@ -418,6 +418,34 @@ The following list describes each possible file path location in the order
...
+
+Windows systems:
+
+.. code-block:: shell
+
+   C:\Users\[username]\mc\certs\ # certificates
+   C:\Users\[username]\mc\certs\CAs\ # Certificate Authorities
+
+When creating a new :ref:`alias <minio-mc-alias>`, the MinIO Client prompts the user whether to accept the deployment's certificates.
+If you decide to trust the certificate, the MinIO Client adds the certificate to the above paths.
```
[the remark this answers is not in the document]
djwfyi: Yes, see https://github.com/minio/mc/pull/4010

source/reference/minio-mc/mc-alias-set.rst
djwfyi: Cleaned this up, as I was not understanding correctly.

ravindk89 (approved):
LGTM

PR #795, opened 2023-03-31
DOCS-660: Add note for Linux gateway migration docs pointing to Container docs

### LINKED ISSUE #660: Update Docker/Podman FS/Migration Guide to use container-specific procedure instead of generic
The disk migration error prompting now after a release links to this docs page: https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html which links me to here for making a new deployment: https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-single-node-single-drive.html#deploy-minio-standalone

This is a docker image. You really want me to mess around with systemd units in a docker image? I'm assuming not and the entire guide is just unsuitable for the (official) docker image, but then the docker image really should link a guide in this error message in the log that is suitable for it instead.

## Expected Behavior
Guide linked as startup error from docker image actually is applicable to docker image

## Current Behavior
Guide linked as startup error from docker image seems to only talk about things you'd do on a standalone system-wide install and not in a container

## Possible Solution
Link an applicable guide instead

## Steps to Reproduce (for bugs)
1. Migrate old docker image to new
2. Get error `minio_1      | ERROR Unable to use the drive /data: Drive /data: found backend type fs, expected xl or xl-single - to migrate to a supported backend visit https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html: Invalid arguments specified`
3. Click URL.

## Context
Currently, my website is down and I'm not sure how to migrate. I'll probably go back to old minio until I can actually get proper instructions on what I'm even supposed to be doing.

## Regression
Yes, regression or rather migration issue

## Your Environment

* Version used (`minio --version`): latest `minio/minio` docker image
* Server setup and configuration: Alpine Linux on the host, docker image uses `/data` volume mounted in and startup command `server /data` which used to work. The above error is entirely unhelpful in telling me what would now work.
* Operating System and version (`uname -a`): Alpine Linux 3.17 with docker 20.10.21 (no kubernetes or anything fancy)

ravindk89:
Closes #660 

There's no promise of a simple fix for being able to link to the container docs for when MinIO server runs in a containerized context. So the easiest thing to do is just link users to the Container docs, which themselves link to the correct deploy tutorials for a container context.

Worth noting that this already existed from the start, it's just that users may not think to change context from "Linux" to "Container" in the top-nav. Maybe we could explore JS to automatically inject the current page subpath into the nav bar when switching contexts for good measure, but we would also need to include some sort of 404-capture for when a page does not exist to at least route to the platform index. Problems for later.

### DIFF
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -20,6 +20,10 @@ Along with the deprecation announcement, MinIO also announced that the feature w
 As of :minio-release:`RELEASE.2022-10-29T06-21-33Z`, the MinIO Gateway and the related filesystem mode code have been removed.
 Deployments still using the `standalone` or `filesystem` MinIO modes that upgrade to :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later receive an error when attempting to start MinIO.
 
+.. cond:: linux
+
+   For deployments running in a container, see the `Container - Migrate from Gateway or Filesystem Mode <https://min.io/docs/minio/container/operations/install-deploy-manage/migrate-fs-gateway.html>`__ tutorial instead.
+
 Overview
 --------
 
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -22,7 +22,9 @@ Deployments still using the `standalone` or `filesystem` MinIO modes that upgrad
 
 .. cond:: linux
 
-   For deployments running in a container, see the `Container - Migrate from Gateway or Filesystem Mode <https://min.io/docs/minio/container/operations/install-deploy-manage/migrate-fs-gateway.html>`__ tutorial instead.
+   .. note::
+
+      For deployments running in a container, see the `Container - Migrate from Gateway or Filesystem Mode <https://min.io/docs/minio/container/operations/install-deploy-manage/migrate-fs-gateway.html>`__ tutorial instead.
 
 Overview
 --------


feorlen (approved):
YES!!

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -20,6 +20,10 @@ Along with the deprecation announcement, MinIO also announced that the feature w
 As of :minio-release:`RELEASE.2022-10-29T06-21-33Z`, the MinIO Gateway and the related filesystem mode code have been removed.
 Deployments still using the `standalone` or `filesystem` MinIO modes that upgrade to :minio-release:`RELEASE.2022-10-29T06-21-33Z` or later receive an error when attempting to start MinIO.
 
+.. cond:: linux
+
+   For deployments running in a container, see the `Container - Migrate from Gateway or Filesystem Mode <https://min.io/docs/minio/container/operations/install-deploy-manage/migrate-fs-gateway.html>`__ tutorial instead.
```
feorlen, line 25: I was hasty with the review button, should the `__` be there?

(But also yes this is very needed.)

djwfyi (approved):
Message is fine.
Consider putting it in a note admonition, maybe. Otherwise, LGTM.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi: Need a trailing underscore or two after links that don't have a role. So, yes, they need to be there.

djwfyi: When it's one or two underscores I have not yet figured out. 😄

ravindk89: The two underscores makes it an "anonymous" link. Without that it becomes a target link, and if any other link on the page matches that URL it breaks.

I find this annoying to keep track of so I just make all full URLs to be `__` by default.

djwfyi (approved):
LGTM

PR #799, opened 2023-04-04
Doc updates for mc RELEASE.2023-03-20T17-17-53Z

djwfyi:
- Adds note that `mc support proxy set|remove` no longer require registration
- Adds info about stale ARNs for `mc replicate status` command
- Additional information for --dry-run flag in `mc rm`
- Thoroughly updates and restructures `mc admin policy` commands
- Adds new `mc batch cancel` command
- Updates deprecated commands for newly deprecated mc admin policy functions
- Fixes toctree on mc admin page to allow for mc admin policy subcommand nesting

Closes #766

Other fixes not part of the release:
- Updates `--fake` flag on `mc rm` to `--dry-run`
- Updates `--fake` flag on `mc mirror` to `--dry-run`

PR #802, opened 2023-04-05
Correcting mc admin policy command links

djwfyi:
Missed some references to deprecated commands during an mc release doc update (PR #799).
This fixes those references to the new commands.


feorlen (approved):
👍🏻

PR #812, opened 2023-04-13
Fixed MINIO_ROOT_PASSWORD to be secret key

eko:
I’ve fixed a little typo that could save some minutes to users:

The MINIO_ROOT_PASSWORD is the secret key, not the access key.


djwfyi (approved):
Looks good!
Thanks for the fix.

PR #816, opened 2023-04-13
DOCS-784: Fix Kafka TLS Client Auth reference

### LINKED ISSUE #784: [BUG] Fix TLS_CLIENT_AUTH for Kafka
https://pkg.go.dev/crypto/tls#ClientAuthType

Our docs for [Kafka notifications](https://min.io/docs/minio/linux/reference/minio-server/minio-server.html#envvar.MINIO_NOTIFY_KAFKA_TLS_CLIENT_AUTH) suggest a string value for `TLS_CLIENT_AUTH`.

Looking closer at the actual Go library, it's an int. Which is odd because the Go library function is a series of const values.

So in order as defined by go it's 0 - 4, which I don't think anyone could possibly guess.

We'll need to update our docs as there's no easy wy to fix this without breaking it for others. Probably we need a table of the go values to the int values and hope Go doesn't change it anytime soon.

ravindk89:
Minor fix since apparently Go structs don't work the way I thought they did.

Can stage on request.

Closes #784

### DIFF
--- a/source/includes/common-mc-admin-config.rst
+++ b/source/includes/common-mc-admin-config.rst
@@ -846,7 +846,7 @@ MinIO supports the following mechanisms:
 .. start-minio-notify-kafka-tls-client-auth
 
 Specify the client authentication type of the Kafka broker(s).
-The following table lists the supported values and their mapping:
+The following table lists the supported values and their mappings
 
 .. list-table::
    :header-rows: 1
--- a/source/includes/common-mc-admin-config.rst
+++ b/source/includes/common-mc-admin-config.rst
@@ -845,12 +845,34 @@ MinIO supports the following mechanisms:
 
 .. start-minio-notify-kafka-tls-client-auth
 
-Specify the client authentication policy of the Kafka broker(s). See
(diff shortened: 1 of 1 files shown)


source/includes/common-mc-admin-config.rst
```diff
@@ -845,12 +845,34 @@ MinIO supports the following mechanisms:
 
 .. start-minio-notify-kafka-tls-client-auth
 
-Specify the client authentication policy of the Kafka broker(s). See
-`ClientAuthType <https://golang.org/pkg/crypto/tls/#ClientAuthType>`__ for 
-more information on possible values for this field.
+Specify the client authentication type of the Kafka broker(s).
+The following table lists the supported values and their mapping:
```
djwfyi, line 849: > mappings

Should be plural, I think?

djwfyi (approved):
One minor suggestion, otherwise LGTM.

source/includes/common-mc-admin-config.rst
ravindk89: ...I don't know for certain, but lets do it anyways.

```suggestion
The following table lists the supported values and their mappings
```

PR #820, opened 2023-04-14
Moves limit upload and limit download flags out of global list

### LINKED ISSUE #787: Update info for --limit-download and --limit-upload flags
`--limit-upload` and `--limit-download` only apply to
- mc mv
- mc cp
- mc mirror

The flags only limit work down on the client side.

> `--limit-upload `and `--limit-download` is for `mc cp`, `mc mirror` and `mc mv` for clients to throttle their upload/download

- [x] Remove the commands from the Global Flags list on the MC Client page.
- [x] Use includes to add the flags to the above command pages

djwfyi:
Uses an include in case we need them for other commands later.

Closes #787

Also update mc mv to use standard optional and required formatting for params.

Staged: http://192.241.195.202:9000/staging/limitflags/reference/minio-mc.html#global-options

### DIFF
--- a/source/includes/linux/minio-client.rst
+++ b/source/includes/linux/minio-client.rst
@@ -1,0 +1,49 @@
+.. start-mc-limit-flags-desc
+
+.. mc-cmd:: --limit-download
+   :optional:
+
+   Limit client-side download rates to no more than a specified rate in KiB/s, MiB/s, or GiB/s.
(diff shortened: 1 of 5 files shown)


ravindk89 (approved):
LGTM

feorlen (approved):
👏🏻

PR #822, opened 2023-04-18
Link to systemctl update docs in rpm/deb install instructions

feorlen:
The update instructions in the "You are running an older version" banner are not correct for systemd managed deployments (also the process recommended for rpm/deb installs). 

Link to the correct update procedure in the install docs. This is not intended to be comprehensive, just pointers for future-admin who might have forgotten.

Staged:
http://192.241.195.202:9000/staging/deb-upgrades/linux/html/operations/install-deploy-manage/deploy-minio-single-node-single-drive.html#download-the-minio-server

### DIFF
--- a/source/includes/common-minio-mc.rst
+++ b/source/includes/common-minio-mc.rst
@@ -42,9 +42,10 @@ commands *may* work as documented, any such usage is at your own risk.
 
 - Brackets ``[]`` indicate optional parameters. 
 - Parameters sharing a line are mutually dependent.
-- Parameters seperated using the pipe ``|`` operator are mutually exclusive.
+- Parameters separated using the pipe ``|`` operator are mutually exclusive.
 
 Copy the example to a text editor and modify as-needed before running the
 command in the terminal/shell.
+You may need to use ``sudo`` if your user does not have write permissions for the path where ``mc`` is installed.
 
 .. end-minio-syntax
--- a/source/includes/linux/common-installation.rst
+++ b/source/includes/linux/common-installation.rst
@@ -1,11 +1,9 @@
 .. start-install-minio-binary-desc
 
-The following tabs provide examples of installing MinIO onto 64-bit Linux
-operating systems using RPM, DEB, or binary. The RPM and DEB packages
-automatically install MinIO to the necessary system paths and create a
-``systemd`` service file for running MinIO automatically. MinIO strongly
-recommends using RPM or DEB installation routes. To update deployments
-managed using ``systemd``, see :ref:`minio-upgrade-systemctl`.
+The following tabs provide examples of installing MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary.
+The RPM and DEB packages automatically install MinIO to the necessary system paths and create a ``minio`` service for ``systemctl``.
+MinIO strongly recommends using the RPM or DEB installation routes.
+To update deployments managed using ``systemctl``, see :ref:`minio-upgrade-systemctl`.
 
 .. dropdown:: amd64 (Intel or AMD 64-bit processors)
    :open:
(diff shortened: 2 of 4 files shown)


source/includes/common-minio-mc.rst
```diff
@@ -42,9 +42,10 @@ commands *may* work as documented, any such usage is at your own risk.
 
 - Brackets ``[]`` indicate optional parameters. 
 - Parameters sharing a line are mutually dependent.
-- Parameters seperated using the pipe ``|`` operator are mutually exclusive.
+- Parameters separated using the pipe ``|`` operator are mutually exclusive.
```
feorlen, line 45: Drive-by spelling fix.

source/includes/common-minio-mc.rst
```diff
@@ -42,9 +42,10 @@ commands *may* work as documented, any such usage is at your own risk.
...
 
 Copy the example to a text editor and modify as-needed before running the
 command in the terminal/shell.
+You may need to use ``sudo`` if your user does not have write permissions for the path where ``mc`` is installed.
```
feorlen, line 49: Another helpful reminder, since we recommend updating `mc` after updating MinIO.

Is this a good place to add this text?

source/includes/linux/common-installation.rst
```diff
@@ -4,8 +4,8 @@ The following tabs provide examples of installing MinIO onto 64-bit Linux
 operating systems using RPM, DEB, or binary. The RPM and DEB packages
 automatically install MinIO to the necessary system paths and create a
 ``systemd`` service file for running MinIO automatically. MinIO strongly
-recommends using RPM or DEB installation routes.
-
+recommends using RPM or DEB installation routes. To update deployments
```
feorlen, line 7: One of two proposed locations for the update link.

The title of the linked section references `systemctl` so I used that. Using both `systemd` and `systemctl` here seems a bit confusing, but so would calling it `systemd` and landing on a page about `systemctl`. Suggestions welcome.

ravindk89: I think just calling it `systemctl` is fine since that is the proc that actually handles the automation.

Maybe for the graph on 4-6 we could state something like "create a ``minio`` service for ``systemctl``"

feorlen: Went with both your suggestions. And joined sentences since I was in there messing with existing text anyway.

What about the other instances of this same original paragraph? Edit or leave them alone?

ravindk89: :thinking: if you want to expand scope and hit the other ones, go for it. Might as well while we're here, so to speak.

source/includes/linux/deploy-standalone.rst
```diff
@@ -1,11 +1,10 @@
 1) Download the MinIO Server
 ----------------------------
 
-The following tabs provide examples of installing MinIO onto 64-bit Linux
-operating systems using RPM, DEB, or binary. The RPM and DEB packages
-automatically install MinIO to the necessary system paths and create a
-``systemd`` service file for running MinIO automatically. MinIO strongly
-recommends using RPM or DEB installation routes.
+The following tabs provide examples of installing MinIO onto 64-bit Linux operating systems using RPM, DEB, or binary.
+The RPM and DEB packages automatically install MinIO to the necessary system paths and create a ``minio`` service for ``systemctl``.
```
ravindk89, line 5: completely optional but it occurs to me we could just move this graf into the tab set and specify RPM or DEB respectively.

Then it only shows up when selecting the relevant tab. For binary builds, the user doesn't care about this information.

ravindk89 (approved):
LGTM mod one suggestion

source/includes/linux/deploy-standalone.rst
feorlen: After discussion, decided to leave it where it is so it's more visible. While the audience I had in mind wanted RPM/DEB and only got `systemctl` because we recommended it, there are other possible ways to trip over this.

PR #831, opened 2023-04-25
Include port number for the example MINIO_SERVER_URL

ktarbet:
The example environment variable setup does not include port 9000.  As a new user this tripped me up.


source/includes/common/common-deploy.rst
```diff
@@ -26,7 +26,7 @@ The following example provides a starting environment file:
 
    # Uncomment the following line and replace the value with the correct hostname for the local machine.
```
ravindk89, line 27:
```suggestion
Uncomment the following line and replace the value with the correct hostname for the local machine and port for the MinIO server (9000 by default).
```

PR #835, opened 2023-04-26
`kubectl minio init` command arguments update

pjuarezd:
Documentation update

closes https://github.com/minio/operator/issues/1568

### DIFF
--- a/source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
+++ b/source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
@@ -83,7 +83,7 @@ The command supports the following flags:
    :optional:
 
    The image to use when deploying the :minio-git:`MinIO Console <console>` in Operator mode, where administrators can create and manage MinIO tenants using a Graphical User Interface.
-   Defaults to the `version bundled in variable DefaultOperatorImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
+   Defaults to the `version bundled in variable DefaultOperatorImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`__.
 
 .. mc-cmd:: --console-tls
    :optional:
@@ -98,13 +98,13 @@ The command supports the following flags:
    :optional:
 
    The default :minio-git:`kes <kes>` image to use when creating a new MinIO tenant. 
-   Defaults to the `version bundled in variable DefaultKESImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
+   Defaults to the `version bundled in variable DefaultKESImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`__.
 
 .. mc-cmd:: --default-minio-image
    :optional:
 
    The default :minio-git:`minio <minio>` image to use when creating a new MinIO tenant. 
-   Defaults to the `version bundled in variable DefaultTenantImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
+   Defaults to the `version bundled in variable DefaultTenantImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`__.
 
 .. mc-cmd:: --image
    :optional:
--- a/source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
+++ b/source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
@@ -83,7 +83,7 @@ The command supports the following flags:
    :optional:
 
    The image to use when deploying the :minio-git:`MinIO Console <console>` in Operator mode, where administrators can create and manage MinIO tenants using a Graphical User Interface.
-   Defaults to ``minio/console:v0.17.3``.
+   Defaults to the `version bundled in variable DefaultOperatorImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
 
 .. mc-cmd:: --console-tls
    :optional:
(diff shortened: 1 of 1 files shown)


source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
```diff
@@ -83,7 +83,7 @@ The command supports the following flags:
    :optional:
 
    The image to use when deploying the :minio-git:`MinIO Console <console>` in Operator mode, where administrators can create and manage MinIO tenants using a Graphical User Interface.
-   Defaults to ``minio/console:v0.17.3``.
+   Defaults to the :minio-git:`latest release of the operator <operator/releases/latest>`.
```
ravindk89, line 86:
```suggestion
   Defaults to the version bundled with the `operator release version <operator/releases/latest>`__ matching the plugin version.
```

Since someone could use this doc with an older `kubectl minio` version, and the console version would be whatever is attached to that version.

I don't think it's necessary to keep a matrix, just mention they are bundled.

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
```diff
@@ -94,23 +94,17 @@ The command supports the following flags:
...
-
-   The default :minio-git:`MinIO Console <console>` image to use when creating a new MinIO tenant. 
-   Defaults to ``minio/console:v0.17.3``.
-
 .. mc-cmd:: --default-kes-image
    :optional:
 
    The default :minio-git:`kes <kes>` image to use when creating a new MinIO tenant. 
-   Defaults to ``minio/kes:v0.18.0``.
+   Defaults to the :minio-git:`latest release of KES <kes/releases/latest>`.
```
ravindk89, line 101: Same comment - e.g. https://github.com/minio/operator/blob/v5.0.4/kubectl-minio/cmd/helpers/constants.go

So it's not guaranteed to be current latest, just whatever is associated to that release.

If this file is generally the right place to look, we could cheat a little bit in the conf and auto-inject the latest release tag to make it easier to link.

Otherwise, it's enough to modify similar to suggestion before, e.g. "defaults to the version bundled in the matching Operator release. See <URL> for more information"

ravindk89 (changes requested):
A few change requests. I think we might be able to cheat a bit if we know for certain that the constants.go file is always what we reference when setting defaults.

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
pjuarezd: right, default value will always be whatever is on the constants.go file, on this case the `DefaultKESImage` env variable.

do you have an example on how could I auto-inject the value from a file?

ravindk89: I have to play with this a bit - I'm not sure if there's an easy way to pull and parse that file.

Right now in Makefile we have a dependency builder that snags stuff with wget and parses it from there. 

it would be nicer if this was a YAML of all of our defaults/constants, but that is for later.

For right now we can maybe say...

```
Defaults to the `version bundled in the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio//cmd/helpers/constants.go>`__
```

pjuarezd: applied, and mentioned the variable that holds the value

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
pjuarezd: mentioned the variable that holds the value

ravindk89 (approved):

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
```diff
@@ -83,7 +83,7 @@ The command supports the following flags:
    :optional:
 
    The image to use when deploying the :minio-git:`MinIO Console <console>` in Operator mode, where administrators can create and manage MinIO tenants using a Graphical User Interface.
-   Defaults to ``minio/console:v0.17.3``.
+   Defaults to the `version bundled in variable DefaultOperatorImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
```
djwfyi, line 86: Needs a trailing double underscore `__` before the `.` at the end. Otherwise Sphinx won't make it a link.

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
```diff
@@ -94,23 +94,17 @@ The command supports the following flags:
...
-
-   The default :minio-git:`MinIO Console <console>` image to use when creating a new MinIO tenant. 
-   Defaults to ``minio/console:v0.17.3``.
-
 .. mc-cmd:: --default-kes-image
    :optional:
 
    The default :minio-git:`kes <kes>` image to use when creating a new MinIO tenant. 
-   Defaults to ``minio/kes:v0.18.0``.
+   Defaults to the `version bundled in variable DefaultKESImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
```
djwfyi, line 101: Link needs a trailing `__` to make it a link.

source/reference/kubectl-minio-plugin/kubectl-minio-init.rst
```diff
@@ -94,23 +94,17 @@ The command supports the following flags:
...
 
 .. mc-cmd:: --default-minio-image
    :optional:
 
    The default :minio-git:`minio <minio>` image to use when creating a new MinIO tenant. 
-   Defaults to ``minio/minio:RELEASE.2022-05-26T05-48-41Z``.
+   Defaults to the `version bundled in variable DefaultTenantImage for the matching Operator release <https://github.com/minio/operator/blob/master/kubectl-minio/cmd/helpers/constants.go>`.
```
djwfyi, line 107: Link needs a trailing `__` to make it a link.

djwfyi (changes requested):
Three fixes for Sphinx links, otherwise LGTM.
I'll do those real quick.

djwfyi (approved):
LGTM

PR #841, opened 2023-05-02
Adds STS info for Operator

### LINKED ISSUE #792: Operator: SDKs can authenticate IAM-style to the Operator
Operator release v5.0.0 added [PR #1414](https://github.com/minio/operator/pull/1414) which allows for STS style uses for authenticating with the Operator.

We now support applications inside a kubernetes cluster to access MinIO with an IAM-style assume role mechanism.

1. Clients should read `AWS_WEB_IDENTITY_TOKEN_FILE` which points to the SA JWT token on the pod
2. client SDK should send the JWT over to the operator via custom endpoint which should be on the Operator
3. Operator should return an STS session if the service-account provided via the pod SA JWT has been assigned any policy

- [ ] Create a K8s-specific Developer doc for this functionality

More details are available at https://github.com/miniohq/engineering/issues/929.

djwfyi:
Creates a new page for STS tokens specifically for Operator.

Closes #792

Staged: http://192.241.195.202:9000/staging/k8s-sts/developers/sts-for-operator.html


source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. important:: 
+
+   MinIO Operator launches with STS *disabled* by default.
+   To use STS with the Operator, you must first explicitly enable it.
+
+How STS Authorization Works in Kubernetes
+-----------------------------------------
+
+An application can use an ``AssumeRole`` call to send a request for temporary credentials to the MinIO Operator.
```
pjuarezd, line 27:
```suggestion
An application can use an ``AssumeRoleWithWebIdentity`` call to send a request for temporary credentials to the MinIO Operator.
```

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+   - Create a :ref:`Policy Binding <minio-operator-sts-policy-binding>` linking the application to the MinIO Tenant's policy
+
+4. Apply the yaml file to create the resources on the deployment
+   
+   .. code-block:: shell
+      :class: copyable
+
+      kubectl apply -k path/to/yaml/file.yaml
+
+5. Use an SDK that supports the ``AssumeRole`` like behavior to send a call from your application to the deployment
```
pjuarezd, line 78:
```suggestion
5. Use an SDK that supports the ``AssumeRoleWithWebIdentity`` like behavior to send a call from your application to the deployment
```

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+   Some SDKs that support ``AssumeRole`` include:
```
pjuarezd, line 80:
```suggestion
   Some SDKs that support ``AssumeRoleWithWebIdentity`` include:
```

pjuarezd (changes requested):
Is possible to include the diagram?

https://github.com/minio/operator/blob/master/docs/images/sts-diagram.png

djwfyi:
> Is possible to include the diagram?
> 
> https://github.com/minio/operator/blob/master/docs/images/sts-diagram.png

I've asked @oscarocastellanos to brandify it, then will add it when he's done.

source/index.rst
```diff
@@ -185,6 +185,7 @@ Any file uploaded to ``play`` should be considered public and non-protected.
 
       /developers/minio-drivers
       /developers/security-token-service
+      /developers/sts-for-operator
```
ravindk89, line 188: Move this down to where line 203 is and it *should* work.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+-----------------------------------------
+
+An application can use an ``AssumeRoleWithWebIdentity`` call to send a request for temporary credentials to the MinIO Operator.
+The Operator checks the validity of the request, retrieves policies for the application, obtains credentials from the tenant, and then passes the credentials back the application.
+The application uses the issued credentials to work with the object storage on the tenant.
+
+The complete process includes the following steps:
+
+1. An application sends an API request to the MinIO Operator containing the name of the application and a service account to use.
+2. The MinIO Operator uses the Kubernetes API to check that the JWT is a valid request.
```
feorlen, line 34: Is there a "first reference" expansion for `JWT` nearby?

oscarocastellanos:
Here's the diagram @djwfyi @pjuarezd [sts-diagram.zip](https://github.com/minio/docs/files/11385073/sts-diagram.zip)
sts-diagram
sts-diagram

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+1. Enable STS functionality for the deployment
+   
+   .. code-block:: shell
+      :class: copyable
+
+      kubectl -n minio-operator set env deployment/minio-operator OPERATOR_STS_ENABLED=on
+   
+   - Replace ``minio-operator`` with the namespace for your deployment.
+   - Replace ``deployment/minio-operator`` with the value for your MinIO deployment
```
feorlen, line 61: How would someone know what `the value for your MinIO deployment` should be? Would this be obvious to someone who knows Kubernetes? (Because that person is most assuredly not me.)

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+2. Ensure an appropriate policy exists on the MinIO Tenant for the application to use
+
+3. Create a yaml document to add necessary resources:
+   - :ref:`Service Account <minio-operator-sts-service-account>` in the MinIO Tenant for the application to use
```
feorlen, line 66: Perhaps an article before `service account` here? Seems either `a` or `the` would work.

source/developers/sts-for-operator.rst
pjuarezd: good point @feorlen , JWT is a JSON web Token, but could expand the description of where it comes from, somenthing like this:


The STS API expects a JSON Web Token (JWT), it is a `Kubernetes service Account Token <https://kubernetes.io/docs/reference/access-authn-authz/authentication/#service-account-tokens>`.

When a pod have a Kubernetes service account associated to it, it will have the JWT  mounted in a well known location in the filesystem, that i  "/var/run/secrets/kubernetes.io/serviceaccount/token".

Take for example the following kubernetes manifest:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  namespace: sts-client
  name: my-service-account
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  namespace: default
spec:
  replicas: 3
  template:
    metadata:
    # ...
    spec:
      serviceAccountName: my-service-account
      containers:
      - name: nginx
        image: nginx:1.14.2

```
In the first yaml document of the manifest a service account `my-service-account` is created.
In the second yaml document a Deployment is created, and is associated to the service account `my-service-account` in the `spec.spec.serviceAccountName` field.

source/developers/sts-for-operator.rst
pjuarezd: For the mayority of the deployments, the operator is going to be on the `minio-operator` namespace by default.

On Openshift deployments coming from [OperatorHub](https://operatorhub.io/operator/minio-operator) the default namespace is `openshift-operators`.

Otherwise, users could use a different namespace with the `kubectl minio init --namespace other-namespace` command https://min.io/docs/minio/kubernetes/upstream/reference/kubectl-minio-plugin/kubectl-minio-init.html#kubectl.minio.init.-namespace.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+.. versionadded:: Operator v5.0.0
+
+   The MinIO Operator supports a set of API calls that allows an application to obtain STS credentials for a MinIO Tenant.
+
+:ref:`STS credentials <minio-security-token-service>` allow an application to access objects on a MinIO Tenant without the need to create any credentials for the application on the tenant.
```
feorlen, line 17: `any` could maybe be dropped here, it doesn't feel like it's adding much.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+   
+   .. code-block:: shell
+      :class: copyable
+
+      kubectl -n minio-operator set env deployment/minio-operator OPERATOR_STS_ENABLED=on
+   
+   - Replace ``minio-operator`` with the namespace for your deployment.
+   - Replace ``deployment/minio-operator`` with the value for your MinIO deployment
+
+2. Ensure an appropriate policy exists on the MinIO Tenant for the application to use
```
ravindk89, line 63: Link to MinIO IAM policy here?

source/developers/sts-for-operator.rst
feorlen: @pjuarezd ooo nice extended description. A bit long to fit in this list, but maybe we can find somewhere useful. @djwfyi ? 

Thanks!

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+
+.. default-domain:: minio
+
+.. contents:: Table of Contents
+   :local:
+   :depth: 2
+
+.. versionadded:: Operator v5.0.0
+
+   The MinIO Operator supports a set of API calls that allows an application to obtain STS credentials for a MinIO Tenant.
```
ravindk89, line 15: Q: @pjuarezd what problem is this solving in K8s vs sending the API request to the tenant itself?

Also, does the `AssumeRoleWithWebIdentity` API remain the same? How does the application specify the tenant it wants to auth against?

ravindk89 (changes requested):
Mostly clarifying questions.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,132 @@
...
+6. The MinIO Operator forwards the temporary credentials back to the application.
+7. The application uses the credentials to send the object storage calls to the MinIO tenant.
+
+Requirements
+------------
+
+STS for the MinIO Operator requires the following:
+
+- MinIO Operator v5.0.0 or later
+- Deployment **must** have :ref:`TLS configured <minio-tls>`.
```
feorlen, line 47: I keep wanting to read this as `The deployment`. I'll let you decide if adding the definite article works in this list.

feorlen: Also a tiny punctuation nit: period at the end of list items, or not?

source/developers/sts-for-operator.rst
pjuarezd: this is a excelent question a bit longer to answer, sorry for the long text.

`Operator STS API` solves some problems:

* Authenticate applications in kubernetes to access MinIO using a kubernetes native authentication mechanism: using kubernetes Service Accounts Tokens, the `Service Account` is a core concept of the Role-based Access Control (RBACK) Native kubernetes Authenticatio and authorization.

* Allows `Infrastructure as code` like configuration to access MinIO in kubernetes, by using  the Tenant CRD + PolicyBinding CRD.

* Allows a "credential-less" authentication in the client code, no need to create and specify IAM user to gain access to MinIO API, just instantiate the SDK client and get access to the 

We had often been asked to provide a way to get credentials for an application in a "credentialless" way, similar to how AWS IMDS allows to get the STS Token of the instance profile assigned to the EC2 instance. [AWS doc example here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html#instance-metadata-security-credentials)

We also have been asked to provide an mechanism to authenticate using kubernetes Service accounts similar to [AWS IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html), Operator STS does it.

ravindk89: Thanks @pjuarezd - so that comes back to how to use the API itself. From what I'm reading in a previous comment, the application would need to have access to the JWT from Kubernetes itself - so that JWT has to be passed to the applications somehow and then maintained in case it ever changes.

that still leaves a question of how the Operator knows which Tenant to generate credentials for. Is it purely by the service account token being namespaced already?

pjuarezd: yeah, regarding that second question

> Also, does the AssumeRoleWithWebIdentity API remain the same? How does the application specify the tenant it wants to auth against?

The request is not the same that doing it directly to the tenant, as the tenant knows nothing about kuberneters RBAC, Operator will do that authentication, and then request an STS token to the tenant using root credentials, somenthing like this:

Authentication with service Accoun

ravindk89: That makes sense. I guess here then the one requirement is we have to assume the K8s admin has a way of passing that SA JWT to the application and keeping it up-to-date, otherwise things crap out. @djwfyi I would suggest mentioning something to that effect if it's not there already and I just dumb brained past it.

pjuarezd: The bridge to know which tenant the sdk App is getting credentials for is the PolicyBinding Custom resource.

1) The client specify the namespace of the tenant in the STS endpoint, ie using the AWS SDK code would look like this

```python
stsUrl = urlparse(f"https://sts.minio.operator.svc.cluster.local:4223/tenant-1")
sts = boto3.client('sts', endpoint_url=stsUrl.geturl(), verify=False)
assumed_role_object = sts.assume_role_with_web_identity(
    RoleArn="arn:aws:iam::111111111:dummyroot",
    RoleSessionName="session 1",
    WebIdentityToken=sa_jwt
)

s3_client = boto3.resource('s3',
    aws_access_key_id=credentials['AccessKeyId'],
    aws_secret_access_key=credentials['SecretAccessKey'],
    aws_session_token=credentials['SessionToken'],
    endpoint_url=tenantUrl.geturl(), verify=False)
```
2) a `PolicyBinding` object have to exist in the tenant namespace, and specify service account name of the JWT, and a list of IAM policies to authorize.


```yaml
apiVersion: sts.min.io/v1alpha1
kind: PolicyBinding
metadata:
  name: binding-1
  namespace: tenant-1
spec:
  application:
    namespace: application-namespace
    serviceaccount: aplication-service-account
  policies:
    - bucket1-rw
    - bucket2-read
```


Because the PolicyBinding is a Custom resource and can be specified as yaml, this opens the door to Infrastructure as code authentication, managed by the kubernetes manager. I am thinking of ArgoCD maybe?

pjuarezd: > That makes sense. I guess here then the one requirement is we have to assume the K8s admin has a way of passing that SA JWT to the application and keeping it up-to-date, otherwise things crap out. @djwfyi I would suggest mentioning something to that effect if it's not there already and I just dumb brained past it.

we not necesarelly need to assume the k8s admin has a way of passing the SA JWT, because the application running in a pod in kubernetes is the one that will have to pass the JWT to the SDK's, for that, the pods always have the JWT   mounted in the filesystem  in the path `/var/run/secrets/kubernetes.io/serviceaccount/token`

I did a bit of explanation of where the JWT comes from in this comment https://github.com/minio/docs/pull/841#discussion_r1183996140

pjuarezd: STS are temporary credentials, that expire and need to be recreated, SDK's currently have a verification flow, if the STS token is expired, then sdk library "should" requests  a new STS token before the next API call, but that is up to how the SDK client is implemented.

SA JWT also has expiration, yes that token expiration is a whole topic, there is long-lived tokens and short-lived tokens, automatic mounted tokens, manually created, etc. But we could just limit ourselfs to require a `valid, non expired Service account Token` regardless if is the one mounted in the pod, or another requested using K8s API by the SDK application.

This medium article might do a better explaining to Service account JWT that me https://medium.com/@th3b3ginn3r/understanding-service-accounts-in-kubernetes-e9d2abe19df8

source/developers/sts-for-operator.rst
djwfyi: Thanks for this, @pjuarezd.
I expanded the first instance of JWT, then included portions of this description in several places.
For example, part of it in the last step of the procedure where the SDK call happens. And more of it in the Service Account description under Example Resources.

source/developers/sts-for-operator.rst
djwfyi: Added some instructions here for namespace and deployment values.

source/developers/sts-for-operator.rst
djwfyi: Yeah, and a period at the end. 😬

source/developers/sts-for-operator.rst
djwfyi: Agreed

source/developers/sts-for-operator.rst
djwfyi: This is all great info. Added applicable changes throughout the PR.

source/developers/sts-for-operator.rst
djwfyi: Me and listing ending punctuation have a long and fraught history that lacks anything resembling consistency.
Fixed both here.

feorlen (approved):
👍🏻

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+   
+   .. code-block:: shell
+      :class: copyable
+
+      kubectl apply -k path/to/yaml/file.yaml
+
+6. Use an SDK that supports the ``AssumeRoleWithWebIdentity`` like behavior to send a call from your application to the deployment
+
+   The STS API expects a JWT for the service account to exist in the Kubernetes environment.
+   This is typically in a well known location on the deployment, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```
pjuarezd, line 103: Maybe is useful mention that if you define the environment variable
`AWS_WEB_IDENTITY_TOKEN_FILE=/var/run/secrets/kubernetes.io/serviceaccount/token` then you don't need to read the JWT token file manually and pass it to the SDK client as parameter.

For example AWS SDK describes it here:
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html#envvars-list-AWS_WEB_IDENTITY_TOKEN_FILE

On the minio Go SDK we hare an example in the tests

https://github.com/minio/minio-go/blob/8c4c0768b1b1d29b52457d8e048144a8e2901009/pkg/credentials/iam_aws_test.go#L353-L355

not sure how much detailed or extended should be in the docs that mention, but for sure we can have code examples of this on the Operator repo, and just mention "oh you can do AWS_WEB_IDENTITY_TOKEN_FILE too" or somenthing like that.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+Example Resources
+-----------------
+
+.. _minio-operator-sts-service-account:
+
+Service Account
+~~~~~~~~~~~~~~~
+
+A Service Account is a :kube-docs:`Kubernetes resource type <reference/access-authn-authz/service-accounts-admin/>` that allows an external application to interact with the Kubernetes deployment.
+When linked to a pod, such as through a deployment's ``.spec.spec.serviceAccountName`` field, Kubernetes stores a :abbr:`JWT (JSON Web Token)` for the service account in a well-known location, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```
pjuarezd, line 124: Rather than save it as a file in the pod file system, k8s mounts it, like this for example

```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    app: nginx
  name: nginx-deployment-7fb96c846b-mvjhs
  namespace: default
spec:
  containers:
  - image: nginx:1.14.2
    name: nginx
...
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-vsvfz
      readOnly: true
...
  volumes:
  - name: kube-api-access-vsvfz
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
...
```

```suggestion
When linked to a pod, such as through a deployment's ``.spec.spec.serviceAccountName`` field, Kubernetes mounts a :abbr:`JWT (JSON Web Token)` for the service account in a well-known location, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```

pjuarezd (commented):
only suggestions, no hard requirements to change anything, wil let you do the switfness of wording it better than me @djwfyi

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+
+.. important:: 
+
+   MinIO Operator launches with STS *disabled* by default.
+   To use STS with the Operator, you must first explicitly enable it.
+
+How STS Authorization Works in Kubernetes
+-----------------------------------------
+
+An application can use an ``AssumeRoleWithWebIdentity`` call to send a request for temporary credentials to the MinIO Operator.
```
ravindk89, line 38: I think here is where we should state that the API call must include the ServiceAccount JWT, which applications deployed in the same cluster can access through the URL @pjuarezd mentioned.

Without that key element, users wont fully understand what to present.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+The application uses the issued credentials to work with the object storage on the tenant.
+
+.. image:: /images/k8s/sts-diagram.png
+   :width: 600px
+   :alt: A diagram showing STS token process flow on a Kubernetes MinIO deployment between the requesting application, MinIO Operator, Kubernetes API, PolicyBinding custom resource definition, and the MinIO tenant.
+   :align: center
+
+The complete process includes the following steps:
+
+1. An application sends an API request to the MinIO Operator containing the name of the application and a service account to use.
```
ravindk89, line 49: Or maybe this is where we need to clarify - from what I'm gathering from Pedro's comments, it's the same AssumeRoleWithWebToken API, but the JWT is the ServiceAccount token for the target tenant.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+
+      kubectl -n minio-operator set env deployment/minio-operator OPERATOR_STS_ENABLED=on
+   
+   - Replace ``minio-operator`` with the namespace for your deployment.
+   - Replace ``deployment/minio-operator`` with the value for your deployment's MinIO Operator.
+
+     You can find the deployment value by running ``kubectl get deployments -n <namespace>``, where you replace ``<namespace>`` with the namespace for the MinIO Operator.
+     Your MinIO Operator namespace is typically ``minio-operator``, though this value can change during install.
+
+2. Ensure an appropriate :ref:`policy <minio-policy>` exists on the MinIO Tenant for the application to use
```
ravindk89, line 84:
```suggestion
2. Ensure an appropriate :ref:`policy <minio-policy>` exists on the MinIO Tenant for the application to use.
   This procedure includes associating that policy to the ServiceAccount through a PolicyBinding object.
```

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+
+3. Create a yaml document to add necessary resources:
```
ravindk89, line 86:
```suggestion
3. Create YAML resources for the Service Account and Policy Binding: 
```

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+2. Ensure an appropriate :ref:`policy <minio-policy>` exists on the MinIO Tenant for the application to use
+
+3. Create a yaml document to add necessary resources: 
+
+   - The :ref:`Service Account <minio-operator-sts-service-account>` in the MinIO Tenant for the application to use.
+
+     For more on service accounts in Kubernetes, see the :kube-docs:`Kubernetes documentation <reference/access-authn-authz/service-accounts-admin/>`.
+   - Create a :ref:`Policy Binding <minio-operator-sts-policy-binding>` linking the application to the MinIO Tenant's policy.
+
+5. Apply the yaml file to create the resources on the deployment
```
ravindk89, line 93: :thought_balloon: @pjuarezd more follow-ups - if I have `minio-tenant-1` namespace and `application-1` namespace, which namespace gets the SA and policy binding?

From the examples below it looks like we apply it to the client/application namespace. Just want to confirm, because if so @djwfyi I suggest we state that here. Even if we have examples below, it's easy to miss. Should also state that in the examples, as by default w/o namespace those will end up in the default.

source/developers/sts-for-operator.rst
ravindk89, line 103:
```suggestion
   This is typically in a well known location on pods in the Kubernetes cluster, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+      :class: copyable
+
+      kubectl apply -k path/to/yaml/file.yaml
+
+6. Use an SDK that supports the ``AssumeRoleWithWebIdentity`` like behavior to send a call from your application to the deployment
+
+   The STS API expects a JWT for the service account to exist in the Kubernetes environment.
+   This is typically in a well known location on the deployment, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
+
+   Some SDKs that support ``AssumeRoleRoleWithWebIdentity`` include:
```
ravindk89, line 105:
```suggestion
   The following MinIO SDKs support ``AssumeRoleRoleWithWebIdentity`` :
```

ravindk89 (changes requested):
Strictly speaking these might not be necessary to go - you can take this in a separate PR if we want to get this out sooner. But I think it would help with removing a few more stumbling blocks.

source/developers/sts-for-operator.rst
pjuarezd: Good observation, the PolicyBinding resource has to be created in the tenant namespace, that is `minio-tenant-1`.
In Daryl's example it will be like this.

```yaml
apiVersion: sts.min.io/v1alpha1
kind: PolicyBinding
metadata:
  name: binding-1
  namespace: minio-tenant-1
spec:
  application:
    namespace: sts-client
    serviceaccount: stsclient-sa
  policies:
    - test-bucket-rw
```

The `application` we are granting access to is stated in the combination of `.spec.application.serviceaccount` and `.spec.application.namespace`, service accounts are namespace bounded, only one service account with certain name can exist in a namespace.

For instance, in the above Daryl's example, the service account with name `sts-client-sa` has to exist in the namespace `sts-client`,  and the manifest would be like this:

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  namespace: sts-client
  name: stsclient-sa
```

Now, there is no restriction of in which namespace the service account have to be in, it could be even in the same tenant namespace, it doesn't matter. The only that matters is that the PolicyBinding states correctly the name of the service account, and the namespace where it lives at.

ravindk89: Makes sense - I'm not particularly familiar with how much namespace isolation affects these things. I assume that as long as the application has the correct kube permissions (however those might work) to access the SA, the rest just sort of happens on its own.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+     Your MinIO Operator namespace is typically ``minio-operator``, though this value can change during install.
+
+2. Ensure an appropriate :ref:`policy <minio-policy>` exists on the MinIO Tenant for the application to use
+
+3. Create a yaml document to add necessary resources: 
+
+   - The :ref:`Service Account <minio-operator-sts-service-account>` in the MinIO Tenant for the application to use.
+
+     For more on service accounts in Kubernetes, see the :kube-docs:`Kubernetes documentation <reference/access-authn-authz/service-accounts-admin/>`.
+   - Create a :ref:`Policy Binding <minio-operator-sts-policy-binding>` linking the application to the MinIO Tenant's policy.
```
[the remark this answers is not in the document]
djwfyi: Rephrased this line to include that info and also added an additional sentence to the Policy Binding section down the page.

source/developers/sts-for-operator.rst
djwfyi: Added some info too this effect in this step of the procedure.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,170 @@
...
+- Allows applications to access objects in MinIO tenants using a Kubernetes-native authentication mechanism.
+  
+  Service Accounts or Service Account Tokens are a core concept of :kube-docs:`Role-Based Access Control (RBAC) <reference/access-authn-authz/rbac/>` :kube-docs:`authentication <reference/access-authn-authz/authentication/#service-account-tokens>` in Kubernetes.
+
+- Implementing STS for MinIO Operator allows you to utilize infrastructure as code principles and configuration by using the tenant custom resource definition (CRD) and a MinIO PolicyBinding CRD.
+
+.. important:: 
+
+   MinIO Operator launches with STS *disabled* by default.
+   To use STS with the Operator, you must first explicitly enable it.
```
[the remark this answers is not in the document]
djwfyi: Tweaked slightly, but added.

source/developers/sts-for-operator.rst
djwfyi: I mentioned this earlier on the page now, so leaving it out here for now.

source/developers/sts-for-operator.rst
djwfyi: I'll add something to this effect, but I find it kind of awkward because the very next step does this.

source/developers/sts-for-operator.rst
djwfyi: Corrected my description of the YAML for the policy binding, which was wrong.
Also added inline comments to the YAML to better clarify what namespace to use.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,186 @@
...
+
+   The procedure on this page includes instructions to enable the STS API in the MinIO Operator.
+
+How STS Authorization Works in Kubernetes
+-----------------------------------------
+
+An application can use an ``AssumeRoleWithWebIdentity`` call including a Service Account's :abbr:`JWT (JSON Web Token)` to send a request for temporary credentials to the MinIO Operator.
```
ravindk89, line 40:
```suggestion
An application can use an ``AssumeRoleWithWebIdentity`` call including a Kubernetes Service Account's :abbr:`JWT (JSON Web Token)` to send a request for temporary credentials to the MinIO Operator.
```

Maybe link to the service account docs here

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,186 @@
...
+The JWT is typically mounted in the deployment from a well known location on the deployment, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```
ravindk89, line 41: From what Pedro said, it's more that

"Kubernetes mounts Service Accounts to pods specifying that SA using <whatever that field was>.
The Pod can access those SAs at a well known location such as <URL>"

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,186 @@
...
+The application uses the issued credentials to work with the object storage on the tenant.
+
+.. image:: /images/k8s/sts-diagram.png
+   :width: 600px
+   :alt: A diagram showing STS token process flow on a Kubernetes MinIO deployment between the requesting application, MinIO Operator, Kubernetes API, PolicyBinding custom resource definition, and the MinIO tenant.
+   :align: center
+
+The complete process includes the following steps:
+
+1. An application sends an API request to the MinIO Operator containing the tenant namespace and a service account to use.
```
ravindk89, line 53:
```suggestion
1. An application sends an AssumeRoleWithWebidentity API request to the MinIO Operator containing the tenant namespace and a service account to use.
```

Link to the API docs.

source/developers/sts-for-operator.rst
```diff
@@ -0,0 +1,186 @@
...
+   
+   .. code-block:: shell
+      :class: copyable
+
+      kubectl apply -k path/to/yaml/file.yaml
+
+5. Use an SDK that supports the ``AssumeRoleWithWebIdentity`` like behavior to send a call from your application to the deployment
+
+   The STS API expects a JWT for the service account to exist in the Kubernetes environment.
+   This is typically mounted from a well known location on pods in the Kubernetes cluster, such as ``/var/run/secrets/kubernetes.io/serviceaccount/token``.
```
ravindk89, line 109: For pods which specify the service account using <that field>, Kubernetes typically mounts that JWT at a well known location such as <thing>

ravindk89 (approved):
Mostly nits, but LGTM once you take/leave them.

pjuarezd (approved):
LGTM!

Issue #798, opened 2023-04-03
[FEATURE] - Document how to deploy MinIO via ArgoCD/GitOps with Helm


cniackz:
**Is your feature request related to a problem? Please describe.**

This is not related to a problem but to a necesity of helping users in a better way, lately I have seen more and more users using ArgoCD and they all have similar questions when performing the deploy. And all of them want to use `helm` on top of ArgoCD, during those meetings, I have seen how this is possible and it can work, but I never had a chance to do it my-self, so I just have part of the procedure. At the end, what I am requesting is to start formal documentation on how this can be achieved, so that when next customer comes with same questions, we can simple redirect the customer to read our Documentation on ArgoCD integration with Helm.

**Describe the solution you'd like**
A clear and concise description of what you want to happen: To have formal documentation on how to deploy MinIO via ArgoCD/Helm.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered: I have heard from some customers, that Kustomize could be an option and even OperatorHub, but at the end all of the customers I have been in touch with, are using just Helm, so if someone knows more from Kustomize and ArgoCD or OperatorHub and ArgoCD, that would be the alternatives.

**Additional context**
Add any other context or screenshots about the feature request here: I just have a small wiki to show how to use Helm and that has been working for old versions of Operator and MinIO and for new versions as well, this wiki can help get ArgoCD working but it is just half of the equation, as I would like to see ArgoCD part also documented and some screenshots on how this can be used, maybe we can start with some initial tags or places to expose this and then we can share and get feedback from the users: https://github.com/cniackz/public/wiki/How-to-install-MinIO-Using-Helm-in-Kubernetes

### Tested locally, you can get steps from:

https://github.com/minio/wiki/wiki/How-to-deploy-MinIO-with-ArgoCD-in-k8s

ravindk89:
@kannappanr one thing to follow-up here is on the helm side for tenant deployment. we can in detail talk about values.yaml and document the mapping, but ideally we would find a way to do this via annotations and auto-generating some level of reference. Otherwise it is far too easy for the docs to fall out of sync w/ the core product.

PR #851, opened 2023-05-12
Helm deploy docs

feorlen:
Document how to deploy with a Helm chart, based on [How to install MinIO Using Helm in Kubernetes](https://github.com/cniackz/public/wiki/How-to-install-MinIO-Using-Helm-in-Kubernetes). 

This addition maintains the existing plugin instructions as the primary path, with Helm as a child page under that. That makes room for future deploy options. 

The existing `common-install-operator-kubectl-plugin-install.rst` is now in two parts, to allow only including the second part in the Helm instructions. Also promoted the Upgrade page to be a sibling to Deploy.

Many of the original instructions are environment setup that enable installing the Helm charts, starting Operator, and creating a tenant. That will vary by the user's environment, so omitted for docs.

Staged:
http://192.241.195.202:9000/staging/DOCS-798/k8s/html/operations/install-deploy-manage/deploy-operator-helm.html

Fixes https://github.com/minio/docs/issues/798


source/includes/common/common-install-operator-kubectl-plugin-install-init.rst
```diff
@@ -0,0 +1,32 @@
+1) Install the MinIO Kubernetes Plugin
```
feorlen, line 1: First half from original Operator instructions using the plugin

source/includes/common/common-install-operator-kubectl-validate-open-console.rst
```diff
@@ -1,36 +1,3 @@
...
-  of the operator. Specify the 
-  :mc-cmd:`kubectl minio init --cluster-domain` argument to set a 
-  different :kube-docs:`cluster domain 
-  <tasks/administer-cluster/dns-custom-nameservers/>` value.
-
-.. important::
-
-   Document all arguments used when initializing the MinIO Operator.
-
 3) Validate the Operator Installation
```
feorlen, line 1: Second half from original Operator instructions using the plugin

source/includes/k8s/deploy-operator.rst
```diff
@@ -57,6 +57,8 @@ The tenant utilizes Persistent Volume Claims to talk to the Persistent Volumes t
    :alt: A diagram of the namespaces and pods used by or maintained by the MinIO Operator.
    :align: center
 
+.. _minio-operator-prerequisites:
```
feorlen, line 60: So I can link back to here

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+
+Prerequesites
+-------------
+
+To install the Operator with Helm you will need the following:
+
+* An existing cluster using Kubernetes 1.19.0 or later (1.21.0 or later recommended).
+* The ``kubectl`` CLI tool on your local host, the same version as the cluster.
+* `Helm <https://helm.sh/docs/intro/install/>`__ version xx or greater.
+* `yq <https://github.com/mikefarah/yq/#install>`__ version xx or greater.
```
feorlen, line 30: I don't know if there are minimum versions of Helm or yq needed.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+* Access to run ``kubectl`` commands on the cluster from your local host.
+
+For more about Operator installation requirements, including TLS certificates, see the :ref:`Operator deployment prerequisites <minio-operator-prerequisites>`.
```
feorlen, line 33: Production deployments likely need this additional info/context from the main Operator deploy page. How much is worth explicitly referencing/duplicating here?

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+
+
+Procedure
+---------
+
+
+Install Operator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Download the Helm charts
```
feorlen, line 47: There's a lot of nested stuff here, looking for ideas for making it easier to navigate. I'm not fond of nested lists, but is that better than several more levels of headers?

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+
+         kubectl get service console -n minio-operator -o yaml > service.yaml
+         yq e -i '.spec.type="NodePort"' service.yaml
+         yq e -i '.spec.ports[0].nodePort = PORT_NUMBER' service.yaml
+
+      Replace ``PORT_NUMBER`` with the port on which to serve the Operator GUI. 
+
+      The file contents resemble the following:
+
+      .. dropdown:: Example ``service.yaml`` file
```
feorlen, line 89: Seems important to show example output, but these yq output examples are long. Hence dropdown. Ideas?

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+         apiVersion: v1
+         kind: Secret
+         metadata:
+           name: console-sa-secret
+           namespace: minio-operator
+           annotations:
+             kubernetes.io/service-account.name: console-sa
+         type: kubernetes.io/service-account-token
+
+   #. ``kubectl apply`` the YAML files to apply the configuration to your deployment:
```
feorlen, line 262: Is this an accurate way to describe what you are doing with kubectl here?

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+
+   .. code-block:: shell
+      :class: copyable
+
+      helm install \
+      --namespace tenant-ns \
+      --create-namespace \
+      tenant-ns tenant-|operator-version-stable|.tgz
+
+#. Expose the tenant Console port
```
feorlen, line 294: This section is from `includes/k8s/quickstart.rst`. It could be made into an include, although there's already an awful lot of includes going on in the Operator pages.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,316 @@
...
+   .. note::
+      
+      The following steps of this procedure assume an active ``kubectl port-forward`` command.
+
+      To configure long term access to the pod, configure :kube-docs:`Ingress <concepts/services-networking/ingress/>` or similar network control components within Kubernetes to route traffic to and from the pod. Configuring Ingress is out of the scope for this documentation.
+
+#. Login to the MinIO Console
+
+   Access the tenant :ref:`minio-console` by opening a browser on the local machine and navigating to ``http://127.0.0.1:9090``.
+   Log in to the Console with the credentials ``myminio | minio123``.
```
feorlen, line 316: These lines are also (condensed) from `includes/k8s/quickstart.rst` is there something better elsewhere? `myminio | minio123` appears to be default credentials, but only existing mention is in a spark/hadoop integrations page. I'm not clear why this isn't minioadmin/minioadmin.

feorlen: After this in the original steps from @cniackz there was another `kubectl port-forward` to enable `mc`. I never got it to work, recommendations appreciated. It would be good to include.

(Also I need to test all this once more to confirm nothing broke in translation to rst.)

source/includes/common/common-install-operator-kubectl-validate-open-console.rst
djwfyi, line 1: The font inside the dropdowns for nodePort and Port Forwarding gets smaller.
@rushenn can you take a look?

This may be intentional, because it's inside a component. Just threw me off when seeing it today.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,315 @@
...
+
+You can deploy a MinIO Tenant using either the :ref:`Operator Console <minio-operator-console>` or Helm.
+To deploy a Tenant with the Console, see :ref:`Deploy and Manage MinIO Tenants <minio-installation>`.
+
+To deploy a Tenant with Helm:
+
+#. The following Helm command creates a MinIO Tenant:
+
+   .. code-block:: shell
+      :class: copyable
```
djwfyi, line 286:
```suggestion
      :class: copyable
      :substitutions:
```

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,315 @@
...
+
+   .. note::
+      
+      The following steps of this procedure assume an active ``kubectl port-forward`` command.
+
+      To configure long term access to the pod, configure :kube-docs:`Ingress <concepts/services-networking/ingress/>` or similar network control components within Kubernetes to route traffic to and from the pod. Configuring Ingress is out of the scope for this documentation.
+
+#. Login to the MinIO Console
+
+   Access the Tenant :ref:`minio-console` by opening a browser on the local machine and navigating to ``http://127.0.0.1:9090``.
```
djwfyi, line 314:
```suggestion
   Access the Tenant's :ref:`minio-console` by opening a browser on the local machine and navigating to ``http://127.0.0.1:9090``.
```

source/operations/install-deploy-manage/deploy-operator-helm.rst
djwfyi: The default access key and secret values are defined in this file in the Helm folder:
https://github.com/minio/operator/blob/a524d928933034e4dfb20437568081a0b7826ab3/helm/tenant/values.yaml#L6

source/operations/install-deploy-manage/deploy-operator-helm.rst
djwfyi: I think the dropdowns are a good solution for this.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,315 @@
...
+      :class: copyable
+
+      helm install \
+      --namespace minio-operator \
+      --create-namespace \
+      minio-operator operator-|operator-version-stable|.tgz
+
+#. Configure the MinIO deployment
+
+   #. Create the YAML configuration files
```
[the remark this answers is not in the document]
djwfyi: @rushenn is this something that CSS could help us with when there are nested ordered lists?

source/operations/install-deploy-manage/deploy-operator-helm.rst
djwfyi: I think you have sufficient information with appropriate linking of where to go for more info. It's fine from my view.

djwfyi (changes requested):
Overall, this is really great capture of the content.
I agree with you that the nesting needs work. Since we use headings for that kind of thing for other procedures, maybe go with that. But I've pinged Rushan in my comments in case there's some magic he can work in the CSS for the nested ordered lists.

I do think we need some bullets for at least the three files that need to be created. And ... some other comments I left but can't remember to summarize now.

Will take a look again after another pass.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,315 @@
...
+   
+      Use ``kubectl`` and ``yq`` to create the following files:
+
+      ``service.yaml``:
```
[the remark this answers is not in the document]
feorlen: I first tried an unordered list and was meh about it as well. I'll cogitate on that while we see if there is some CSS magic possible.

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: Oh yeah much better 👍🏻

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: Aha! I presume you could update that if you knew about it. Like maybe if the docs mentioned it. 😄

feorlen: @cniackz what is important to know about modifying the contents of our Helm charts before using? Is that a normal thing to do?

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,315 @@
...
+
+      curl -O https://raw.githubusercontent.com/minio/operator/master/helm-releases/operator-|operator-version-stable|.tgz
+      curl -O https://raw.githubusercontent.com/minio/operator/master/helm-releases/tenant-|operator-version-stable|.tgz
+
+#. Deploy Operator
+
+   The following Helm command deploys the MinIO Operator using the downloaded chart:
+
+   .. code-block:: shell
+      :class: copyable
```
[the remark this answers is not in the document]
feorlen: Whoops! Good catch.

source/operations/install-deploy-manage/deploy-operator-helm.rst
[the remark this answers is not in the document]
feorlen: This is what it looks like with numbers, digits, then unordered list

Example from rendered docs page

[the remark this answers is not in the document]
feorlen: That's a lot of gray boxes. Maybe leaving the list item title (`operator.yaml:`) in body text would be better? But it _is_ a filename. 🤔 

Same as above, except the unordered list item titles are body text style

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: Added a brief mention about customization. I have more detail for a future update.

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: reworked this, as it presumes the plugin install method.

cniackz (changes requested):
Let's use `neat` to reduce the number of lines and display just relevant parts.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,410 @@
...
+                    schedulerName: default-scheduler
+                    securityContext:
+                      fsGroup: 1000
+                      runAsGroup: 1000
+                      runAsNonRoot: true
+                      runAsUser: 1000
+                    serviceAccount: minio-operator
+                    serviceAccountName: minio-operator
+                    terminationGracePeriodSeconds: 30
+              status:
```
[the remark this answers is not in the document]
cniackz: status is a dynamic side of k8s, and it keeps on changing and people does not need to apply these files with status, it is more an informative part of a living object. But for a static one, status shouldn't be present as is not needed.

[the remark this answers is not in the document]
feorlen: Is `neat` a common tool someone would likely already use or be familiar with? I'm reluctant to add another dependency, but I'd like to understand more about `neat`.

Another alternative is to not show the entire yaml, but only the "important bits." (Whatever we decide those might be.)

@ravindk89 @djwfyi Thoughts about including an optional dependency in these instructions? The example yaml files are what I got in my test environment, I presume ones from a full production setup would be longer.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,410 @@
...
+   .. code-block:: shell
+      :class: copyable
+      :substitutions:
+
+      helm install \
+      --namespace minio-operator \
+      --create-namespace \
+      minio-operator operator-|operator-version-stable|.tgz
+
+#. Configure the MinIO deployment
```
ravindk89, line 74: This isn't the MinIO deployment, right? We're creating services for the Operator pods, Operator console pods, and the console JWT secret.

source/operations/install-deploy-manage/deploy-operator-helm.rst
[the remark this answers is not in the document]
ravindk89: I have to admit having full YAMLs is dicey - they can also change between MinIO Operator versions, so we have a maintenance burden here.

I'm also a little confused here - why do we need to generate these YAML files and apply them? The `service.yaml` and `operator.yaml` are created from the existing console and operator resources. Then we re-apply them. What is the purpose of this?

Does the Operator Helm Chart not pre-create all of these required items @cniackz

ravindk89 (changes requested):
A few things to address. Still reviewing, but these came up as requiring some discussion.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,410 @@
...
+A `Helm chart <https://helm.sh/docs/topics/charts/>`__ is a set of YAML files, templates, and other files that define the deployment details.
+The following procedure uses a Helm Chart to install the :ref:`MinIO Kubernetes Operator <minio-operator-installation>` to a Kubernetes cluster.
+
+
+Prerequisites
+-------------
+
+To install the Operator with Helm you will need the following:
+
+* An existing cluster using Kubernetes 1.19.0 or later (1.21.0 or later recommended).
```
ravindk89, line 27: https://endoflife.date/kubernetes

Kubernetes has a *really* aggressive EOL schedule. Even so, I would suggest we maybe make this a replacement variable and use it consistently in our docs (`|recommended-k8s-version|` or something) and always pin to the one with active support (1.25 right now)

ravindk89 (commented):
A few things to address. Still reviewing, but these came up as requiring some discussion.

source/includes/eks/deploy-minio-on-elastic-kubernetes-service.rst
```diff
@@ -57,10 +57,6 @@ You can use this policy as a baseline for manual Operator installations.
 Procedure
 ---------
 
-.. include:: /includes/common/common-install-operator-kubectl-plugin.rst
+.. include:: /includes/common/common-install-operator-kubectl-plugin-install-init.rst
+.. include:: /includes/common/common-install-operator-kubectl-validate-open-console.rst
```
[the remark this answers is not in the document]
feorlen: Seeing as yesterday I yanked the thing that was the reason for include fiddling, there might be an obvious way to address this concern 🗑️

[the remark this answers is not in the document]
ravindk89: I defer to your judgement/experimentation here :D Ultimately this isn't much of an include, so if we save some complexity at the cost of some verbiage, I think that's fine.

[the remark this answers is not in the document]
feorlen: Reverted the split (by hand, Because Reasons).

source/includes/common/common-install-operator-kubectl-validate-open-console.rst
feorlen: This happens anywhere there's a dropdown as far as I can tell. (I didn't notice because font settings in my usual browser.) Opened https://github.com/minio/docs/issues/869 to track.

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: Ah yes. That might conceivably be relevant to the Tenant section, but not here. Perhaps "Configure Operator" would be better. I'll bring this up when we talk next week, hopefully I'll be more enlightened about this entire yaml section after that.

source/operations/install-deploy-manage/deploy-operator-helm.rst
feorlen: I guess we could parse the json version of that page. And by _we_ I mean someone who can operate `jq` better than I.

feorlen: Reworked text to send readers to the Prerequisites link for k8s version. Programmatic substitution can come later.

source/operations/install-deploy-manage/deploy-operator-helm.rst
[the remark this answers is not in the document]
feorlen: Punting on the CSS update for now. Going with option 2 and manually making `A.`, `B.`,  etc

source/operations/install-deploy-manage/deploy-operator-helm.rst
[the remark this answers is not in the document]
feorlen: Further discussion at https://github.com/minio/docs/pull/851#discussion_r1204731785

cniackz:
I had a meeting with Andrea, I think she might be right, and preserving the same YAML without neat might help end user to map this faster and achieve the Helm installation.

I leave then this to you and from my point of view, this PR looks good to me.

cniackz (approved):
LGTM

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,447 @@
...
+        kubectl port-forward svc/myminio-hl 9000 -n tenant-ns
+
+   * Create an alias for the Tenant service:
+
+     .. code-block:: shell
+	:class: copyable
+
+        mc alias set myminio https://localhost:9000 minio minio123 --insecure
+
+     This example uses the non-TLS ``myminio-hl`` service, which requires ``--insecure``.
```
feorlen, line 438: `--insecure` should be a link, but I haven't figured out how to do that.

source/operations/install-deploy-manage/deploy-operator-helm.rst
```diff
@@ -0,0 +1,447 @@
...
+      :substitutions:
+
+      helm install \
+      --namespace minio-operator \
+      --create-namespace \
+      minio-operator operator-|operator-version-stable|.tgz
+
+#. Configure Operator
+
+   A. Create the YAML configuration files
```
ravindk89, line 76: @cniackz just for my edification - are we just *persisting* the configurations to the namespace, as helm does not do that for us?

ravindk89 (approved):
Let's get this out there and do polish on another pass. Great work!

PR #862, opened 2023-05-25
Broken link to decommission page

djwfyi:
Corrects a link that resulted in a 404 from K8s tenant page to decommission page.


krisis (approved):
LGTM

Issue #818, opened 2023-04-14
[BUG] gateway/fs migration docs


poornas:
**Describe the bug**

Some misleading steps in the docs needs to be clarified/removed
https://min.io/docs/minio/linux/operations/install-deploy-manage/migrate-fs-gateway.html and https://blog.min.io/minio-gateway-migration/ 

- Export/Import the existing deployment’s configurations with `mc admin config export/import` only to fs. For gateway all config are in env. vars. MINIO_CACHE_* and MINIO_GATEWAY_SSE env are no longer relevant in the newer cluster - all other env.var can be brought in as is
- Export/Import bucket metadata and iam assets applies only to fs mode
- Tiering not available in gateway mode nor in FS mode. Tiering is only available in SNSD, SNMD and MNMD

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

ravindk89:
Based on initial walkthrough:

- gateway is envvar only and most should be 1:1 transfer.
- We can just drop the tiering section

Maybe tabs to deal with FS vs Gateway? Or make them two procedures on the same page with shared steps.

PR #894, opened 2023-06-23
Fix docs for gateway/filesystem deployment migration

feorlen:
The instructions to migrate a MinIO Gateway or filesystem mode deployment are incorrect. Separate out the steps that differ for each into tabs and clean up where needed. Several migration steps for filesystem mode don't apply to gateway, and tiers don't apply to either.

Staged:
http://192.241.195.202:9000/staging/DOCS-818/linux/html/operations/install-deploy-manage/migrate-fs-gateway.html

Fixes https://github.com/minio/docs/issues/818

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -278,7 +278,7 @@ Glossary
 
    standalone deployment
      A :term:`single-node single-drive` (SNSD) MinIO deployment.
-     This term previously referred to the deprecated :ref:`Gateway and Filesystem Mode <minio-gateway-migration>` deployment types.
+     This term previously referred to the deprecated :ref:`Gateway or Filesystem Mode <minio-gateway-migration>` deployment types.
 
    SUBNET
      `MinIO's Subscription Network <https://min.io/pricing?jmp=docs>`__ tracks support tickets and provides 24 hour direct-to-engineer access for subscribed accounts.
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -275,7 +275,11 @@ Glossary
      A deployment uses a single external key to decrypt any object throughout the deployment.
 
      See also: :term:`SSE-KMS`, :term:`SSE-C`, :term:`encryption at rest`, :term:`network encryption`.
-   
+
+   standalone deployment
+     A :term:`single-node single-drive` (SNSD) MinIO deployment.
+     This term previously referred to the deprecated :ref:`Gateway and Filesystem Mode <minio-gateway-migration>` deployment types.
+
    SUBNET
      `MinIO's Subscription Network <https://min.io/pricing?jmp=docs>`__ tracks support tickets and provides 24 hour direct-to-engineer access for subscribed accounts.
 
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -95,11 +95,10 @@ Procedure
 
          Migrate configuration settings:
 
-	 If your deployment uses :ref:`environment variables <minio-server-environment-variables>` for configuration settings, copy the environment variables from the existing deployment's ``/etc/default/minio` file to the same file in the new deployment.
+	 If your deployment uses :ref:`environment variables <minio-server-environment-variables>` for configuration settings, copy the environment variables from the existing deployment's ``/etc/default/minio`` file to the same file in the new deployment.
          You may omit any ``MINIO_CACHE_*`` and ``MINIO_GATEWAY_SSE`` environment variables, as these are no longer used.                                                               
 
-	 If you use :mc-cmd:`mc admin config set <mc admin config set>` for configuration settings, duplicate the existing settings for the new deployment.
-         You can examine runtime settings using ``env | grep MINIO_`` or, for deployments using MinIO's systemd service, check the contents of ``/etc/default/minio``.
+	 If you use :mc-cmd:`mc admin config set <mc admin config set>` for configuration settings, duplicate the existing settings for the new deployment using the new MinIO Client.
 
       .. tab-item:: Filesystem mode
 
--- a/source/operations/install-deploy-manage/migrate-fs-gateway.rst
+++ b/source/operations/install-deploy-manage/migrate-fs-gateway.rst
@@ -47,7 +47,6 @@ Procedure
    You can set MinIO configuration settings in environment variables and using :mc-cmd:`mc admin config set <mc admin config set>`.
    Depending on your current deployment setup, you may need to retrieve the values for both.
 
-   This procedure does not cover migrating environment variables due to the variety of configuration methods.
    You can examine any runtime settings using ``env | grep MINIO_`` or, for deployments using MinIO's systemd service, check the contents of ``/etc/default/minio``.
 
 #. Create a new Single-Node Single-Drive MinIO deployment.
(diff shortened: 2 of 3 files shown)


source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -73,128 +72,143 @@ Procedure
...
 
-   Use the :mc-cmd:`mc admin config export <mc admin config export>` export command with the existing MinIO Client to retrieve the configurations defined for the existing standalone MinIO deployment.
+   .. tab-set::
 
-   .. code-block:: shell
-      :class: copyable
+      .. tab-item:: Gateway
 
-      mc admin config export ALIAS > config.txt
+         a. Migrate environment variables
```
feorlen, line 81: I can't remember if we prefer `a` or `A` for second level ordered lists.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -73,128 +72,143 @@ Procedure
...
 
-   - Use the existing MinIO Client.
-   - Replace ``ALIAS`` with the alias used for the existing standalone deployment you are retrieving values from. 
+            Copy the :ref:`environment variables <minio-server-environment-variables>` from the existing deployment's ``/etc/default/minio`` file to the same file in the new deployment.
```
feorlen, line 83: Is there a more accurate way to describe this?

The existing instructions suggest the old and new deployments could be on the same host. (Explicitly discusses using the same filesystem, which may or may not be the same host.) I couldn't figure that out, in part because then how do you have different `/etc/default/minio` files. (Maybe there's a flag to specify?)

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -73,128 +72,143 @@ Procedure
...
+            You may omit any ``MINIO_CACHE_*`` and ``MINIO_GATEWAY_SSE`` environment variables, as these are no longer used.
 
-#. Import **configurations** from existing standalone deployment to new deployment with the new MinIO Client
+      .. tab-item:: Filesystem mode
```
feorlen, line 86: My understanding is the filesystem mode instructions are (mostly) ok. I haven't tested them yet, the old version of the install docs reference instructions on the download site that no longer exist so it's confusing.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
djwfyi, line 1: We should add a graph explaining the difference between `filesystem` mode and `gateway` mode. People might not know which tab to click on otherwise.

djwfyi (changes requested):
Tweaks and suggestions for your consideration.

source/glossary.rst
```diff
@@ -275,7 +275,11 @@ Glossary
      A deployment uses a single external key to decrypt any object throughout the deployment.
 
      See also: :term:`SSE-KMS`, :term:`SSE-C`, :term:`encryption at rest`, :term:`network encryption`.
-   
+
+   standalone deployment
+     A :term:`single-node single-drive` (SNSD) MinIO deployment.
+     This term previously referred to the deprecated :ref:`Gateway and Filesystem Mode <minio-gateway-migration>` deployment types.
```
feorlen, line 281: Now that I think about it, seems odd to add a term for `gateway` since it's gone. But `standalone` is still in use to mean SNSD. Searching for gateway should turn up this and hopefully clarify.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -61,7 +60,7 @@ Procedure
 
    Set the port to a custom point different than the existing standalone deployment.
 
-#. Add an alias for the new deployment with :mc:`mc alias set` with the new MinIO Client from the previous step
+#. Add an alias for the new deployment with :mc:`mc alias set` with the new MinIO Client from the previous step.
```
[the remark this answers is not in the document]
feorlen: Went with a minor tweak to your suggestion 👍🏻

source/operations/install-deploy-manage/migrate-fs-gateway.rst
feorlen: Excellent idea.

Added a short description for each. I went back to older versions of the docs for likely text, feedback appreciated on accuracy.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
poornas: >I couldn't figure that out, in part because then how do you have different /etc/default/minio files
you could save the env in different files - it is just that the service file should refer the correct env file - see [here](https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-single-node-single-drive.html#create-the-systemd-service-file)

```suggestion
            Copy the :ref:`environment variables <minio-server-environment-variables>` from the existing deployment's ``/etc/default/minio`` file to the EnvironmentFile specified in the new deployment's service file.
```

feorlen: So you can have more than one deployment if you are using systemd. (I'm not, my VMs have some sort of systemd problem.)

Manually incorporated your suggestion, thank you @poornas

ravindk89 (commented):
Not much more to add beyond Daryl's comments, as we're mostly nitpicking over how much scope to go into here.

source/glossary.rst
feorlen: You have better context than I do. I'm still fuzzy on how exactly they are related.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
feorlen: Whatever we do now is better than it was, so deal with that if/when it comes up.

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -73,128 +72,133 @@ Procedure
...
+               :class: copyable
 
-   The export includes the bucket metadata only.
-   No objects export from the existing deployment with this command.
+               mc admin config import ALIAS < config.txt
 
-   .. code-block:: shell
-      :class: copyable
+            - Use the new MinIO Client.
+            - Replace ``ALIAS`` with the alias for the new deployment.
```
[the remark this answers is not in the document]
feorlen: I think this captures your _or something_:
```
If :mc:`~mc admin config import` reports an error for a configuration key, comment it out with ``#`` at the beginning of the relevant line and try again.                    
When you are finished migrating the deployment, verify the current syntax for the target MinIO Server version and set any needed keys manually using :mc:`mc admin config set`.
```

source/operations/install-deploy-manage/migrate-fs-gateway.rst
```diff
@@ -73,128 +72,129 @@ Procedure
...
-   - Replace ``ALIAS`` with the alias for the new deployment.
+            .. code-block:: shell
+               :class: copyable
 
-#. Restart the server for the new deployment with the new MinIO Client
+               mc admin config export ALIAS > config.txt
 
-   .. code-block:: shell
-      :class: copyable
+            - Use the existing MinIO Client.
```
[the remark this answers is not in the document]
feorlen: The SNSD instructions referenced above cover installing `mc`, but it's on another page and a bit buried. Useful to mention it again here 👍🏻 

I _think_ the export commands are still ok, but of course that could change at any moment. Let me see if I can come up with a workable middle ground.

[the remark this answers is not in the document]
djwfyi: But does it cover adding two separate versions of mc at the same time? Because your instructions rely on having both versions accessible to run.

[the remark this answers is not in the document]
feorlen: I'm actually just adding that right now 👍🏻

poornas (approved):

ravindk89 (approved):
As per internal discussion, this is already an improvement and we can push this out and revisit later.

Issue #909, opened 2023-07-05
[RELEASE] MinIO SERVER RELEASE.2023-06-29T05-12-28Z


ravindk89:
See https://github.com/minio/minio/releases/tag/RELEASE.2023-06-29T05-12-28Z for full release notes

- [ ] ~New Lifecycle Config `ExpiredObjectAllVersions` in [PR17521](https://github.com/minio/minio/pull/17521)~ 
  ~However, this is not in `mc` yet, so might not be documentable - or would require creating a JSON rule and importing it from there~
- [ ] We now allow site replication for non-distributed (SNSD/SNMD) setups [PR 17524](https://github.com/minio/minio/pull/17524). We could make a simple note of this as a "You can do this in testing, but not recommended in production"
- [x] We could note that we now log "skipped" files in decommissioning as part of logging [PR 17522](https://github.com/minio/minio/pull/17522)

feorlen:
https://github.com/minio/minio/pull/17524 is noted as a bug fix, possibly for an edge case with test environments and multiple containers from the same image? The reported issue is a recent regression, unclear what/if docs changes are appropriate. Is there something helpful to say?

feorlen:
Engineering intends to update `mc ilm rule` so users won't need to import JSON to get the https://github.com/minio/minio/pull/17521 functionality.

PR #925, opened 2023-07-20
Decommissioning logs skipped delete markers and expired objects

feorlen:
Add mention of logging skipped items

Partially addresses https://github.com/minio/docs/issues/909

### DIFF
--- a/source/operations/install-deploy-manage/decommission-server-pool.rst
+++ b/source/operations/install-deploy-manage/decommission-server-pool.rst
@@ -170,7 +170,8 @@ Decommissioning Ignores Expired Objects and Trailing ``DeleteMarker``
 Starting with :minio-release:`RELEASE.2023-05-27T05-56-19Z`, decommissioning ignores objects where the only remaining version is a ``DeleteMarker``.
 This avoids creating empty metadata on the remaining server pool(s) for objects that are effectively fully deleted.
 
-Starting with :minio-release:`minio-lifecycle-management-scanner`, decommissioning also ignores object versions which have expired based on the configured :ref:`lifecycle rules <minio-lifecycle-management-expiration>` for the parent bucket.
+Starting with :minio-release:`RELEASE.2023-06-23T20-26-00Z`, decommissioning also ignores object versions which have expired based on the configured :ref:`lifecycle rules <minio-lifecycle-management-expiration>` for the parent bucket.
+Starting with :minio-release:`RELEASE.2023-06-29T05-12-28Z`, decommissioning logs ignored delete markers and expired objects.
 
 Once the decommissioning process completes, you can safely shut down that pool.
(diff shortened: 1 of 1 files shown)


source/operations/install-deploy-manage/decommission-server-pool.rst
```diff
@@ -170,7 +170,8 @@ Decommissioning Ignores Expired Objects and Trailing ``DeleteMarker``
 Starting with :minio-release:`RELEASE.2023-05-27T05-56-19Z`, decommissioning ignores objects where the only remaining version is a ``DeleteMarker``.
 This avoids creating empty metadata on the remaining server pool(s) for objects that are effectively fully deleted.
 
-Starting with :minio-release:`minio-lifecycle-management-scanner`, decommissioning also ignores object versions which have expired based on the configured :ref:`lifecycle rules <minio-lifecycle-management-expiration>` for the parent bucket.
+Starting with :minio-release:`RELEASE.2023-06-23T20-26-00Z`, decommissioning also ignores object versions which have expired based on the configured :ref:`lifecycle rules <minio-lifecycle-management-expiration>` for the parent bucket.
+Starting with :minio-release:`RELEASE.2023-06-29T05-12-28Z`, decommissioning logs ignored delete markers and expired objects.
```
feorlen, line 174: When a delete marker or expired object is skipped during decommissioning, is its name logged as well? From the diff, it looks like a generic message that something was skipped. I'm not sure if the name is added to the message elsewhere.
@harshavardhana

ravindk89, line 174:
```suggestion
Starting with :minio-release:`RELEASE.2023-06-29T05-12-28Z`, decommissioning-related logs ignore delete markers and expired objects.
```

Just to avoid the concept of decommissioning logs being distinct from the system logs (which I might be wrong about but I think it's all one log)

ravindk89 (approved):
One nit, otherwise LGTM pending the remaining engineering question.

source/operations/install-deploy-manage/decommission-server-pool.rst
feorlen: I'm gonna merge this with your suggestion, can update to make less vague if/when new details surface.

harshavardhana: We do not as in in a log we send a trace if someone is listening via `mc admin trace --call decom`

feorlen: Ok so this isn't logging at all. But it's viewable by tracing that call type (along with whatever was there already). 👍🏻

feorlen: reworked because not logging

PR #932, opened 2023-07-28
Correct ChaCha20Poly-1305 to ChaCha20-Poly1305

aofei:
It's [ChaCha20-Poly1305](https://en.wikipedia.org/wiki/ChaCha20-Poly1305), not ChaCha20Poly-1305. Sorry a little bit OCD.


ravindk89:
@aead just checking this isn't a Go-related string we need to keep as-is

ravindk89:
Merging, assuming this is fine.

Issue #919, opened 2023-07-17
Refresh AD/LDAP Docs


ravindk89:
A user in Community Slack pointed out issues following out AD/LDAP guidance.

In reviewing the [code](https://github.com/minio/minio/blob/master/internal/config/identity/ldap/config.go#L91) it looks like there has been a lot of drift. The initial AD/LDAP work was a straight migration from github, and I did not revisit it at that time or immediately thereafter (focus was OIDC/KeyCloak as far as external auth goes).

Unfortunately as it stands the docs wont get you very far, so this needs a complete retooling and likely some level of testing against an AD/LDAP server (We have a test bed somewhere internally for this, need to hunt it down)

# Goals

- [ ] Update reference documentation to be accurate to latest stable
- [ ] Update tutorial documentation to be accurate to latest stable

# Stretch Goals

- [ ] Validate against AD/LDAP setup (Stretch due to time/resource constraints)

PR #939, opened 2023-08-04
Updated LDAP docs

feorlen:
Revised instructions with the current recommended configuration options. 

Staged:
file:///Users/andrealongo/repos/docs/build/DOCS-919/linux/html/operations/external-iam/configure-ad-ldap-external-identity-management.html

Fixes https://github.com/minio/docs/issues/919

### DIFF
--- a/source/administration/console/security-and-access.rst
+++ b/source/administration/console/security-and-access.rst
@@ -145,6 +145,8 @@ Configuring an external IDP enables Single-Sign On workflows, where applications
 Use the the screens in this section to view, add, or edit OIDC configurations for the deployment.
 MinIO supports any number of active OIDC configurations.
 
+.. _minio-console-admin-identity-ldap:
+
 LDAP
 ~~~~
 
--- a/source/administration/identity-access-management/ad-ldap-access-management.rst
+++ b/source/administration/identity-access-management/ad-ldap-access-management.rst
@@ -57,7 +57,7 @@ Access Keys are long-lived credentials which inherit their privileges from the p
 The parent user can further restrict those privileges while creating the access keys. 
 Use either of the following methods to create a new access keys:
 
-- Log into the :ref:`MinIO Console <minio-console>` using the AD/LDAP-managed user credentials. From the :guilabel:`Identity` section of the left navigation, select :guilabel:`Access Keys` followed by the :guilabel:`Create access keys +` button.
+- Log into the :ref:`MinIO Console <minio-console>` using the AD/LDAP-managed user credentials. From the :guilabel:`User` section of the left navigation, select :guilabel:`Access Keys` followed by the :guilabel:`Create access keys +` button.
 
 - Use the :mc:`mc admin user svcacct add` command to create the access keys. Specify the user Distinguished Name as the username to which to associate the access keys.
 
--- a/source/includes/common-minio-external-auth.rst
+++ b/source/includes/common-minio-external-auth.rst
@@ -163,7 +163,7 @@ provider configuration.
 
 Specify the hostname for the Active Directory / LDAP server. For example:
 
-``https://ldapserver.com:636``
+``ldapserver.com:636``
 
 .. end-minio-ad-ldap-server-addr
 
--- a/source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
+++ b/source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
@@ -10,6 +10,7 @@ Configure MinIO for Authentication using Active Directory / LDAP
    :local:
    :depth: 2
 
+
 Overview
 --------
 
@@ -29,7 +30,9 @@ The procedure on this page provides instructions for:
    - Accessing the MinIO Console using AD/LDAP credentials.
    - Using the MinIO ``AssumeRoleWithLDAPIdentity`` Security Token Service (STS) API to generate temporary credentials for use by applications.
 
-This procedure is generic for AD/LDAP services. Defer to the documentation for the AD/LDAP provider of your choice for specific instructions or procedures on configuration of user identities.
+This procedure is generic for AD/LDAP services.
+Defer to the documentation for the AD/LDAP provider of your choice for specific instructions or procedures on configuration of user identities.
+
 
 Prerequisites
 -------------
@@ -57,7 +60,6 @@ Instructions on configuring AD/LDAP are out of scope for this procedure.
      This may require configuration or deployment of additional Kubernetes network components and/or enabling access to the public internet.
 
 MinIO requires a read-only access keys with which it :ref:`binds <minio-external-identity-management-ad-ldap-lookup-bind>` to perform authenticated user and group queries.
-
 Ensure each AD/LDAP user and group intended for use with MinIO has a corresponding :ref:`policy <minio-external-identity-management-ad-ldap-access-control>` on the MinIO deployment. 
 An AD/LDAP user with no assigned policy *and* with membership in groups with no assigned policy has no permission to access any action or resource on the MinIO cluster.
 
@@ -85,7 +87,7 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no 
    Install and Configure ``mc`` with Access to the MinIO Cluster
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   This procedure uses :mc:`mc` for performing operations on the MinIO cluster. 
+   This procedure uses :mc:`mc` for performing operations on the MinIO cluster.
    Install ``mc`` on a machine with network access to the cluster.
    See the ``mc`` :ref:`Installation Quickstart <mc-install>` for instructions on downloading and installing ``mc``.
 
@@ -99,12 +101,12 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no 
 
    .. include:: /includes/k8s/steps-configure-ad-ldap-external-identity-management.rst
 
-
 .. Doing this the quick and dirty way. Need to revise later to be proper full includes via stepfiles
 
 .. cond:: linux or container or macos or windows
 
    .. _minio-external-identity-management-ad-ldap-configure:
+
 
    Procedure
    ---------
@@ -112,27 +114,63 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no 
    1) Set the Active Directory / LDAP Configuration Settings
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   You can configure the AD/LDAP provider using either
-   environment variables *or* server runtime configuration settings. Both
-   methods require starting/restarting the MinIO deployment to apply changes. The
-   following tabs provide a quick reference of all required and optional
-   environment variables and configuration settings respectively:
+   Configure the AD/LDAP provider using one of the following:
+
+   * MinIO Client
+   * environment variables
+   * MinIO Console
+
+   All methods require starting/restarting the MinIO deployment to apply changes.
+
+   The following tabs provide a quick reference of the three LDAP configuration methods:
 
    .. tab-set::
 
+      .. tab-item:: MinIO Client
+
+         MinIO supports specifying the AD/LDAP provider settings using :mc:`mc idp ldap` commands.
+
+	 Confirm this:
+         For distributed deployments, the :mc:`mc idp ldap` command applies the configuration to all nodes in the deployment. 
+
+         The following example code sets *all* configuration settings related to configuring an AD/LDAP provider for external identity management.
+	 The minimum *required* settings are:
+
+         - :mc-conf:`identity_ldap server_addr <identity_ldap.server_addr>`
+         - :mc-conf:`identity_ldap lookup_bind_dn <identity_ldap.lookup_bind_dn>`
+         - :mc-conf:`identity_ldap lookup_bind_password <identity_ldap.lookup_bind_password>`
+         - :mc-conf:`identity_ldap user_dn_search_base_dn <identity_ldap.user_dn_search_base_dn>`
+         - :mc-conf:`identity_ldap user_dn_search_filter <identity_ldap.user_dn_search_filter>`
+
+        .. code-block:: shell
+           :class: copyable
+
+	   mc idp ldap add ALIAS                                                   \
+	      server_addr="ldaps.example.net:636"                                  \
+              lookup_bind_dn="CN=xxxxx,OU=xxxxx,OU=xxxxx,DC=example,DC=net"        \
+	      lookup_bind_password="xxxxxxxx"                                      \
+	      user_dn_search_base_dn="DC=example,DC=net"                           \
+	      user_dn_search_filter="(&(objectCategory=user)(sAMAccountName=%s))"  \
+	      group_search_filter= "(&(objectClass=group)(member=%d))"             \
+	      group_search_base_dn="ou=MinIO Users,dc=example,dc=net"              \
+	      server_insecure=on
+
+        For more complete documentation on these settings, see :mc:`mc idp ldap`.
+
+	.. admonition:: :mc:`mc idp ldap` recommended
+           :class: note
+
+           :mc:`mc idp ldap` offers additional features and improved validation over :mc-cmd:`mc admin config set` runtime configuration settings.
+           The :mc-conf:`identity_ldap` configuration key remains available for existing scripts and tools.
+
       .. tab-item:: Environment Variables
 
-         MinIO supports specifying the AD/LDAP provider
-         settings using :ref:`environment variables
-         <minio-server-envvar-external-identity-management-ad-ldap>`. The 
-         :mc:`minio server` process applies the specified settings on its next
-         startup. For distributed deployments, specify these settings across all
-         nodes in the deployment using the *same* values consistently.
+         MinIO supports specifying the AD/LDAP provider settings using :ref:`environment variables <minio-server-envvar-external-identity-management-ad-ldap>`.
+	 The :mc:`minio server` process applies the specified settings on its next startup.
+	 For distributed deployments, specify these settings across all nodes in the deployment using the *same* values consistently.
 
-         The following example code sets *all* environment variables related to
-         configuring an AD/LDAP provider for external
-         identity management. The minimum *required* variable are:
-         
+         The following example code sets *all* environment variables related to configuring an AD/LDAP provider for external identity management. The minimum *required* variable are:
+
          - :envvar:`MINIO_IDENTITY_LDAP_SERVER_ADDR`
          - :envvar:`MINIO_IDENTITY_LDAP_LOOKUP_BIND_DN`
          - :envvar:`MINIO_IDENTITY_LDAP_LOOKUP_BIND_PASSWORD`
@@ -150,50 +188,24 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no 
             export MINIO_IDENTITY_LDAP_GROUP_SEARCH_FILTER="(&(objectClass=group)(member=%d))"
             export MINIO_IDENTITY_LDAP_GROUP_SEARCH_BASE_DN="ou=MinIO Users,dc=example,dc=net"
 
-         For complete documentation on these variables, see
-         :ref:`minio-server-envvar-external-identity-management-ad-ldap`
+            export MINIO_IDENTITY_LDAP_STS_EXPIRY="1h"
+            export MINIO_IDENTITY_LDAP_USERNAME_FORMAT="uid=%s,cn=miniousers,dc=myldapserver,dc=net,userPrincipalName=%s,cn=miniousers,dc=myldapserver,dc=net"
+	    export MINIO_IDENTITY_LDAP_TLS_SKIP_VERIFY="off"
+	    export MINIO_IDENTITY_LDAP_SERVER_INSECURE="off"
+	    export MINIO_IDENTITY_LDAP_SERVER_STARTTLS="off"
+	    export MINIO_IDENTITY_LDAP_COMMENT="LDAP test server"
 
-      .. tab-item:: Configuration Settings
+         For complete documentation on these variables, see :ref:`minio-server-envvar-external-identity-management-ad-ldap`
 
-         MinIO supports specifying the AD/LDAP provider
-         settings using :mc-conf:`configuration settings <identity_ldap>`. The 
-         :mc:`minio server` process applies the specified settings on its next
-         startup. For distributed deployments, the :mc:`mc admin config`
-         command applies the configuration to all nodes in the deployment.
+      .. tab-item:: MinIO Console
 
-         The following example code sets *all* configuration settings related to
-         configuring an AD/LDAP provider for external
-         identity management. The minimum *required* setting are:
-         
-         - :mc-conf:`identity_ldap server_addr <identity_ldap.server_addr>`
-
-         - :mc-conf:`identity_ldap lookup_bind_dn <identity_ldap.lookup_bind_dn>`
-
-         - :mc-conf:`identity_ldap lookup_bind_password <identity_ldap.lookup_bind_password>`
-         
-         - :mc-conf:`identity_ldap user_dn_search_base_dn <identity_ldap.user_dn_search_base_dn>`
-         
-         - :mc-conf:`identity_ldap user_dn_search_filter <identity_ldap.user_dn_search_filter>`
-
-         .. code-block:: shell
-            :class: copyable
-
-            mc admin config set ALIAS/ identity_ldap \
-               server_addr="ldaps.example.net:636" \
-               lookup_bind_dn="CN=xxxxx,OU=xxxxx,OU=xxxxx,DC=example,DC=net" \
-               lookup_bind_password="xxxxxxxx" \
-               user_dn_search_base_dn="DC=example,DC=net" \
-               user_dn_search_filter="(&(objectCategory=user)(sAMAccountName=%s))" \
-               group_search_filter= "(&(objectClass=group)(member=%d))" \
-               group_search_base_dn="ou=MinIO Users,dc=example,dc=net" 
-
-         For more complete documentation on these settings, see
-         :mc-conf:`identity_ldap`.
+	 link to something about Console
+	 :ref:`minio-console-admin-identity-ldap`
 
    2) Restart the MinIO Deployment
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   You must restart the MinIO deployment to apply the configuration changes. 
+   You must restart the MinIO deployment to apply the configuration changes.
    Use the :mc-cmd:`mc admin service restart` command to restart the deployment.
 
    .. code-block:: shell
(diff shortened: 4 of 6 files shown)


source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,76 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+	      group_search_base_dn="ou=MinIO Users,dc=example,dc=net"              \
+	      server_insecure=on
+
+        For more complete documentation on these settings, see :mc:`mc idp ldap`.
+
+	.. admonition:: :mc:`mc idp ldap` recommended
+           :class: note
+
+           :mc:`mc idp ldap` offers additional features and improved validation over :mc-cmd:`mc admin config set` runtime configuration settings.
+           The :mc-conf:`identity_ldap` configuration key remains available for existing scripts and tools.
```
feorlen, line 164: Need to decide what to say about configuring with `identity_ldap`. I haven't seen that it's formally deprecated, but the newer `mc idp ldap` is preferred.

My understanding is doing it with the configuration key _should_ emit a warning. Not sure the status of that. See https://github.com/minio/minio/issues/17790

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +188,54 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
             export MINIO_IDENTITY_LDAP_GROUP_SEARCH_BASE_DN="ou=MinIO Users,dc=example,dc=net"
 
-         For complete documentation on these variables, see
-         :ref:`minio-server-envvar-external-identity-management-ad-ldap`
+            export MINIO_IDENTITY_LDAP_STS_EXPIRY="1h"
+            export MINIO_IDENTITY_LDAP_USERNAME_FORMAT="uid=%s,cn=miniousers,dc=myldapserver,dc=net,userPrincipalName=%s,cn=miniousers,dc=myldapserver,dc=net"
+	    export MINIO_IDENTITY_LDAP_TLS_SKIP_VERIFY="off"
+	    export MINIO_IDENTITY_LDAP_SERVER_INSECURE="off"
+	    export MINIO_IDENTITY_LDAP_SERVER_STARTTLS="off"
+	    export MINIO_IDENTITY_LDAP_COMMENT="LDAP test server"
```
feorlen, line 196: TODO: confirm these example values are reasonable

source/administration/console/security-and-access.rst
```diff
@@ -145,6 +145,8 @@ Configuring an external IDP enables Single-Sign On workflows, where applications
 Use the the screens in this section to view, add, or edit OIDC configurations for the deployment.
 MinIO supports any number of active OIDC configurations.
 
+.. _minio-console-admin-identity-ldap:
+
 LDAP
```
feorlen, line 150: This Console page is fairly vague, contemplating what to add about current LDAP configuration that isn't too out of character with the rest of it.

See also https://github.com/minio/console/issues/2967

feorlen: The Console steps in the main LDAP page is partly an include, which could be used here as well.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,83 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+	      user_dn_search_filter="(&(objectCategory=user)(sAMAccountName=%s))"  \
+	      group_search_filter= "(&(objectClass=group)(member=%d))"             \
+	      group_search_base_dn="ou=MinIO Users,dc=example,dc=net"              \
+              enabled="true"                                                       \
+              sts_expiry="1h"                                                      \
+              username_format="uid=%s,cn=miniousers,dc=myldapserver,dc=net,userPrincipalName=%s,cn=miniousers,dc=myldapserver,dc=net"                                            \
+              tls_skip_verify="off"                                                \
+              server_insecure=off                                                  \
+              server_starttls="off"                                                \
+	      comment="Test LDAP server"
```
feorlen, line 161: This example contains all the relevant envvars/configuration key settings. Is this/these sample values reasonable?

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,83 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+
+        For more complete documentation on these settings, see :mc:`mc idp ldap`.
+
+	.. admonition:: :mc:`mc idp ldap` recommended
+           :class: note
+
+           :mc:`mc idp ldap` offers additional features and improved validation over :mc-cmd:`mc admin config set` runtime configuration settings.
+           :mc:`mc idp ldap` supports the same settings as :mc:`mc admin config` and the :mc-conf:`identity_ldap` configuration key.
```
feorlen, line 169: This is an attempt to not have a second page that very nearly duplicates the existing `mc admin config identity_ldap` items. Not entirely convinced this is wise.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +195,60 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
 
-         The following example code sets *all* configuration settings related to
-         configuring an AD/LDAP provider for external
-         identity management. The minimum *required* setting are:
-         
-         - :mc-conf:`identity_ldap server_addr <identity_ldap.server_addr>`
+      .. tab-item:: MinIO Console
 
-         - :mc-conf:`identity_ldap lookup_bind_dn <identity_ldap.lookup_bind_dn>`
+         MinIO supports specifying the AD/LDAP provider settings using the :ref:`MinIO Console <minio-console>`.
```
feorlen, line 209: Does configuring LDAP from Console do it for the entire deployment? (Like the `mc` methods do.)

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: Also: better way to format the long line for `username_format`?

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
ravindk89: My understanding is we prefer users utilize `mc idp ldap` as it is just less error prone and a bit easier to manage. It effectively sets the config settings under the hood.

I don't think we would ever deprecate the direct config route - but what we probably _should_ do is add a new tab to the LDAP (and OpenID ) pages that uses the MC CLI, then envvars, then config - where we state in the latter tabs that we recommend `mc idp <thing>`.

source/administration/console/security-and-access.rst
```diff
@@ -154,4 +156,4 @@ Configuring an external IDentity Provider (IDP) enables Single-Sign On (SSO) wor
 Use the the screens in this section to view, add, or edit an LDAP configuration for the deployment.
 MinIO only supports one active LDAP configuration.
 
-MinIO queries the active Active Directory / LDAP server to verify the credentials specified by the application and optionally return a list of groups in which the user has membership. 
\ No newline at end of file
+MinIO queries the active Active Directory / LDAP server to verify the credentials specified by the application and optionally return a list of groups in which the user has membership.
```
ravindk89, line 159: Hm. Rather than "Optionally", it's "if configured". But that is a hairy sentence.

```suggestion
MinIO queries the Active Directory / LDAP server to verify the client-specified credentials. 
MinIO also performs a group lookup on the AD/LDAP server if configured to do so.
```

Either way wasn't sure if "the active Active Directory.." was intentional. Probably a whoopsie from the before times.

source/includes/common-minio-external-auth.rst
```diff
@@ -317,6 +317,23 @@ Specify a comment to associate to the AD/LDAP configuration.
 
 .. end-minio-ad-ldap-comment
 
+.. start-minio-ad-ldap-console-enable
+
+#. Log in to the MinIO Console as an admin user, either your root user or one with the ``consoleAdmin`` :ref:`policy <minio-policy>`.
```
ravindk89, line 322: Slight simplification

```suggestion
#. Log in to the MinIO Console as either the :ref:`root <minio-users-root>` user or a MinIO user with the  :userpolicy:`consoleAdmin` policy.
```

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
ravindk89: I think those are fine for sample values - we cannot reasonably dig int AD/LDAP syntax here.

As far as formatting...not really :grimacing: 

They'll have to scroll, and that is the life that AD/LDAP admins are committed to.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,83 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+         For distributed deployments, the :mc:`mc idp ldap` command applies the configuration to all nodes in the deployment. 
+
+         The following example code sets *all* configuration settings related to configuring an AD/LDAP provider for external identity management.
+	 The minimum *required* settings are:
+
+         - :mc-conf:`server_addr <identity_ldap.server_addr>`
+         - :mc-conf:`lookup_bind_dn <identity_ldap.lookup_bind_dn>`
+         - :mc-conf:`lookup_bind_password <identity_ldap.lookup_bind_password>`
+         - :mc-conf:`user_dn_search_base_dn <identity_ldap.user_dn_search_base_dn>`
+         - :mc-conf:`user_dn_search_filter <identity_ldap.user_dn_search_filter>`
```
ravindk89, line 142: This is more of a note:

So this will point the user ad the config settings instead of `mc idp ldap add <param>`. But we don't yet have dedicated pages for each param, so we don't have a choice here _yet_.

I think if we want to go properly into showcasing this command we're going to have to commit to breaking it all out into standalone pages.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
ravindk89: I think this is fine for now. Maybe as a second pass we will figure out how to cleanly remove the reliance on referencing config settings at all.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,83 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
-         startup. For distributed deployments, specify these settings across all
-         nodes in the deployment using the *same* values consistently.
-
-         The following example code sets *all* environment variables related to
-         configuring an AD/LDAP provider for external
-         identity management. The minimum *required* variable are:
-         
+         MinIO supports specifying the AD/LDAP provider settings using :ref:`environment variables <minio-server-envvar-external-identity-management-ad-ldap>`.
+	 The :mc:`minio server` process applies the specified settings on its next startup.
+	 For distributed deployments, specify these settings across all nodes in the deployment using the *same* values consistently.
```
ravindk89, line 177:
```suggestion
	 For distributed deployments, specify these settings across all nodes in the deployment using the *same* values.
	 Any differences between server configs will result in startup or configuration failures.
```

Just to make the consequence clear.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
ravindk89: I believe so - it should be the rough equivalent of using `mc` from an API perspective.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +195,60 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
    .. code-block:: shell
       :class: copyable
 
       mc admin service restart ALIAS
 
-   Replace ``ALIAS`` with the :ref:`alias <alias>` of the deployment to 
-   restart.
+   Replace ``ALIAS`` with the :ref:`alias <alias>` of the deployment to restart.
+
+   If you configured AD/LDAP from the MinIO Console, no additional action is required.
```
ravindk89, line 228: Maybe move this up to just after the first line in case someone reads and executes in strict ordering here

ravindk89 (changes requested):
Minimal notes and the open question as to whether we want to commit to carving out `mc admin config set identity_ldap` now - which will require us to add child pages to `mc idp ldap`.

source/administration/console/security-and-access.rst
feorlen: Keeping the anchor, but leaving this page as it is for now.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: new issue for this https://github.com/minio/docs/issues/940

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
djwfyi: I think the long line is fine as is. There will be a horizontal scroll bar in the rendered site for people who want to see the whole thing.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -29,7 +30,9 @@ The procedure on this page provides instructions for:
    - Accessing the MinIO Console using AD/LDAP credentials.
    - Using the MinIO ``AssumeRoleWithLDAPIdentity`` Security Token Service (STS) API to generate temporary credentials for use by applications.
 
-This procedure is generic for AD/LDAP services. Defer to the documentation for the AD/LDAP provider of your choice for specific instructions or procedures on configuration of user identities.
+This procedure is generic for AD/LDAP services.
+Defer to the documentation for the AD/LDAP provider of your choice for specific instructions or procedures on configuration of user identities.
```
djwfyi, line 34:
```suggestion
See the documentation for the AD/LDAP provider of your choice for specific instructions or procedures on configuration of user identities.
```

`Defer` strikes me as too ... posh?

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -99,40 +101,84 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
-         nodes in the deployment using the *same* values consistently.
-
-         The following example code sets *all* environment variables related to
-         configuring an AD/LDAP provider for external
-         identity management. The minimum *required* variable are:
-         
+         MinIO supports specifying the AD/LDAP provider settings using :ref:`environment variables <minio-server-envvar-external-identity-management-ad-ldap>`.
+	 The :mc:`minio server` process applies the specified settings on its next startup.
+	 For distributed deployments, specify these settings across all nodes in the deployment using the *same* values.
+	 Any differences between server configs will result in startup or configuration failures.
```
djwfyi, line 178:
```suggestion
	 Any differences between server configs between nodes will result in startup or configuration failures.
```

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +196,62 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
 
    2) Restart the MinIO Deployment
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   You must restart the MinIO deployment to apply the configuration changes. 
-   Use the :mc-cmd:`mc admin service restart` command to restart the deployment.
+   You must restart the MinIO deployment to apply the configuration changes.
+
+   If you configured AD/LDAP from the MinIO Console, no additional action is required.
+   The MinIO Console automatically restarts the deployment when the new AD/LDAP configuration is saved.
```
djwfyi, line 223:
```suggestion
   The MinIO Console automatically restarts the deployment after saving the new AD/LDAP configuration.
```

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +196,62 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
-   (STS) endpoint, and logging the user into the MinIO deployment.
-
-   Starting in :minio-release:`RELEASE.2021-07-08T01-15-01Z`, the MinIO Console is
-   embedded in the MinIO server. You can access the Console by opening the root URL
-   for the MinIO cluster. For example, ``https://minio.example.net:9000``.
+   The MinIO Console supports the full workflow of authenticating to the AD/LDAP provider, generating temporary credentials using the MinIO :ref:`minio-sts-assumerolewithldapidentity` Security Token Service (STS) endpoint, and logging the user into the MinIO deployment.
 
-   From the Console, click :guilabel:`BUTTON` to begin the Active Directory / LDAP
-   authentication flow.
+   Starting in :minio-release:`RELEASE.2021-07-08T01-15-01Z`, the MinIO Console is embedded in the MinIO server.
```
djwfyi, line 239: That's two years ago now. I think we can drop this line.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +196,62 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+   Once logged in, you can perform any action for which the authenticated user is :ref:`authorized <minio-external-identity-management-ad-ldap-access-control>`.
 
-   You can also create :ref:`access keys <minio-idp-service-account>` for
-   supporting applications which must perform operations on MinIO. Access Keys
-   are long-lived credentials which inherit their privileges from the parent user.
-   The parent user can further restrict those privileges while creating the service
-   account. 
+   You can also create :ref:`access keys <minio-idp-service-account>` for supporting applications which must perform operations on MinIO.
+   Access Keys are long-lived credentials which inherit their privileges from the parent user.
+   The parent user can further restrict those privileges while creating the service account.
```
djwfyi, line 246: I think we need to be clear that `service account` is the same as `access keys`.
Otherwise, this is a jarring switch when the rest of the paragraph is talking about access keys.

But I think we call them `service accounts` in most places in MinIO, so maybe it's worth intentionally linking the two.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +196,62 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
 
    4) Generate S3-Compatible Temporary Credentials using AD/LDAP Credentials
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   MinIO requires clients authenticate using :s3-api:`AWS Signature Version 4
-   protocol <sig-v4-authenticating-requests.html>` with support for the deprecated
-   Signature Version 2 protocol. Specifically, clients must present a valid access
-   key and secret key to access any S3 or MinIO administrative API, such as
-   ``PUT``, ``GET``, and ``DELETE`` operations.
+   MinIO requires clients authenticate using :s3-api:`AWS Signature Version 4 protocol <sig-v4-authenticating-requests.html>` with support for the deprecated Signature Version 2 protocol.
```
djwfyi, line 251:
```suggestion
   MinIO requires that clients authenticate using :s3-api:`AWS Signature Version 4 protocol <sig-v4-authenticating-requests.html>` with support for the deprecated Signature Version 2 protocol.
```

or

```suggestion
   MinIO requires clients to authenticate using :s3-api:`AWS Signature Version 4 protocol <sig-v4-authenticating-requests.html>` with support for the deprecated Signature Version 2 protocol.
```

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -150,99 +196,62 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
...
+   Specifically, clients must present a valid access key and secret key to access any S3 or MinIO administrative API, such as ``PUT``, ``GET``, and ``DELETE`` operations.
 
-   Applications can generate temporary access credentials as-needed using the
-   :ref:`minio-sts-assumerolewithldapidentity` Security Token Service (STS) API
-   endpoint and AD/LDAP user credentials. MinIO provides an example Go application
-   :minio-git:`ldap.go <minio/blob/master/docs/sts/ldap.go>` with an example of
-   managing this workflow.
+   Applications can generate temporary access credentials as-needed using the :ref:`minio-sts-assumerolewithldapidentity` Security Token Service (STS) API endpoint and AD/LDAP user credentials. MinIO provides an example Go application :minio-git:`ldap.go <minio/blob/master/docs/sts/ldap.go>` with an example of managing this workflow.
```
djwfyi, line 254:
```suggestion
   Applications can generate temporary access credentials as-needed using the :ref:`minio-sts-assumerolewithldapidentity` Security Token Service (STS) API endpoint and AD/LDAP user credentials. 
   MinIO provides an example Go application :minio-git:`ldap.go <minio/blob/master/docs/sts/ldap.go>` that manages this workflow.
```

djwfyi (changes requested):
A few wording suggestions, and I'll take another look after.

source/administration/console/security-and-access.rst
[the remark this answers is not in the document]
feorlen: Hmm Ravind's suggested change _should_ have updated this. I'll see if something went weird with that.

source/administration/identity-access-management/ad-ldap-access-management.rst
```diff
@@ -57,7 +57,7 @@ Access Keys are long-lived credentials which inherit their privileges from the p
 The parent user can further restrict those privileges while creating the access keys. 
 Use either of the following methods to create a new access keys:
 
-- Log into the :ref:`MinIO Console <minio-console>` using the AD/LDAP-managed user credentials. From the :guilabel:`Identity` section of the left navigation, select :guilabel:`Access Keys` followed by the :guilabel:`Create access keys +` button.
+- Log into the :ref:`MinIO Console <minio-console>` using the AD/LDAP-managed user credentials. From the :guilabel:`User` section of the left navigation, select :guilabel:`Access Keys` followed by the :guilabel:`Create access keys +` button.
```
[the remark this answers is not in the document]
feorlen: It's always about _when_ the UI is going to change, isn't it. (Why I did not add screenshots.) I like your reword, thanks!

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
donatello: Yes, `mc admin config set ALIAS identity_ldap ...` is not deprecated - they are just lower level commands that were used previously - they will be available. The newer `mc idp ldap` is higher level and does more LDAP validation (for e.g. it should have better error messages when something is misconfigured or not working).

On minio/minio#17790 the problem is that the config parameter name is not right `identity_ldap_server_insecure="on"` - this has a unnecessary `identity_ldap_` prefix and it should throw an error, but there is an issue with the parser that may need some larger changes. I am marking it as low priority for now, but will get around to fixing it eventually.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: I had that thought as well, although it's used elsewhere too. A second 🤔 confirms it. ✂️

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: I like the second, thanks!

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: Aha! It was specifically about that parameter, not anything with `identity_ldap` more generally. That makes sense. 

The follow-on PR will separate the two methods even more and hopefully reduce confusion.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: oh that's a good clarification 👍🏻

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: The page this links to says `service account` is an outdated name for `access key`. Which suggests favoring `access key`, like what `mc admin user svcacct` does. (Except, of course, all the _other_ places that use `service account`.)

This is confusing _for me_, much less a novice docs reader.

Is there an established way to differentiate between "an access key I, a human, uses to authenticate" and "an access key for an automated service?"

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
```diff
@@ -175,7 +175,7 @@ An AD/LDAP user with no assigned policy *and* with membership in groups with no
          MinIO supports specifying the AD/LDAP provider settings using :ref:`environment variables <minio-server-envvar-external-identity-management-ad-ldap>`.
 	 The :mc:`minio server` process applies the specified settings on its next startup.
 	 For distributed deployments, specify these settings across all nodes in the deployment using the *same* values.
-	 Any differences between server configs will result in startup or configuration failures.
+	 Any differences between server configs between nodes will result in startup or configuration failures.
```
djwfyi, line 178: Ooh, those two `between`s are very close together.

```suggestion
	 Any differences in server configurations between nodes will result in startup or configuration failures.
```

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
djwfyi: With all of the other `mc` command reworking that's happened, maybe move this functionality to `mc admin user accesskey`. Or even just pull it out to `mc admin accesskey`. But that's not something done in the docs.

As long as `svcacct` is the command, I don't think we can avoid the confusion.

djwfyi (approved):
One of my suggestions created a clash of `between`s. 
Otherwise, this looks good to me until we resolve the issue with service accounts/access keys.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
feorlen: ooo I didn't notice that either. Good call.

source/includes/common-minio-external-auth.rst
```diff
@@ -317,6 +317,23 @@ Specify a comment to associate to the AD/LDAP configuration.
...
+   The minimum required settings are:
+
+   - Server Address
+   - Lookup Bind DN
+   - Lookup Bind Password
+   - User DN Search Base
+   - User DN Search Filter
+
+   Not all configuration options are available in the MinIO Console.
+   For additional settings, use the MinIO Client or environment variables.
```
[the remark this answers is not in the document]
feorlen: Couldn't figure out the ideal links (to the other tabs) but I got something.

source/operations/external-iam/configure-ad-ldap-external-identity-management.rst
ravindk89: I think I raised this to @kannappanr a while back, but I'm not sure if/when we will have engineering resources to add it as an alias (effectively).

ravindk89 (approved):
I don't have anything else to add - lets get it up, and we can deal with any trailing bits from there.

PR #941, opened 2023-08-09
Update configure-minio-kes-hashicorp.rst

bhara03021989:
Adding /v1/key/list/* in the doc instead of  /v1/key/list which gives Authentication issue .

policy:
minio:
allow:
- /v1/key/create/* # You can replace these wildcard '' with a string prefix to restrict key names
- /v1/key/generate/ # e.g. '/minio-'
- /v1/key/decrypt/*
- /v1/key/bulk/decrypt
- /v1/key/list
- /v1/status

### DIFF
--- a/source/operations/server-side-encryption/configure-minio-kes-hashicorp.rst
+++ b/source/operations/server-side-encryption/configure-minio-kes-hashicorp.rst
@@ -309,7 +309,7 @@ The following section describes each of the |KES-git| configuration settings for
              - /v1/key/generate/*
              - /v1/key/decrypt/*
              - /v1/key/bulk/decrypt
-             - /v1/key/list
+             - /v1/key/list/*
              - /v1/status
              - /v1/metrics
              - /v1/log/audit


ravindk89 (commented):
So this fixes the issue in only one area - the permission here is referenced in half a dozen other files as well.

I'll take this change just to patch the one file, but we also need to fix the stepfiles on top of that. I am already working on those changes separately plus validating whether we *need* listing for **MinIO**. That is a distinct thing from what someone administrating KES might need to list and manage keys (cc @aead )

bhara03021989:
@ravindk89  If the "list" is not provided we would not be able to select the created keys from the web console against the bucket. So , I believe listing is necessary. Also , if not listed on the console the error comes as **"not authorized: insufficient permissions"** not authentication sorry my bad in the comment !

ravindk89:
No you're fine - it makes sense if it's a Console-related change vs a MinIO-server related behavior. 

Either way will work this across the remaining pages, which in and of themselves require some restructuring.

PR #946, opened 2023-08-10
Adding C++ SDK into the list

cniackz:
### Objective:

To add C++ SDK in the page: https://min.io/docs/minio/linux/developers/minio-drivers.html

Screenshot 2023-08-10 at 12 39 28 PM

### DIFF
--- a/source/developers/minio-drivers.rst
+++ b/source/developers/minio-drivers.rst
@@ -25,8 +25,6 @@ MinIO publishes the following Software Development Kits (SDK):
 C++ (``minio-cpp``)
 -------------------
 
-Latest Version
-  |cpp-sdk-version|
 
 Reference
   `MinIO C++ SDK Reference <https://minio-cpp.min.io/>`__
--- a/source/developers/minio-drivers.rst
+++ b/source/developers/minio-drivers.rst
@@ -18,43 +18,40 @@ MinIO publishes the following Software Development Kits (SDK):
 4. :ref:`Java <java-sdk>`
 5. :ref:`JavaScript <javascript-sdk>`
 6. :ref:`Python <python-sdk>`
+7. :ref: `C++ <cpp-sdk>`
 
+.. _cpp-sdk:
 
-..
-  C++ repo does not have any releases yet. Once released, unblock this section and add to toctree and numbered list.
-  
-  .. _cpp-sdk:
+C++ (``minio-cpp``)
+-------------------
 
-  C++ (``minio-cpp``)
-  -------------------
+Latest Version
+  |cpp-sdk-version|
 
-  Latest Version
-    |cpp-sdk-version|
+Reference
+  `MinIO C++ SDK Reference <https://minio-cpp.min.io/>`__
 
-  Reference
-    :doc:`MinIO C++ SDK Reference </developers/cpp/minio-cpp>`
+Install
+  - ``vcpkg``
 
-  Install
-    - ``vcpkg``
+    .. code-block:: shell
+       :class: copyable
+
+       vcpkg install minio-cpp
+
+  - Source
              
-      .. code-block:: shell
-         :class: copyable
+    .. code-block:: shell
+       :class: copyable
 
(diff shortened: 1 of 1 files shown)


source/developers/minio-drivers.rst
```diff
@@ -18,43 +18,40 @@ MinIO publishes the following Software Development Kits (SDK):
...
 
-  C++ (``minio-cpp``)
-  -------------------
+Latest Version
+  |cpp-sdk-version|
 
-  Latest Version
-    |cpp-sdk-version|
+Reference
+  :doc:`MinIO C++ SDK Reference <https://minio-cpp.min.io/>`
```
ravindk89, line 32:
```suggestion
  `MinIO C++ SDK Reference <https://minio-cpp.min.io/>`__
```

This will create a hyperlink to the doxygen reference

sync-docs.sh
```diff
@@ -17,7 +17,7 @@ function replace() {
 
 function main() {
     if [ "$#" -eq 0 ]; then
-	SDKS="dotnet go java js py hs"
+	SDKS="dotnet go java js py hs cpp"
```
ravindk89, line 20: So - the C++ docs do *not* have an `API.md` in the docs directory as it's a doxygen.

If you look at lines 42 and 43 we curl for both `API.md` and `README.md`

so this will introduce a guaranteed 404 and delay in builds for the C++ guide.

We also are not (at the point of this PR) creating a standalone C++ page (we can do that on a second pass along with more `case esac` statements in bash.

We can either:

1. Revert this change and come back to it later , or
2. Create `/developers/cpp/minio-cpp.rst`
3. Include `/developers/cpp/quickstart.md`
4. Not have a TOC for that page

You can look at `minio-java.rst` for an idea of how that would owrk.

ravindk89 (changes requested):

sync-docs.sh
cniackz: I see your point and you are right, I will revert this change for later.

source/developers/minio-drivers.rst
```diff
@@ -18,43 +18,40 @@ MinIO publishes the following Software Development Kits (SDK):
 4. :ref:`Java <java-sdk>`
 5. :ref:`JavaScript <javascript-sdk>`
 6. :ref:`Python <python-sdk>`
+7. :ref: `C++ <cpp-sdk>`
```
ravindk89, line 21:
```suggestion
7. :ref:`C++ <cpp-sdk>`
```

source/developers/minio-drivers.rst
```diff
@@ -18,43 +18,40 @@ MinIO publishes the following Software Development Kits (SDK):
...
-  C++ repo does not have any releases yet. Once released, unblock this section and add to toctree and numbered list.
-  
-  .. _cpp-sdk:
+C++ (``minio-cpp``)
+-------------------
 
-  C++ (``minio-cpp``)
-  -------------------
+Latest Version
+  |cpp-sdk-version|
```
ravindk89, line 29: Hm. Dang. This requires the Makefile to sync the latest version.

Also https://github.com/minio/minio-cpp implies there has not been a relase in an entire year. I am not sure that is correct (@balamurugana ? )

Removing this for now until we have an answer.

ravindk89, line 29:
```suggestion
```

source/developers/minio-drivers.rst
```diff
@@ -18,43 +18,38 @@ MinIO publishes the following Software Development Kits (SDK):
...
-  
-  .. _cpp-sdk:
+C++ (``minio-cpp``)
+-------------------
 
-  C++ (``minio-cpp``)
-  -------------------
 
-  Latest Version
-    |cpp-sdk-version|
```
ravindk89, line 32: Whoops. Deleted my comment.

@balamurugana how are we versioning minio-cpp + there has not been a release in ~1 year.

ravindk89:
We can merge as-is and fixup from there.

PR #949, opened 2023-08-10
DOCS-912: Part 1: Cleaning up Vault

ravindk89:
Staged: http://192.241.195.202:9000/staging/DOCS-912/linux/operations/server-side-encryption/configure-minio-kes-hashicorp.html

As part of digging into this, I'm trying to unwind the significant amount of deduplication effort made initially.

As it turns out, that makes updating these pages a lot harder.

There's also the issue that the "Quickstarts" for local development aren't all that helpful in terms of being at all different from what the production side does. That is, you can derive what to do in a local environment from the production steps.

SO

This is part 1:

1) Fix the issue of incorrect permissions across the board
2) Validate the procedure and start removing cruft
3) Begin removing single-sourcing
4) Use API-key based authentication
5) Many other minor cleanups

If this feels like a good way to move forward, I will *then*

1) Remove all the "Quick" stuff
2) Begin removing single-sourced stuff that no longer needs to exist
3) Simplify as much as possible the TLS configurations for the Container, Windows, and MacOS setups
4) Remove the heavy dependence on substitutions to make this all work.

This is going to be *messy*. I have to build across multiple platforms at some point, but *this* PR is just going to deal with the most glaring issues and get us to a working tutorial.

What has not yet been done is validating AWS Secrets Manager against the user report of varying errors in Console.


source/includes/linux/common-minio-kes.rst
```diff
@@ -88,41 +88,26 @@ For existing MinIO deployments, run the following command on each MinIO host to
...
 
-         # If the Vault certs are self-signed or use a non-global CA
-         # Include those CA certs as well
+Depending on your Vault configuration, you may also need to create a dedicated set of TLS certificates for KES to connect and authenticate to Vault.
 
-         -rw-r--r-- 1 kes:kes |kescertpath|/vault-CA.cert
+Defer to your organizations best practices around generating production-ready TLS certificates.
 
-   .. tab-item:: MinIO Hosts
+Place the certificates and corresponding private keys a directory the KES service user can access and read their contents.
```
feorlen, line 98:
```suggestion
Place the certificates and corresponding private keys in a directory the KES service user can access and read their contents.
```

feorlen:
I guess some of these abbreviations should have been expanded? Most I know, but I was confused by EK nearly to the end

Text beginning "As part of this procedure, you will" and containing numerous unexpanded acronyms

source/includes/linux/common-minio-kes.rst
djwfyi:
```suggestion
Place the certificates and corresponding private keys in a directory that the KES service user has permissions to access and read the directory's contents.
```

djwfyi (approved):
One phrasing suggestion for clarity, otherwise LTGM.

ravindk89:
@feorlen hm. I think I'll have to address the CSS on the abbreviations as part of this. We rely on them heavily for these pages anyways.

Issue #900, opened 2023-06-28
[Release] Operator 5.0.6 changes


djwfyi:
Operator [v5.0.6](https://github.com/minio/operator/releases/tag/v5.0.6) has changes that impact the docs:

- [x] (Helm) Adds support for custom extra resources to add to the container ([PR #1643](https://github.com/minio/operator/pull/1643))
- [x] MinIO can now run in restricted namespaces ([PR #1651](https://github.com/minio/operator/pull/1651))
- [x] Feature: Adds support for [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) ([PR #1644](https://github.com/minio/operator/pull/1644))
- [x] (Helm) Adds example to the yaml for bucket schemas ([PR #1657](https://github.com/minio/operator/pull/1657))

djwfyi:
@feorlen checking on the Helm stuff, please.

feorlen:
https://github.com/minio/operator/pull/1657 doesn't appear to require any additional docs.

Docs don't cover the structure or contents of Helm charts other than to say you can modify them as needed. We link to Helm docs and presume anyone who does that is familiar with the format. And they will see this comment when they get there.

feorlen:
For https://github.com/minio/operator/pull/1643, following up with Engineering about adding an example for this in the YAML file. The docs don't currently cover how to modify the Helm chart, and this appears to be a fairly uncommon use case.

feorlen:
Note: current docs have a single mention of init containers https://min.io/docs/minio/kubernetes/upstream/operations/installation.html#minio-operator-components

pjuarezd:
> 

This would be an practical example
https://github.com/minio/operator/pull/1739

feorlen:
Helm `extraResources` (https://github.com/minio/operator/pull/1643) handled with a comment in the Helm chart.

See https://github.com/minio/operator/pull/1739

PR #967, opened 2023-08-23
Optional custom init container

feorlen:
Operator 5.0.6+ allows custom init containers, should your pod require additional initialization. Add a brief mention in the Operator install page and link to the k8s docs.

Staged
http://192.241.195.202:9000/staging/DOCS-900-2/k8s/operations/installation.html#minio-operator-components

Partially addresses https://github.com/minio/docs/issues/900

### DIFF
--- a/source/includes/k8s/deploy-operator.rst
+++ b/source/includes/k8s/deploy-operator.rst
@@ -50,6 +50,8 @@ Each pod runs three containers:
 - SideCar container that monitors configuration secrets for the tenant and updates them as they change.
   This container also monitors for root credentials and creates an error if it does not find root credentials. 
 
+Starting with v5.0.6, the MinIO Operator supports custom :kube-docs:`init containers <concepts/workloads/pods/init-containers>` for additional pod initialization that may be required for your environment.
+
 The tenant utilizes Persistent Volume Claims to talk to the Persistent Volumes that store the objects.
 
 .. image:: /images/k8s/OperatorsComponent-Diagram.png


source/includes/k8s/deploy-operator.rst
```diff
@@ -50,6 +50,8 @@ Each pod runs three containers:
 - SideCar container that monitors configuration secrets for the tenant and updates them as they change.
   This container also monitors for root credentials and creates an error if it does not find root credentials. 
 
+Starting with v5.0.6, the MinIO Operator supports custom :kube-docs:`init containers <concepts/workloads/pods/init-containers>` for additional pod initialization that may be required for your environment.
```
ravindk89, line 53: I'm OK with just saying this, *but* I think we will get a follow-on of "how do I do this".

I am *guessing* either via modifying the tenant YAML manually *or* through the Operator Console - *assuming* we already support that UI/UX. If not would it only be via YAML?

ravindk89 (approved):
LGTM mod one nit that we can push to a separate scope/

source/includes/k8s/deploy-operator.rst
feorlen: > I think we will get a follow-on of "how do I do this".

Someone will want more, for sure. From the conversation with Celis, it involves creating a new container... and after that is where I get lost. So far, the only reported use for a special init container is to fix file permissions in a particular type of setup. And this change allows someone to do that if they need to.

Celis also suggested documenting what our MinIO init container does, perhaps info about custom ones could be included if we did that.

I'll open a docs issue for this, although it is likely to sit in the backlog a while.

PR #972, opened 2023-08-25
Batch job Permissions

### LINKED ISSUE #954: Add section on permissions for Batch Replication
**Summary**

As per internal discussion, we should document the minimum set of policy actions required to run batch replication:

```yaml
{
    "Version" : "2012-10-17",
    "Statement" : [
        {
            "Effect" : "Allow",
            "Action": [
                "admin:StartBatchJob",
                "admin:ListBatchJobs",
                "admin:DescribeBatchJob",
                "admin:CancelBatchJob"
            ],
            "Resource" : [
                "arn:aws:s3:::BUCKET/*"
            ]
        }
    ]
}
```

We have a [generic section](https://min.io/docs/minio/linux/administration/batch-framework.html#access-to-mc-batch) that could use a these details explicitly. We can also use the canned "For example, the ``consoleAdmin`` built-in policy has sufficient permissions..."

Need to verify if setting the resource to a bucket limits the batch job execution to only that bucket and/or prefix.

This is *just* for starting the batch job. The credentials provided *to* the batch job will further allow or restrict actions - that is, if the batch YAML credentials specified for the source or remote targets do not provide sufficient permissions, the job would still fail. We could probably mention this under the access section

**Goals**

List the in-scope goals

- [ ] Document minimum admin permissions to start/stop/list batch jobs
- [ ] Make note that access for source/remote dictates success of batch job itself

**Non-Goals**

Deep dive into policy curation for batch jobs

**Additional context**

I'd suggest modifying the header from `Access to mc batch` to something like `Access Control for Batch Jobs` and cover the Batch itself + the access for source/remote.

djwfyi:
Adds information to the batch job doc about available permissions.
Updates the batch job policy list to include `cancel`.

Closes #954 

Staged at http://192.241.195.202:9000/staging/permissions/linux/administration/batch-framework.html#access-to-mc-batch

### DIFF
--- a/source/administration/batch-framework.rst
+++ b/source/administration/batch-framework.rst
@@ -79,7 +79,27 @@ The :mc:`mc batch` commands include
 Access to ``mc batch``
 ----------------------
 
-You can use MinIO's :ref:`Policy Based Access Control <minio-policy>` and the :ref:`administrative policy actions <minio-policy-mc-admin-actions>` to restrict who can start a batch job, retrieve a list of running jobs, or describe a running job.
+The batch function itself is a server process not limited by access keys or policies.
+Instead, use MinIO's :ref:`Policy Based Access Control <minio-policy>` and the :ref:`administrative policy actions <minio-policy-mc-admin-actions>` to restrict who can start a batch job, retrieve a list of running jobs, or describe a running job.
+
+MinIO provides the following admin policy actions to control access to Batch Replication functions:
+
+``admin:ListBatchJobs``
+  Grants the user the ability to see batch jobs currently in process.
+
+``admin:DescribeBatchJobs``
+  Grants the user the ability to see the definition details of batch job currently in process.
+
+``admin:StartBatchJob``
+  Grants the user the ability to start a batch job.
+  The job may be further restricted by the credentials the job uses to access either the source or target deployments.
+
+``admin:CancelBatchJob``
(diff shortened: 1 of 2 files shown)


source/administration/batch-framework.rst
```diff
@@ -79,16 +79,42 @@ The :mc:`mc batch` commands include
 Access to ``mc batch``
 ----------------------
 
-You can use MinIO's :ref:`Policy Based Access Control <minio-policy>` and the :ref:`administrative policy actions <minio-policy-mc-admin-actions>` to restrict who can start a batch job, retrieve a list of running jobs, or describe a running job.
+The batch function itself is a server process not limited by access keys or policies.
```
feorlen, line 82: I think this means if you have shell access we can't stop you?

Perhaps something along these lines?
```suggestion
The batch function itself is a server process, with access control managed by OS permissions rather than MinIO access keys or policies.
```

feorlen (approved):
Left a suggestion for your consideration, there was a bit I didn't entirely understand. The rest is 👍🏻  I'll approve and let you sort that out as you prefer.

source/administration/batch-framework.rst
djwfyi: Tweaked the wording and expanded the section a little bit to hopefully better clarify my intent with this section.

feorlen (approved):
LGTM

PR #973, opened 2023-08-28
Documents the scanner.speed config and minio_scanner_speed envvar

### LINKED ISSUE #796: docs: document `MINIO_SCANNER_SPEED` envvar
**Is your feature request related to a problem? Please describe.**
I'm facing performance issues with a versioned bucket (keep 3 latest versions) storing lots of small files (partitioned parquet files).

I have a batch process that deletes parts (using a prefix) of the bucket before writing new files there (new files aren't named the same).
I hit timeouts on the delete operation 9 times out of 10, it only works when the healing process has done a cleanup just before the process starts.

My guess is that the way things work, leads to a proliferation a small outdated files that are kept on disk.
When trying to delete parts of the bucket, it takes too much time to go through all these small files leading to the timeout.

I think I can mitigate this issue by tuning the object scanner settings and the healing settings.
However, as far as I can tell,  these settings aren't documented on https://min.io/docs/minio/linux/ (using the search bar).

- https://min.io/docs/minio/linux/administration/object-management/object-lifecycle-management.html#lifecycle-management-object-scanner mentions the object scanner, but not the related settings

**Describe the solution you'd like**
It would be nice if some documentation/pointer was added for:
- `mc admin config set alias/ scanner`
- `mc admin config set alias/ heal`

**Describe alternatives you've considered**
I'm reading the source code for now, as well as related issues.

djwfyi:
Closes #796

Staged:

- http://192.241.195.202:9000/staging/speed/linux/reference/minio-mc-admin/mc-admin-config.html#scanner
- http://192.241.195.202:9000/staging/speed/linux/reference/minio-server/minio-server.html#envvar.MINIO_SCANNER_SPEED

And a few other paragraphs here there and yon.

### DIFF
--- a/source/includes/common/scanner.rst
+++ b/source/includes/common/scanner.rst
@@ -1,0 +1,35 @@
+.. start-scanner-speed-config
+
+You can adjust how MinIO balances the scanner performance with read/write operations using either the :envvar:`MINIO_SCANNER_SPEED` environment variable or the :mc-conf:`scanner speed <scanner.speed>` configuration setting.
+
+.. end-scanner-speed-config
+
+
+.. start-scanner-speed-values
+
+MinIO utilizes the scanner for :ref:`bucket replication <minio-bucket-replication>`, :ref:`site replication <minio-site-replication-overview>`, and :ref:`lifecycle management <minio-lifecycle-management>` tasks.
+
+Valid values include:
+
+.. list-table::
+   :stub-columns: 1
+   :widths: 30 70
+   :width: 100%
+   
+   * - ``fastest``
+     - Removes wait time for scanner activity, maximizing scanner performance at the possible cost of some read or write performance.
+   
+   * - ``fast``
+     - Adds a small wait time for scanner activity if MinIO notices a latency in read or write processes slight risk of some loss of read or write performance.
+   
+   * - ``default``
+     - Adds a moderate wait time for scanner activity if MinIO notices a latency in any read or write processes.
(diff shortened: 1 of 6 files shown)


ravindk89 (changes requested):
Mostly ideas to tighten/make consistent the language and try to create a 'scale' language to describe how these work. Thoughts:

- Either use 'wait time' or 'delay', not both.
- Try to use similar scalars since we are being generic
- I thought the `"add"` language might imply additive rather than a hard-set value.

I experimented with None -> short - moderate -> medium -> large wait times. 

Could also use your original language and do something like

None -> small -> moderate -> medium -> long delay

source/includes/common/scanner.rst
```diff
@@ -0,0 +1,43 @@
...
+Valid values include:
+
+.. list-table::
+   :stub-columns: 1
+   :widths: 30 70
+   :width: 100%
+   
+   * - ``fastest``
+     - Removes scanner wait on read/write latency, allowing the scanner to operate at full-speed and IOPS consumption.
+       This setting may result in reduced read/write performance for IOPS-limited systems.
```
ravindk89, line 21: Nit - thinking about it we can just mirror `fast` here and call it done w.r.t. the IOPS bit.

ravindk89, line 21:
```suggestion
       This setting may result in reduced read and write performance.
```

ravindk89 (approved):
LGTM mod nit

feorlen (approved):

PR #975, opened 2023-08-29
Add additional context for what local means in the batch context

### LINKED ISSUE #924: Improvement for Batch Replication
https://github.com/minio/minio/discussions/17679

We could be slightly more clear in what we mean when we talk about the `local` deployment. More specifically, omit `endpoint`  and `credentials` on the local, whether source or remote.

djwfyi:
Closes #924

Staged:
- http://192.241.195.202:9000/staging/batch/linux/administration/batch-framework.html#local-deployment

### DIFF
--- a/source/administration/batch-framework.rst
+++ b/source/administration/batch-framework.rst
@@ -127,7 +127,7 @@ Either the ``source`` or the ``target`` **must** be the :ref:`local <minio-batch
 Starting with the MinIO Server ``RELEASE.2023-05-04T21-44-30Z``, the other deployment can be either another MinIO deployment or any S3-compatible location.
 
 The batch job definition file can limit the replication by bucket, prefix, and/or filters to only replicate certain objects.
-The access to objects and buckets for the replicate process may be restricted by the credentials you provide in the YAML for either the source or target destinations. 
+The access to objects and buckets for the replication process may be restricted by the credentials you provide in the YAML for either the source or target destinations. 
 
 .. versionchanged:: MinIO Server RELEASE.2023-04-07T05-28-58Z
 
--- a/source/administration/batch-framework.rst
+++ b/source/administration/batch-framework.rst
@@ -111,19 +111,23 @@ Job Types
 
    Depending on the job type, the success or failure of any batch job may be impacted by the credentials given in the batch job's YAML for the source or target deployments.
 
+.. _minio-batch-local:
+
+``Local`` Deployment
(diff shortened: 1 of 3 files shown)


source/reference/minio-mc/mc-batch-generate.rst
```diff
@@ -180,7 +180,8 @@ For the **source deployment**
          | Defaults to ``auto``.
```
ravindk89, line 181: For line 170 - either the `source` or `remote` must have `endpoint: "local"` right? if so i think the comment needs to state *that*. Spitballing:

```
If the source is the :ref:`alias` specified to the command, specify "local" to direct MinIO to use that alias for the endpoint and credentials values. You must omit the `credentials` key for the "local" deployment.
```

ravindk89: Maybe also include (for endpoint) to include the URL scheme (as an example you can do something like `Location of the source deployment, such as ``https://minio.example.net``.`

djwfyi: Dropping `source`. That is a leftover for when the source deployment had to be local.
And making some additional tweaks to the page related to that.
Otherwise mostly taking this suggestion.

source/administration/batch-framework.rst
```diff
@@ -111,19 +111,23 @@ Job Types
...
 
-   ``replicate`` batch jobs also support :mc:`mc mirror`-like behavior when presented an S3-compatible source or target.
+Use the ``replicate`` job type to create a batch job that replicates objects from one MinIO deployment (the ``source`` deployment) to another MinIO deployment (the ``target`` deployment).
+Either the ``source`` or the ``target`` **must** be the :ref:`local <minio-batch-local>` deployment.
+Starting with the MinIO Server ``RELEASE.2023-05-04T21-44-30Z``, the other deployment can be either another MinIO deployment or any S3-compatible location.
 
-At least one of the deployment locations, either the source or the target, must be ``local``.
-The definition file can limit the replication by bucket, prefix, and/or filters to only replicate certain objects.
+The batch job definition file can limit the replication by bucket, prefix, and/or filters to only replicate certain objects.
+The access to objects and buckets for the replicate process may be restricted by the credentials you provide in the YAML for either the source or target destinations.
```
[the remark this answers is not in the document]
djwfyi: Ah, good catch. Fixed.

PR #990, opened 2023-09-06
Reminder to enable minio.service

vdm:


source/includes/linux/common-installation.rst
```diff
@@ -312,6 +312,13 @@ transient and should resolve as the deployment comes online.
       
    If any drives remain offline after starting MinIO, check and cure any issues blocking their functionality before starting production workloads.
 
+By default the systemd unit is not enabled and so will not be started after a host reboot. To change this:
```
ravindk89, line 315:
```suggestion
By default the systemd unit is not enabled and so will not be started after a host reboot. To change this:

To ensure the MinIO service starts after a host reboot, you must explicitly enable the ``minio.service`` using the ``systemctl enable`` command:


```

ravindk89: Fixups to structure/style. otherwise good addition, thanks!

ravindk89 (approved):

PR #1008, opened 2023-09-20
DOCS-998: Tiering requires source and remote

### LINKED ISSUE #998: Specified remote tier is already in use - issue after PVC is removed.
Issue on this scenario: 
1. Minio is working on EKS (PVC storage), remote tier is configured with S3 bucket, ppensearch is configured and fully working
2. I'm recreating PVC - storage where OpenCTI bucket is located.
3. Executing the command to check if the remote tier is still exist: mc ilm tier myminio -> the result is that there is no remote tier what is expected since the PVC was recreated.
4. Trying to add the remote tier the same as it was before recreating PVC
5. Getting error after executing this command: mc ilm tier add s3 myminio S3 .... : **mc: <ERROR> Unable to configure remote tier target. Specified remote tier is already in use.**

I'm not able to add remote tier since mc is showing me that remote tier is not existing but when I'm trying to create the remote tier it is getting me the error that it is already in use. It looks like there are some leftover of the configuration and mc client is not able to manage with it. Maybe something need to be removed from OpenSearch? 

## Expected Behavior
I would like to add the same Remote Tier for ILM after PVC is recreated. 

## Current Behavior
I'm not able to add remote tier since mc is showing me that remote tier is not existing but when I'm trying to create the remote tier it is getting me the error that it is already in use. It looks like there are some leftover of the configuration and mc client is not able to manage with it. Maybe something need to be removed from OpenSearch? 

## Your Environment
* Version used (`minio --version`): RELEASE.2023-05-18T00-05-36Z
* Server setup and configuration: EKS AWS using PVC for persistent storage

ravindk89:
Closes #998 

Staged: 

- http://192.241.195.202:9000/staging/DOCS-998/linux/administration/object-management/object-lifecycle-management.html#object-transition-tiering
- http://localhost:8000/DOCS-998/linux/html/administration/object-management/object-lifecycle-management.html#exclusive-access-to-remote-data

# Summary

We sometimes have users who conceptualize tiering as a backup/recovery suite, or something where remote data can be used to restore data loss on the source.

As per engineering we are improving the guidance here to make it explicit that loss of data on either the source or remote is essentially data loss for those objects.

### DIFF
--- a/source/administration/object-management/object-lifecycle-management.rst
+++ b/source/administration/object-management/object-lifecycle-management.rst
@@ -58,6 +58,13 @@ The output includes the storage tier of each object:
 - ``STANDARD`` marks objects stored on the MinIO deployment.
 - ``WARM`` marks objects stored on the remote tier with matching name.
 
+.. important::
+
+   MinIO Object Transition does **not** provide backup and recovery functionality.
+   You cannot use the remote tier as a recovery source in the event of data loss in MinIO.
+
+   Use either :ref:`site replication <minio-site-replication-overview>` or :ref:`bucket replication <minio-bucket-replication>` to support backup/recovery or :abbr:`BC/DR (Business Continuity / Disaster Recovery)` requirements.
+
 Exclusive Access to Remote Data
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/includes/common-minio-tiering.rst
+++ b/source/includes/common-minio-tiering.rst
@@ -87,24 +87,25 @@ complete documentation on MinIO users and policies respectively.
 
 .. start-transition-bucket-access-desc
 
-MinIO *requires* exclusive access to the transitioned data on the remote storage
-tier. MinIO ignores any objects in the remote bucket or bucket prefix not
-explicitly managed by the MinIO deployment. Automatic transition and transparent
-object retrieval depend on the following assumptions:
+MinIO *requires* exclusive access to the transitioned data on the remote storage tier.
+Object metadata on the "hot" MinIO source is strongly linked to the object data on the "warm/cold" remote tier.
+MinIO cannot retrieve object data without access to the remote, nor can the remote be used to restore lost metadata on the source.
+
+All access to the transitioned objects *must* occur through MinIO via S3 API operations only.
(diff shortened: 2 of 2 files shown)


source/administration/object-management/object-lifecycle-management.rst
```diff
@@ -58,6 +58,13 @@ The output includes the storage tier of each object:
 - ``STANDARD`` marks objects stored on the MinIO deployment.
 - ``WARM`` marks objects stored on the remote tier with matching name.
 
+.. important::
+
+   MinIO Object Transition does **not** provide backup and recovery functionality.
```
djwfyi, line 63: I think we need to add something about when **to** use object transition.
Maybe something like:

```suggestion
   MinIO Object Transition is intended to move older and/or less frequently used objects from faster and typically more expensive storage devices to slower, typically less expensive options.
   Object Transition does **not** provide backup and recovery functionality.
```

Maybe something like the first line further up in the page rather than as part of this admonition. But this seems very harsh with no clear pointer as to when it is a good option.

djwfyi (changes requested):
Most looks good to me, but I think we should add a statement somewhere on the page about when object transition is a good option to use.

See my suggestion.

source/administration/object-management/object-lifecycle-management.rst
ravindk89: Hm. I could repeat what is on line 41 down here just as an emphasis...

```suggestion
   MinIO Object Transition supports cost-saving strategies around tiering older objects from the faster and typically more expensive "hot" tier to slower, less expensive "warm/cold" tiers.
   Object Transition does **not** provide backup and recovery functionality such as restoring data lost on the hot tier using the warm or cold tier.
```

djwfyi: You need '...**to** less expensive "warm/cold" tiers.'
But I think this works.

djwfyi (approved):
One tweak, but I like your rewording of my suggestion.

source/administration/object-management/object-lifecycle-management.rst
djwfyi: Github didn't put my comment where I expected. Anyway, I like your tweak, but you're missing a `to` after the last comma in line 65.

PR #1010, opened 2023-09-21
Batch replication requires realtime tier

djwfyi:
Based on discussion of https://github.com/minio/minio/pull/18044, adding info to the batch replication doc about requiring a realtime tier for replication.

Not staged.

### DIFF
--- a/source/administration/batch-framework.rst
+++ b/source/administration/batch-framework.rst
@@ -124,7 +124,9 @@ Replicate
 
 Use the ``replicate`` job type to create a batch job that replicates objects from one MinIO deployment (the ``source`` deployment) to another MinIO deployment (the ``target`` deployment).
 Either the ``source`` or the ``target`` **must** be the :ref:`local <minio-batch-local>` deployment.
-Starting with the MinIO Server ``RELEASE.2023-05-04T21-44-30Z``, the other deployment can be either another MinIO deployment or any S3-compatible location.
+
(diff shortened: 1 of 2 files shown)


source/administration/batch-framework.rst
```diff
@@ -124,7 +124,9 @@ Replicate
 
 Use the ``replicate`` job type to create a batch job that replicates objects from one MinIO deployment (the ``source`` deployment) to another MinIO deployment (the ``target`` deployment).
 Either the ``source`` or the ``target`` **must** be the :ref:`local <minio-batch-local>` deployment.
-Starting with the MinIO Server ``RELEASE.2023-05-04T21-44-30Z``, the other deployment can be either another MinIO deployment or any S3-compatible location.
+
+Starting with the MinIO Server ``RELEASE.2023-05-04T21-44-30Z``, the other deployment can be either another MinIO deployment or any S3-compatible location using a realtime storage class.
+For example, when replicating objects from S3 to MinIO, MinIO skips any objects tiered to the ``GLACIER`` storage class.
```
ravindk89, line 129: @harshavardhana looking at [the original pr](https://github.com/minio/minio/pull/18044/files#diff-97fbdba84496b6198f8f728ed0bf5959ad4dc3ffbd821f90296e9e148eb2f462R386) , we're only checking for `GLACIER`.

S3 docs also have `DEEP_ARCHIVE` and `GLACIER_IR`, which we aren't explicitly checking for.

If so, this might work-as is - but we also might imply that we are also skipping other glacier storage classes. just a thought.

ravindk89 (approved):

feorlen (approved):

PR #1031, opened 2023-10-10
Removing nonexistant flag

cniackz:
### Objective:

To correct our documentation because `--default-console-image` does not exist in our Operator plugin.

### Related:

* This is related to https://github.com/minio/operator/issues/1453 but will partially fix the issue, there is more work to do. I will create another PR for the missing update spot.


cniackz:
Thank you Ravind 👍

PR #1028, opened 2023-10-06
Breaking up envvars and config settings into separate settings pages

djwfyi:
Creates a slew of folders and pages nested under the mc-server page to store settings.
Breaks up all of the environment variable options formerly in the mc-server page into these separate pages.
Moves the config settings formerly in the mc admin config page to the appropriate new settings pages.

Closes #1017 

Staged:
http://192.241.195.202:9000/staging/server-breakup/linux/reference/minio-server/settings.html

PR #1038, opened 2023-10-16
DRAFT: yamlinclude initial commit

ravindk89:
This just has the basic YAMLINCLUDE logic.

I have to figure out how best to use this. The place it might make the most sense is the absurdly long `common-mc-admin-config.rst`, but before doing *that* I want to let #1028 complete.

So for now will let this sit and think about whether this actually adds value to our workflow.

### DIFF
--- a/sphinxext/yamlinclude.py
+++ b/sphinxext/yamlinclude.py
@@ -1,0 +1,96 @@
+import os.path
+
+from docutils import statemachine
+
+from docutils.parsers.rst import directives
+
+from sphinx.util.docutils import SphinxDirective
+
+import yaml
+
+class YamlInclude(SphinxDirective):
+   
+   has_content = False
+   required_arguments = 1 # Must have the file to include
+   optional_arguments = 0 # Must only specify the file to include
+
+   option_spec = {
+      "key": directives.unchanged_required,
+      "subs" : directives.unchanged,
+      "includepath" : directives.unchanged,
+      "debug" : directives.flag,
+   }
+   
+   def run(self):
+      
+      # Subs should be specified as a multi-line indented set of comma-separated key values
+      # :subs:
+      #   foo,bar
+      #   bin,baz
+      #
+      # We parse that into a dictionary for lookup later as part of the substitution loop.
+      subs = {}
+      if ('subs') in self.options:
+         for i in self.options.get('subs').splitlines():
+            kv = i.split(',')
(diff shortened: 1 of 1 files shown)


sphinxext/yamlinclude.py
```diff
@@ -0,0 +1,96 @@
...
+
+      path = os.path.normpath(os.path.join(source_dir, self.arguments[0]))
+      if ('debug') in self.options: print ("Path is " + path)
+
+      # We open the YAML file and do a lookup based on the key
+      # This should theoretically be O(1) regardless of file size
+
+      with open(path, "r") as file:
+         try:
+            yfile = yaml.safe_load(file)
```
i80and, line 66: Limit the try statement to just this line

sphinxext/yamlinclude.py
```diff
@@ -0,0 +1,96 @@
+import os.path
```
i80and, line 1: I would recommend running `black` on this file

sphinxext/yamlinclude.py
```diff
@@ -0,0 +1,96 @@
...
+   def run(self):
+      
+      # Subs should be specified as a multi-line indented set of comma-separated key values
+      # :subs:
+      #   foo,bar
+      #   bin,baz
+      #
+      # We parse that into a dictionary for lookup later as part of the substitution loop.
+      subs = {}
+      if ('subs') in self.options:
```
i80and, line 33: No need for parenthesis -- makes it look at first glance like you might be creating a tuple

sphinxext/yamlinclude.py
```diff
@@ -0,0 +1,96 @@
...
+         for i in self.options.get('subs').splitlines():
```
i80and, line 34: Just `self.options["subs"]` is fine. `get()` is unnecessary

sphinxext/yamlinclude.py
```diff
@@ -0,0 +1,96 @@
...
+            kv = i.split(',')
+            subs[kv[0]] = kv[1]
+
+      if ('debug') in self.options: print ("Substitution dictionary is "); print(subs)
+
+      key = self.options.get('key')
```
i80and, line 40: You probably want `self.options["key"]` unless you actually want to handle the case where `key` doesn't exist

djwfyi:
Will revisit at a later time, perhaps.

PR #1052, opened 2023-10-27
DOCS-1040: Clarify behavior with name conflicts on prefixes

### LINKED ISSUE #1040: Clarify behavior with name conflicts on prefixes
**Summary**

See https://github.com/minio/minio/issues/18148#issuecomment-1744000795 , https://github.com/minio/minio/issues/18262

https://min.io/docs/minio/linux/operations/concepts/thresholds.html#conflicting-objects <- at some point we dropped the error functionality here. So the PUT will succeed, but LIST operations will not return the object *or* any others within the prefix conflict
GET/HEAD will work.

The solution is to simply not use this pattern of creating explicit prefixes which do or can or might match an object name.

ravindk89:
Closes #1040 

Tried to keep this simple:

1. Use better examples
2. Try to use more clear language
3. Keep the consequence clear.

### DIFF
--- a/source/operations/concepts/thresholds.rst
+++ b/source/operations/concepts/thresholds.rst
@@ -138,4 +138,4 @@ For example, the following operations create a namespace conflicts
    PUT data/invoices/2024/january
    PUT data/invoices/2024/january/vendors.csv <- collides with existing object
 
-While GET or HEAD operations against those objects can succeed, MinIO would return an empty result set for LIST operations at the ``/invoices/2024/january`` path.
+While you can perform GET or HEAD operations against these objects, the name collision causes LIST operations to return an empty result set at the ``/invoices/2024/january`` path.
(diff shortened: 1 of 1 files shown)


source/operations/concepts/thresholds.rst
```diff
@@ -122,18 +122,20 @@ MinIO recommends using LInux operating system with an XFS based filesystem for p
...
 
 .. code-block::
-   
-   PUT <bucketname>/a/b
-   PUT <bucketname>/a/b/1.txt
+
+   PUT data/invoices/2024/january
+   PUT data/invoices/2024/january/vendors.csv <- collides with existing object
+
+While GET or HEAD operations against those objects can succeed, MinIO would return an empty result set for LIST operations at the ``/invoices/2024/january`` path.
```
feorlen, line 141: Is this an accurate understanding? 
> You are technically allowed to create these, but you won't have good time afterwards.
Seems you shouldn't be allowed in the first place, but that's a different conversation.

A minor wording tweak, if you like:
```suggestion
While you can perform GET or HEAD operations against these objects, the name collision causes LIST operations to return an empty result set at the ``/invoices/2024/january`` path.
```

feorlen (approved):
LGTM
Left a minor reword suggestion you may use if you like.

PR #1081, opened 2023-11-30
Use 9001 as the default console address port

jingsam:
As state in  https://github.com/minio/minio/pull/18561#issuecomment-1833303435 , consistent use 9001 as the console port， to avoid confusion for new users.


ravindk89 (approved):

ravindk89:
@jingsam thanks for taking the time to do this!

Issue #1016, opened 2023-10-02
[RELEASE] MinIO Server release 2023-09-27T15-22-50Z


djwfyi:
- [ ] Add info about tiering not requiring versioning ([PR #18108](https://github.com/minio/minio/pull/18108))

Something along the lines of the following at https://min.io/docs/minio/linux/administration/object-management/object-lifecycle-management.html#object-transition-tiering:

```
Starting with version X, MinIO no longer requires the remote tier to have versioning enabled.

Prior to X, you must enable versioning on the remote *and* ensure the provided remote access credentials have permissions to perform versioned operations.
```

feorlen:
Maybe this is https://github.com/minio/minio/pull/18108 ? 
tier: Avoid doing versioned operations since not required anymore

djwfyi:
Yeah, I updated the description. Thanks for the catch, @feorlen .

feorlen:
The versioned-or-not question could be a previous miscommunication. The behavior seems always (-ish) to have been like this, but not well explained. @vadmeste is confirming.

Perhaps something like this:
> The remote bucket may be versioned or not. If it is versioned, you must have s3:DeleteObjectVersion and s3:GetObjectVersion permissions.

PR #1089, opened 2023-12-19
Remote tier needs DeleteObjectVersion if remote bucket is versioned

feorlen:
If the remote tier bucket is versioned, MinIO requires additional permissions to successfully transition objects. But it is not recommended, as each object version becomes a distinct object on the remote tier.

This is not new, it was omitted from the docs. Mention, but discourage.

Staged:
http://192.241.195.202:9000/staging/DOCS-1016/linux/reference/minio-mc/mc-ilm-tier.html#transition-permissions

Fixes https://github.com/minio/docs/issues/1016

### DIFF
--- a/source/extra/examples/LifecycleManagementUser.json
+++ b/source/extra/examples/LifecycleManagementUser.json
@@ -15,7 +15,8 @@
             "Action": [
                "s3:GetObject",
                "s3:PutObject",
-               "s3:DeleteObject"
+               "s3:DeleteObject",
+               "s3:DeleteObjectVersion"
             ],
             "Effect": "Allow",
             "Resource": [
--- a/source/extra/examples/LifecycleManagementUser.json
+++ b/source/extra/examples/LifecycleManagementUser.json
@@ -16,7 +16,6 @@
                "s3:GetObject",
                "s3:PutObject",
                "s3:DeleteObject",
-               "s3:DeleteObjectVersion"
             ],
             "Effect": "Allow",
             "Resource": [
--- a/source/extra/examples/LifecycleManagementUser.json
+++ b/source/extra/examples/LifecycleManagementUser.json
@@ -15,7 +15,7 @@
             "Action": [
                "s3:GetObject",
                "s3:PutObject",
-               "s3:DeleteObject",
+               "s3:DeleteObject"
             ],
             "Effect": "Allow",
             "Resource": [
--- a/source/reference/minio-mc/mc-ilm-tier.rst
+++ b/source/reference/minio-mc/mc-ilm-tier.rst
@@ -92,15 +92,22 @@ Transition Permissions
 
 Object transition lifecycle management rules require additional permissions on the remote storage tier. 
 Specifically, MinIO requires the remote tier credentials provide read, write, list, and delete permissions.
-If the remote bucket is versioned, the ``s3:DeleteObjectVersion`` permission is also required.
 
-For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permissions for transitioning versioned objects into and out of the remote tier:
+For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permissions for transitioning objects into and out of the remote tier:
 
 .. literalinclude:: /extra/examples/LifecycleManagementUser.json
    :language: json
    :class: copyable
 
 Modify the ``Resource`` for the bucket into which MinIO tiers objects.
+
+.. admonition:: Avoid remote tier object versioning
+   :class: important
+
+   MinIO strongly recommends against object versioning for remote tiers.
+   If the remote tier bucket is versioned, each source object version is transitioned to a *unique object* in the remote tier.
+   
+   If your environment requires versioning for the remote tier, you must also allow the ``s3:DeleteObjectVersion`` permission.
 
 Defer to the documentation for the supported tiering targets for more complete information on configuring users and permissions to support MinIO tiering:
 
--- a/source/reference/minio-mc/mc-ilm-tier.rst
+++ b/source/reference/minio-mc/mc-ilm-tier.rst
@@ -81,7 +81,7 @@ To create tiers for object transition, MinIO requires the following administrati
 - :policy-action:`admin:SetTier`
 - :policy-action:`admin:ListTier`
 
-For example, the following policy provides permission for configuring object transition lifecycle management rules on any bucket in the cluster:
+For example, the following policy provides sufficient permissions for configuring object transition lifecycle management rules on any bucket in the cluster:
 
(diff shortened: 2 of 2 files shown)


harshavardhana:
This is not true anymore @feorlen // cc @krisis @vadmeste

vadmeste:
@harshavardhana this is not obvious but it is true; currently we require s3:DeleteObjectVersion if the remote bucket is versioned, otherwise tiering will not work at all. I tried to remove this requirement before but then @kannappanr told me we need to do a hard rather than a soft delete in that case so our users will not get a lot of unused versions in their AWS S3 account.

source/reference/minio-mc/mc-ilm-tier.rst
```diff
@@ -92,8 +92,9 @@ Transition Permissions
 
 Object transition lifecycle management rules require additional permissions on the remote storage tier. 
 Specifically, MinIO requires the remote tier credentials provide read, write, list, and delete permissions.
+If the remote bucket is versioned, delete version permission is also required.
 
-For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permission for transitioning objects into and out of the remote tier:
+For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permission for transitioning versioned objects into and out of the remote tier:
```
djwfyi, line 97:
```suggestion
For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permissions for transitioning versioned objects into and out of the remote tier:
```

source/reference/minio-mc/mc-ilm-tier.rst
```diff
@@ -92,8 +92,9 @@ Transition Permissions
 
 Object transition lifecycle management rules require additional permissions on the remote storage tier. 
 Specifically, MinIO requires the remote tier credentials provide read, write, list, and delete permissions.
+If the remote bucket is versioned, delete version permission is also required.
```
djwfyi, line 95:
```suggestion
If the remote bucket is versioned, the ``s3:DeleteObjectVersion`` permission is also required.
```

source/reference/minio-mc/mc-ilm-tier.rst
```diff
@@ -81,7 +81,7 @@ To create tiers for object transition, MinIO requires the following administrati
 - :policy-action:`admin:SetTier`
 - :policy-action:`admin:ListTier`
 
-For example, the following policy provides permission for configuring object transition lifecycle management rules on any bucket in the cluster:.
+For example, the following policy provides permission for configuring object transition lifecycle management rules on any bucket in the cluster:
```
djwfyi, line 84:
```suggestion
For example, the following policy provides sufficient permissions for configuring object transition lifecycle management rules on any bucket in the cluster:
```

djwfyi (approved):
Couple of suggestions to consider, since the policies contain more than a single permission.

Otherwise, LGTM.

ravindk89 (approved):
LGTM post daryl's suggestions.

feorlen:
Added docs team review suggestions while we sort out the nuances of behavior.

harshavardhana:
> @harshavardhana this is not obvious but it is true; currently we require s3:DeleteObjectVersion if the remote bucket is versioned, otherwise tiering will not work at all. I tried to remove this requirement before but then @kannappanr told me we need to do a hard rather than a soft delete in that case so our users will not get a lot of unused versions in their AWS S3 account.

Okay I see looks like then its needed, however I think we should strongly discourage from users to carrying versioning to their warm-tier.

feorlen:
> we should strongly discourage from users to carrying versioning to their warm-tier.

The [Object Lifecycle Management](https://min.io/docs/minio/linux/administration/object-management/object-lifecycle-management.html#object-transition-tiering) page (one place a more detailed discussion would go) doesn't offer an opinion on whether to version or not. It only talks about behavior concerning current or not-current versions of an object. The versioning page doesn't mention this at all.

If the remote should _not_ be versioned even if the corresponding hot is, what is useful to say should someone choose to do it anyway? Are the extraneous versions ignored and/or not accessible? @harshavardhana

ravindk89:
@krisis should be back tomorrow and may comment, or @poornas who I think has also touched this subsystem, but IIRC versioning on the remote is not required because the nature of transition applies a UUID to the object already. Versioning on the remote becomes unnecessary, because we would not "write" to the same UUID twice.

krisis:
>If the remote should not be versioned even if the corresponding hot is, what is useful to say should someone choose to do it anyway? Are the extraneous versions ignored and/or not accessible? @harshavardhana

Each version of an object on the hot tier is transitioned to a unique object on the warm tier. MinIO's tiering feature doesn't require users to enable versioning on warm tier. @ravindk89's guess based on the uuid is correct. 

In the event that user has versioning enabled on this bucket (in the warm tier), they require the relevant IAM permission to delete versions.

We should recommend against enabling versioning on the (warm tier) bucket used for tiering.

harshavardhana:
> We should recommend against enabling versioning on the (warm tier) bucket used for tiering.

Yes a strong a recommendation

source/extra/examples/LifecycleManagementUser.json
```diff
@@ -15,7 +15,7 @@
             "Action": [
                "s3:GetObject",
                "s3:PutObject",
-               "s3:DeleteObject"
+               "s3:DeleteObject",
```
vadmeste, line 18: I don't think this is a valid JSON with this extra comma

source/reference/minio-mc/mc-ilm-tier.rst
```diff
@@ -93,14 +93,22 @@ Transition Permissions
...
-For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permission for transitioning objects into and out of the remote tier:
+For example, if the remote storage tier implements AWS IAM policy-based access control, the following policy provides the necessary permissions for transitioning objects into and out of the remote tier:
 
 .. literalinclude:: /extra/examples/LifecycleManagementUser.json
    :language: json
    :class: copyable
 
 Modify the ``Resource`` for the bucket into which MinIO tiers objects.
 
+.. admonition:: Avoid remote tier object versioning
```
vadmeste, line 104: this is probably not very clear ? it is better to clearly state that we are talking about versioning in the tiering backend. Something like this maybe? `Avoid enabling versioning in the remote tier`

source/extra/examples/LifecycleManagementUser.json
feorlen: Oops thank you!

source/reference/minio-mc/mc-ilm-tier.rst
feorlen: I like yours better! Also did some minor edits in the paragraph below.

PR #1095, opened 2023-12-24
Fix the name of RoleArn param in AssumeRoleWithWebIdentity requests

tsuzu:
In `AssumeRoleWithWebIdentity` requests, the RoleArn form paramter is designated to specify the role to be assumed. It is crucial to note that the correct key is RoleArn, not RoleARN.

When I used the "RoleARN" form key instead of "RoleArn," I encountered an error similar to the following:
```xml
<ErrorResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/"><Error><Type></Type><Code>InvalidParameterValue</Code><Message>Role arn:minio:iam:::role/dummy-internal does not exist</Message></Error><RequestId>XXX</RequestId></ErrorResponse>
```

Despite this, certain documentation erroneously employs RoleARN. To rectify this inconsistency, I have replaced the instances of RoleARN with the correct RoleArn. However, for other documents that use RoleARN without specifying the form parameter, no modifications have been made.

Ref: https://github.com/minio/minio/blob/ba245c6c463c1ae9b652aa4cf44b2e603c68b045/cmd/sts-handlers.go#L50C2-L50C12


harshavardhana (approved):

ravindk89 (approved):

ravindk89:
Thank you for your contribution @tsuzu , we really appreciate it!

PR #1100, opened 2023-12-29
PowerShell commands and a request type for access management fix

### LINKED ISSUE #1064: Access Management docs - readonly vs. PUT in example table for Operations user
In the [built in policy](https://min.io/docs/minio/kubernetes/openshift/administration/identity-access-management/policy-based-access-control.html#built-in-policies) chapter of the [access management](https://min.io/docs/minio/kubernetes/openshift/administration/identity-access-management/policy-based-access-control.html) documentation is an example with three different users:
 
> <p>For example, consider the following table of users. Each user is assigned
> a <a class="reference internal" href="https://min.io/docs/minio/kubernetes/openshift/administration/identity-access-management/policy-based-access-control.html#minio-policy-built-in"><span class="std std-ref">built-in policy</span></a> or
> a supported <a class="reference internal" href="https://min.io/docs/minio/kubernetes/openshift/administration/identity-access-management/policy-based-access-control.html#minio-policy-actions"><span class="std std-ref">action</span></a>. The table
> describes a subset of operations a client could perform if authenticated
> as that user:</p>
> <div class="table-responsive scrollbar">
> 
> User | Policy | Operations
> -- | -- | --
> Operations | `readwrite` on `finance` bucket `readonly` on `audit` bucket | `PUT` and `GET` on `finance` bucket. `PUT` on `audit` bucket
> Auditing | ... | ...
> Admin | ... | ...

Source:
https://github.com/minio/docs/blob/2c5502f3c67730c64c64f86e75b704c7745b87ba/source/security/minio-identity-management/policy-based-access-control.rst?plain=1#L114-L119

Question:
Under the `Operations` user row the documentation states that the `readonly` policy on the `audit` bucket is applied, but on the operations column `PUT` is mentioned for the `audit` bucket.

Shouldn't that be either `GET` on the `audit` bucket as the operation or `writeonly` as the policy in case `PUT` shall be allowed?

### LINKED ISSUE #1086: Healthcheck documentation fails to mention that healthchecks are uncallable if User-Agent string contains "Mozilla" and "MINIO_BROWSER_REDIRECT" environment variable is not set to False
When MINIO_BROWSER_REDIRECT is not set, or has been explicitly set to `true`, MinIO has middleware which will detect if the User Agent string contains the sub-string `Mozilla`. If it does detect the sub-string, MinIO will redirect the request to the Admin Console. This includes all healthcheck endpoints so all healthchecks will result in a 307 response, and if the redirect is followed always result in a 200 because the login screen returns a 200.

The documentation uses curl as an example on how to call the endpoints, and will not run into the issue as curl sets the User Agent string to `curl/[version]` (see: https://everything.curl.dev/http/requests/user-agent). However if trying the equivalent commands in PowerShell (including the curl alias in PowerShell) you will run into the redirect as the default User-Agent string for PowerShell contains `Mozilla` (see https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.4#-useragent)

This can be demonstrated using the cluster maintenance check on single node cluster (as it should always fail) and calling the following and noting the return code is 307 as opposed to 412:
`Invoke-WebRequest -Uri https://minio:9000/minio/health/cluster?maintenance=true -MaximumRedirection 0`

Without the `-MaximumRedirection` parameter you follow the redirect and get a 200 response instead of the 412:
`Invoke-WebRequest -Uri https://minio:9000/minio/health/cluster?maintenance=true`

Getting the proper return code may be accomplished by setting `-UserAgent` to `""`:
`Invoke-WebRequest -Uri https://minio:9000/minio/health/cluster?maintenance=true -UserAgent ""`

djwfyi:
- Fixes a request type in an access management table

Closes #1064

- Adds an admonition about PowerShell commands to Healthcheck API page

Closes #1086

Staged:
- [Request type fix](http://192.241.195.202:9000/staging/cleanups/windows/administration/identity-access-management/policy-based-access-control.html#userpolicy.writeonly)
- [Healthcheck API](http://192.241.195.202:9000/staging/cleanups/windows/operations/monitoring/healthcheck-probe.html)

### DIFF
--- a/source/operations/monitoring/healthcheck-probe.rst
+++ b/source/operations/monitoring/healthcheck-probe.rst
@@ -16,21 +16,6 @@ endpoints return an HTTP status code indicating whether the underlying
 resource is healthy or satisfies read/write quorum. MinIO exposes no other data
 through these endpoints.
 
-.. cond:: windows
-
-   .. admonition:: Checks Using PowerShell
-      :class: note
-
-      When using PowerShell to run the following the commands, you must explicitly set ``-UserAgent ""``.
-      Otherwise, the healthcheck redirects to the Admin Console instead of the desired Healthcheck API.
-
-      For PowerShell commands, you could replace each of the ``curl`` examples below with something like the following:
-
-      .. code-block:: powershell
-         :class: copyable
-
-         Invoke-WebRequest -Uri https://minio.example.net:9000/minio/health/cluster?maintenance=true -UserAgent ""
-
 Node Liveness
 -------------
 
--- a/source/operations/monitoring/healthcheck-probe.rst
+++ b/source/operations/monitoring/healthcheck-probe.rst
@@ -16,6 +16,21 @@ endpoints return an HTTP status code indicating whether the underlying
 resource is healthy or satisfies read/write quorum. MinIO exposes no other data
 through these endpoints.
 
+.. cond:: windows
+
+   .. admonition:: Checks Using PowerShell
+      :class: note
+
+      When using PowerShell to run the following the commands, you must explicitly set ``-UserAgent ""``.
+      Otherwise, the healthcheck redirects to the Admin Console instead of the desired Healthcheck API.
+
+      For PowerShell commands, you could replace each of the ``curl`` examples below with something like the following:
+
+      .. code-block:: powershell
+         :class: copyable
+
+         Invoke-WebRequest -Uri https://minio.example.net:9000/minio/health/cluster?maintenance=true -UserAgent ""
+
 Node Liveness
 -------------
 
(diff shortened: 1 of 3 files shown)


source/operations/monitoring/healthcheck-probe.rst
```diff
@@ -16,6 +16,21 @@ endpoints return an HTTP status code indicating whether the underlying
 resource is healthy or satisfies read/write quorum. MinIO exposes no other data
 through these endpoints.
 
+.. cond:: windows
+
+   .. admonition:: Checks Using PowerShell
+      :class: note
+
+      When using PowerShell to run the following the commands, you must explicitly set ``-UserAgent ""``.
+      Otherwise, the healthcheck redirects to the Admin Console instead of the desired Healthcheck API.
```
harshavardhana, line 25: This is not needed anymore @djwfyi MinIO server now handles health requests from user-agents with Mozilla in them.

djwfyi: Thanks. Removed the admonition.

ravindk89 (approved):

feorlen (approved):

PR #1109, opened 2024-01-15
Update `mc share download` syntax

mozafar:
In following page the syntax section is not for download:
https://min.io/docs/minio/linux/reference/minio-mc/mc-share-download.html#syntax
This PR will changes syntax from upload  to download


ravindk89 (approved):

ravindk89:
Thanks for the assist @mozafar !

PR #1111, opened 2024-01-18
Adding mc version info for license add or update

djwfyi:
Per internal discussion, this adds some admonitions for mc license commands and the SUBNET registration section of the troubleshooting page.

### DIFF
--- a/source/operations/troubleshooting.rst
+++ b/source/operations/troubleshooting.rst
@@ -79,6 +79,10 @@ You can restart through the Console by selecting :guilabel:`Restart` in the top 
 
       You can register for SUBNET from the command line.
 
+      .. important:: 
+
+         Registering your deployment with SUBNET or updating the license for a deployment requires :ref:`MinIO Client <minio-client> version ``RELEASE.2023-11-20T16-30-59Z`` or newer.
+
       Refer to :mc:`mc license register` for instructions.
 
       For clusters without direct Internet access, refer to the instructions in the :ref:`airgap example <minio-license-register-airgap>` of the :mc:`mc license register` documentation.
--- a/source/reference/minio-mc/mc-license-register.rst
+++ b/source/reference/minio-mc/mc-license-register.rst
@@ -11,6 +11,10 @@
 .. mc:: mc support register
 .. mc:: mc license register
 
+.. important:: 
+
+   ``mc license register`` requires :ref:`MinIO Client <minio-client> version ``RELEASE.2023-11-20T16-30-59Z`` or newer.
+
 Command History
 ---------------
 
--- a/source/reference/minio-mc/mc-license-update.rst
+++ b/source/reference/minio-mc/mc-license-update.rst
@@ -10,6 +10,10 @@
 
 .. mc:: mc license update
 
+.. important:: 
+
+   ``mc license update`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or newer.
+
 Description
 -----------
 


feorlen (approved):

feorlen:
Oh an idea I just thought of. Perhaps something to the effect of "If you are not able to update, register using the console"

source/reference/minio-mc/mc-license-register.rst
```diff
@@ -11,6 +11,10 @@
 .. mc:: mc support register
 .. mc:: mc license register
 
+.. important:: 
+
+   ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or newer.
+
 Command History
```
anjalshireesh, line 18: The `Command History` was useful earlier for customers to know which command they can use with their version of `mc`. 

However, now that the older versions of `mc` will not work at all for registration, it might be better to remove this section altogether.

anjalshireesh (changes requested):
Looks good @djwfyi , I just have one suggestion related to removing the `Command History` section.

source/reference/minio-mc/mc-license-register.rst
ravindk89: @anjalshireesh does that mean, in general, that users *must* update their `minio` server before registration?
Since generally we want `mc` and `minio` to be closely aligned.

What versions of `mc` are effectively no longer compatible with SUBNET?

source/reference/minio-mc/mc-license-update.rst
```diff
@@ -10,6 +10,10 @@
 
 .. mc:: mc license update
 
+.. important:: 
+
+   ``mc license update`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or newer.
```
ravindk89, line 15: I think we'll need to suggest a server version here too. Most users/customers want a minimum viable target.

ravindk89 (changes requested):
Two thoughts, one of which is for eng. to follow up on.

source/reference/minio-mc/mc-license-register.rst
djwfyi: That makes sense. Removing.

source/reference/minio-mc/mc-license-update.rst
djwfyi: Added.

source/reference/minio-mc/mc-license-register.rst
anjalshireesh: @ravindk89 

> does that mean, in general, that users must update their minio server before registration?
Since generally we want mc and minio to be closely aligned.

No, only `mc` needs to be upgraded. While somehow the message that `mc` and `minio` should be closely aligned has got popular, it is not correct and ideally latest `mc` should always maintain compatibility with older `minio` versions. cc @abperiasamy 

Also,  there are other alternatives if upgrading mc is not possible for some important (customer-specific) reason, like using console for registration or using a lower level mc command.

> What versions of mc are effectively no longer compatible with SUBNET?

Any version lower than `RELEASE.2023-11-20T16-30-59Z` cannot be used for `mc license register` or `mc license update`. For already-registered clusters, older mc versions will still work for the `mc support` commands.

source/reference/minio-mc/mc-license-update.rst
anjalshireesh: No, there is absolutely no need to upgrade `minio` server and it will be a very unfair ask for customers to upgrade their servers. Please revert this change.

source/reference/minio-mc/mc-license-update.rst
```diff
@@ -10,6 +10,11 @@
 
 .. mc:: mc license update
 
+.. important:: 
+
+   ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
```
anjalshireesh, line 15: The command is wrongly mentioned as `mc license register` instead of `mc license update`.
Also, there is is no requirement of minio server minimum version to be `RELEASE.2023-11-20T22-40-`.

anjalshireesh (changes requested):

source/reference/minio-mc/mc-license-update.rst
djwfyi: I thought the standard was that the MinIO Client and the MinIO Server must stay in sync?
Otherwise, the MinIO Client can show options that won't work for the MinIO Server.

source/reference/minio-mc/mc-license-update.rst
djwfyi: Good catch on the command.
But standard in docs is that the MC client and the MinIO Server should be in sync.

source/reference/minio-mc/mc-license-update.rst
harshavardhana: mc must be upgraded if server is upgraded not the other way around.

Yes if they upgrade mc we will ask them to upgrade MinIO if they need new functionality but it is not a hard requirement unless you want the new feature in mc to work.

source/reference/minio-mc.rst
```diff
@@ -44,6 +44,12 @@ See :ref:`minio-mc-commands` for a list of supported commands.
    is the best option for applications which trigger AGPLv3 obligations where
    open-sourcing the application is not an option.   
 
+Version
+-------
+
+The MinIO Client releases separately from the MinIO Server.
+You must use the MinIO Client version that releases at the same time (for example, on the same day) as the MinIO Server version or a later version of the MinIO Client.
```
feorlen, line 51: This text could also be the same as whatever you decide to do for troubleshooting.rst

source/operations/troubleshooting.rst
```diff
@@ -82,7 +82,9 @@ You can restart through the Console by selecting :guilabel:`Restart` in the top
       .. important:: 
       
          ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
-         MinIO strongly recommends always using the MinIO Client and MinIO Server versions released together.
+         Use the MinIO Client released at the close to the same time (for example, released on the same day) or after the version of MinIO Server you are using.
```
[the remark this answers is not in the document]
djwfyi: Oh, yeah, that's MUCH better. Thanks!

source/reference/minio-mc.rst
djwfyi: Replaced throughout.

feorlen (approved):

ravindk89 (changes requested):
Forgot to hit 'send'

source/reference/minio-mc.rst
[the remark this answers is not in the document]
djwfyi: Tweaked and rearranged some, but generally taking this wording in some combination with what @feorlen suggested. Applied throughout.

source/reference/minio-mc/mc-license-update.rst
djwfyi: Applying some "thread the needle" wording to allow for the required MC version while cautioning that skewing too far from the Server version may result in warnings or errors in other commands/places.

source/reference/minio-mc/mc-license-register.rst
```diff
@@ -11,32 +11,12 @@
...
-     - First available
-
-   * - RELEASE.2022-02-13T23-26-13Z 
-     - ``mc support register``
-     - Command moved to an ``mc support`` function
-
-   * - RELEASE.2022-07-15T09-20-55Z
-     - ``mc license register``
-     - Command moved to an ``mc license`` function
+   ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
```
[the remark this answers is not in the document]
djwfyi: Ah, yeah, I had added that back when I still had a hard "must align versions" in my head.

source/operations/troubleshooting.rst
```diff
@@ -82,7 +82,13 @@ You can restart through the Console by selecting :guilabel:`Restart` in the top
...
       
          ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
-         Use a MinIO Client version no older than your MinIO Server. For example, a MinIO Client released the same day or later than your MinIO Server version.
+
+         For best functionality and compatibility, use a MinIO Client version released closely to your MinIO Server version. 
+         For example, a MinIO Client released the same day or later than your MinIO Server version.
+
+         You can install a version of the MinIO Client that is more recent than the MinIO Server version. 
+         However, if the MinIO Client version skews too far from the MinIO Server version, you may see increased warnings or errors as a result of the differences.
+         For example, while core S3 APIs around copying (:mc`mc cp`) may remain unchanged, some features or flags may only be available or stable if the client and server versions are aligned.
```
feorlen, line 91: formatting typo in ref

djwfyi: So there is! Ack!

feorlen: yay catching it now and not later 🤝

source/operations/troubleshooting.rst
```diff
@@ -79,6 +79,19 @@ You can restart through the Console by selecting :guilabel:`Restart` in the top
...
+         ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
+
+         For best functionality and compatibility, use a MinIO Client version released closely to your MinIO Server version. 
+         For example, a MinIO Client released the same day or later than your MinIO Server version.
+
+         You can install a version of the MinIO Client that is more recent than the MinIO Server version. 
+         However, if the MinIO Client version skews too far from the MinIO Server version, you may see increased warnings or errors as a result of the differences.
+         For example, while core S3 APIs around copying (:mc:`mc cp`) may remain unchanged, some features or flags may only be available or stable if the client and server versions are aligned.
+
+         If you are unable to upgrade the MinIO Client to the required or later version, register using the Console instead.
```
ravindk89, line 93: :thinking: I think it's enough to link to the Version section of mc here as "For more information on managing version alignment between minio and mc, see blah". otherwise this is a long admonition.

source/operations/troubleshooting.rst
```diff
@@ -79,6 +79,19 @@ You can restart through the Console by selecting :guilabel:`Restart` in the top
 
       You can register for SUBNET from the command line.
 
+      .. important:: 
+      
+         ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later and MinIO Server ``RELEASE.2023-11-20T22-40-07Z`` or later.
```
ravindk89, line 84: Align with mc license register

```suggestion
         ``mc license register`` requires :ref:`MinIO Client <minio-client>` version ``RELEASE.2023-11-20T16-30-59Z`` or later.
```

source/reference/minio-mc.rst
```diff
@@ -44,6 +44,18 @@ See :ref:`minio-mc-commands` for a list of supported commands.
    is the best option for applications which trigger AGPLv3 obligations where
    open-sourcing the application is not an option.   
 
+Version
```
ravindk89, line 47: make this linkable, possibly rename ot "Version Alignment With MinIO Server"

ravindk89 (approved):
LGTM mod nits, which are mostly just removing stuff and linking around.

PR #1121, opened 2024-02-06
Reminder to install mc, not Midnight Commander

feorlen:
Add admonition to the Linux `mc` install steps about installing the correct binary from `dl.min.io`. If you use your native package manager, you are very likely to get GNU Midnight Commander instead. 

Staged
http://192.241.195.202:9000/staging/not-midnight-commander/linux/reference/minio-mc.html#install-mc

### DIFF
--- a/source/includes/minio-mc-installation.rst
+++ b/source/includes/minio-mc-installation.rst
@@ -51,6 +51,13 @@
 
          mc --help
 
+      .. admonition:: Install from the MinIO Download Page
+         :class: note
+
+         To ensure you install the correct binary, download ``mc`` from the `MinIO Download Page <https://dl.min.io/client/mc/release/>`__.
+         Other repositories, including those used by common Linux package managers, may have other software with a similar name.
+	 
+
    .. tab-item:: macOS
 
       .. code-block:: shell
--- a/source/includes/minio-mc-installation.rst
+++ b/source/includes/minio-mc-installation.rst
@@ -54,9 +54,12 @@
       .. admonition:: Install from the MinIO Download Page
          :class: note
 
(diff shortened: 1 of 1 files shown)


source/includes/minio-mc-installation.rst
```diff
@@ -51,6 +51,13 @@
 
          mc --help
 
+      .. admonition:: Install from the MinIO Download Page
+         :class: note
+
+         To ensure you install the correct binary, download ``mc`` from the `MinIO Download Page <https://dl.min.io/client/mc/release/>`__.
+         Other repositories, including those used by common Linux package managers, may have other software with a similar name.
```
ravindk89, line 58: So I *think* we can safely go farther here.

I am actually seeing `minio-client` in the arch/extra repos. Nothing in the ubuntu repos (thankfully?), and I can't be damned to figure our RHEL right now.

That  said, spitballing:

```suggestion
         MinIO does not officially publish it's binaries to common Linux repositories or package managers (Ubuntu, RHEL, Archlinux/AUR).
         The only official source of MinIO binaries is the <download page URL>.
         
         If installing the MinIO Client through your package manager, take extra steps to ensure you download the correct package.
         For example, some operating systems will install Midnight Commander ``mc`` by default, or will rename the MinIO client to ``mcli``.
         
         All documentation assumes the installation of the *official* ``mc`` client binary through the download page *only* with no changes to binary naming.
```

ravindk89 (changes requested):

source/includes/minio-mc-installation.rst
ravindk89: I did this as a suggestion and did it wrong :face_with_spiral_eyes:

feorlen:
@ravindk89 approved in a separate conversation, merging

PR #1126, opened 2024-02-09
Allow links into closed tabs

feorlen:
JS update to allow a link into a closed tab. 

Tested with 
```
:mc-conf:`~browser.hsts_seconds`
```
which goes to a tab in [this section](https://min.io/docs/minio/linux/reference/minio-server/settings/console.html#strict-transport-security)

@rushenn Ravind and Andrea came up with this, need a JS intervention to ensure it's actually reasonable. 🙏🏻 

Staged
http://192.241.195.202:9000/staging/link-into-tab/linux/reference/minio-server/settings/console.html#strict-transport-security

### DIFF
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -505,12 +505,9 @@ window.addEventListener("DOMContentLoaded", (event) => {
     }
   })();
 
-// ------------------------------------------------
-// Tab switching for hidden anchors
-// ------------------------------------------------
-
-  // Target ID - where the browser should scroll to
-  const hashElem = document.getElementById(window.location.hash.substring(1));
+  // ------------------------------------------------
+  // Tab switching for hidden anchors
+  // ------------------------------------------------
 
   // If the element is hidden, neither scrollIntoView nor hashScroll will work.
   // This function checks whether the element is visible or not.
@@ -518,22 +515,43 @@ window.addEventListener("DOMContentLoaded", (event) => {
     var style = window.getComputedStyle(elem);
     return (elem.offsetWidth > 0 || elem.offsetHeight > 0) && style.display !== 'none' && style.visibility !== 'hidden';
   }
-
-  // Check if the hash is valid
-  // If it is, but not visible, the corresponding tab is also not visible.
-  // Make the tab active and scroll to it
-  if(hashElem) {
-    if(!isHashElemeVisible(hashElem)) {
-      hashElem.closest(".sd-tab-content").previousElementSibling.click();
-
-      setTimeout(() => {
-        hashElem.scrollIntoView();
-      }, 100);
-    }
-    else {
-      hashElem.scrollIntoView();
+  
+  function hashScroll() {
+    // Target ID - where the browser should scroll to
+    const hashElem = document.getElementById(window.location.hash.substring(1));
+  
+    // Check if the hash is valid
+    // If it is, but not visible, the corresponding tab is also not visible.
+    // Make the tab active and scroll to it
+    if(hashElem) {
+      if(!isHashElemeVisible(hashElem)) {
+        hashElem.closest(".sd-tab-content").previousElementSibling.click();
+  
+        setTimeout(() => {
+          hashElem.scrollIntoView({ behavior: 'smooth' });
+        }, 100);
+      }
+      else {
+        hashElem.scrollIntoView({ behavior: 'smooth' });
+      }
     }
   }
+  
+  // Listen for hash changes
+  window.addEventListener("hashchange", () => {
+    hashScroll();
+  });
+
+  // Listen for click events on internal references
+  const references = document.querySelectorAll("a.reference.internal");
+
+  if(references.length > 0) {
+    references.forEach((item) => {
+      item.addEventListener("click", () => {
+        hashScroll();
+      });
+    });
+  }
+  
+  hashScroll()
 });
-
-
--- a/source/_static/js/main.js
+++ b/source/_static/js/main.js
@@ -504,15 +504,36 @@ window.addEventListener("DOMContentLoaded", (event) => {
       });
     }
   })();
+
 // ------------------------------------------------
 // Tab switching for hidden anchors
 // ------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


source/_static/js/main.js
```diff
@@ -504,4 +504,12 @@ window.addEventListener("DOMContentLoaded", (event) => {
       });
     }
   })();
-});
\ No newline at end of file
+});
+
+
+window.addEventListener("hashchange", function() {
+  document.getElementById((document.getElementById(window.location.hash.substring(1)))
```
ravindk89, line 511: So I think we also need this somewhere above line 507, so that if someone links directly to a 'hidden' anchor item, it gets picked up.

This only triggers if the hash _changes_ on the page, not if the anchor already exists.

ravindk89 (dismissed):

source/_static/js/main.js
```diff
@@ -504,4 +504,12 @@ window.addEventListener("DOMContentLoaded", (event) => {
       });
     }
   })();
-});
\ No newline at end of file
+});
```
ravindk89, line 507:
```suggestion
// ------------------------------------------------
// Tab switching for hidden anchors
// ------------------------------------------------

    document.getElementById((document.getElementById(window.location.hash.substring(1)))
        .closest("div")
        .previousSibling.getAttributes("for"))
    .click();

});
```

ravindk89: Maybe? Not sure if I got the spaces correct.

feorlen: I am but a lowly C programmer, but I will apply this and see if it works.

feorlen: @ravindk89 fresh stage with this change

ravindk89: Hm. We're missing something here, because now the scrolldown doesn't trigger at all. Something in this logic must be interrupting the anchor scroll.

There is a `scrollIntoView()` function but I'm not sure how to apply it here.

source/_static/js/main.js
```diff
@@ -504,4 +504,21 @@ window.addEventListener("DOMContentLoaded", (event) => {
...
   })();
-});
\ No newline at end of file
+// ------------------------------------------------
+// Tab switching for hidden anchors
+// ------------------------------------------------
+
+    document.getElementById((document.getElementById(window.location.hash.substring(1)))
+        .closest("div")
+        .previousSibling.getAttributes("for"))
```
rushenn, line 513: There seems to be an error in this function:

Screenshot 2024-02-21 at 12 30 09

I think we can add a little more context and rewrite it as shown below:

```
  // Target ID - where the browser need to scroll
  const hashElem = document.getElementById(window.location.hash.substring(1));

  // If the element is hidden, neither scrollIntoView nor hashScroll would work. So this function is to check whether the elem is visible or not.
  function isHashElemeVisible(elem) {
    var style = window.getComputedStyle(elem);
    return (elem.offsetWidth > 0 || elem.offsetHeight > 0) && style.display !== 'none' && style.visibility !== 'hidden';
  }

  // First check if the hash is valid. 
  // And if it's valid but hidden, it means the corresponding tab is hidden. So, make the tab active and  
  if(hashElem) {
    if(!isHashElemeVisible(hashElem)) {
      hashElem.closest(".sd-tab-content").previousElementSibling.click();
    
      setTimeout(() => {
        hashElem.scrollIntoView();
      }, 100);
    }
    else {
      hashElem.scrollIntoView();
    }
  }
});
```

rushenn:
@feorlen Sorry for the delayed response, somehow I missed the notification for this conversation. Added a comment, please take a look.

source/_static/js/main.js
```diff
@@ -504,15 +504,36 @@ window.addEventListener("DOMContentLoaded", (event) => {
...
 
+      setTimeout(() => {
+        hashElem.scrollIntoView();
+      }, 100);
+    }
+    else {
+      hashElem.scrollIntoView();
+    }
+  }
 });
```
ravindk89, line 539: @rushenn do we duplicate the code from above down here, or is this chunk OK as is?

rushenn: @ravindk89 

I've added the `else` part because in some cases, I've noticed that Chrome fails to scroll even when the hash is valid. So, adding `scrollIntoView` ensures consistent scrolling across all browsers.

The code below is no longer needed. Can we remove it, @feorlen?

```
window.addEventListener("hashchange", function() {
  document.getElementById((document.getElementById(window.location.hash.substring(1)))
          .closest("div")
	  .previousSibling.getAttribute("for"))
        .click();
});
```

rushenn: Also, we might need to disable the smooth scroll behavior on `_layout.scss@51` and `_base.scss@14`. Smooth scrolling might prevent the browser from completing the scroll. We can test this behavior once merged and make this change only if necessary.

feorlen: @rushenn  Dead code removed and PR restaged. It's not working for me, however.
http://192.241.195.202:9000/staging/link-into-tab/linux/reference/minio-server/settings/console.html#content-security-policy

rushenn: It seems to be working for me. I tried the following URL, and I can see that the tab is opened and scrolled to the given anchor as well.

Do you see any errors related to the function in the browser console? 

http://192.241.195.202:9000/staging/link-into-tab/linux/reference/minio-server/settings/console.html#mc-conf.browser.hsts_seconds

rushenn:
@ravindk89 Got it, currently, it works only on the page load. I didn't think about the hash-change on click. Will send a fix soon.

rushenn:
@ravindk89 @feorlen Updated the code here https://github.com/minio/docs/pull/1126/commits/2659212edb0e1cd6203bbf0c7f51221792d6266f

It should work now on both hash change and clicks as well.

feorlen:
Restaged and working for me. 🎉 
@rushenn @ravindk89

ravindk89:
Great! Let's shipit

feorlen:
Tested after merge, works 🎉

PR #1148, opened 2024-03-06
Update minio-mc.rst the error parameters, uppercase JSON is not avail…

opencmit2:
Update minio-mc.rst the error parameters, uppercase JSON is not available。

image


djwfyi (approved):
Thanks for the catch and the fix!

PR #1154, opened 2024-03-13
Update top navigation to match the parent site

rushenn:


djwfyi (approved):
Top nav looks good and matches what's on the main site.

PR #1170, opened 2024-03-25
Clarify mc support inspect usage

feorlen:
If a customer's SUBNET issue requires information from `mc support inspect` we will give them the correct command to run.

Staged:
http://192.241.195.202:9000/staging/mc-support-inspect/linux/reference/minio-mc/mc-support-inspect.html


djwfyi (approved):
LGTM

PR #1192, opened 2024-04-19
Add concept pages for Scanner and Healing

### LINKED ISSUE #1168: Improve documentation about the scanner
Research and document the finer workings of the scanner.

Below are some questions we may consider answering:

- How does the scanner determine the order of bucket/object scanning?
        New buckets first, then random.
- How is "Last full scan time" calculated, and what does it mean?
- How does a machine restart affect the scanner?
        Scanning process starts over from scratch.
- What is the meaning of the scanner metric, and how can I use that to determine how long it will take to finish scanning?

djwfyi:
Adds concept pages for healing and the scanner.
Updates other related files for better context.
    
Adds ref links and other related updates.
    
Closes #1168

Staging info:

- [Healing](http://192.241.195.202:9000/staging/scanner/linux/operations/concepts/healing.html)
- [Scanner](http://192.241.195.202:9000/staging/scanner/linux/operations/concepts/scanner.html)

### DIFF
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -40,7 +40,7 @@ Glossary
      MinIO combats bit rot with :term:`hashing` and :term:`erasure coding`.
 
    bit rot healing
-     Objects corrupted due to bit rot automatically restore to a healed state at time of read.
+     Objects corrupted due to bit rot automatically restore to a healed state at time of read during any ``GET`` operation.
      MinIO captures and heals corrupted objects on the fly with its :term:`hashing` implementation.
 
    bucket
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -223,7 +223,10 @@ Glossary
      
      - lifecycle management rules requiring object transition
      - bucket or site replication status
-     - object :term:`bit rot` and healing
+     - object :term:`bit rot` and :term:`healing`
+     - usage data
+
+     For more, see :ref:`minio-concepts-scanner`.
 
    self signed certificates
      A self-signed certificate is one created by, issued by, and signed by the company or developer responsible for the content the certificate secures.
--- a/source/glossary.rst
+++ b/source/glossary.rst
@@ -40,7 +40,7 @@ Glossary
      MinIO combats bit rot with :term:`hashing` and :term:`erasure coding`.
 
    bit rot healing
-     Objects corrupted due to bit rot automatically restore to a healed state at time of read during any ``GET`` operation.
+     Objects corrupted due to bit rot are automatically :term:`healed <healing>` during any ``GET`` or ``HEAD`` operation.
      MinIO captures and heals corrupted objects on the fly with its :term:`hashing` implementation.
 
    bucket
--- a/source/operations/concepts.rst
+++ b/source/operations/concepts.rst
@@ -186,12 +186,14 @@ Some common reasons for bit rot include:
 - accidental overwrites
 
 MinIO uses a hashing algorithm to confirm the integrity of an object.
-This algorithm automatically applies at the time of any ``GET`` request that calls an object.
+This algorithm automatically applies at the time of any ``GET`` and ``HEAD`` operations for an object.
+For objects in a versioned bucket, a ``PUT`` operation can also trigger healing if MinIO identifies version inconsistencies.
 If an object becomes corrupted by bit rot, MinIO can automatically :ref:`heal <minio-concepts-healing>` the object depending on the availability of parity shards for the object.
 
-While bit rot healing can be done during the routine checks of the :ref:`MinIO Scanner <minio-concepts-scanner>`, this function is off by default.
-Bit rot is a low risk, especially if an object has multiple shards spread across multiple disks.
-Due to the relatively high performance cost compared to a relatively low benefit, MinIO generally recommends not activating bit rot checks during scanning.
+MinIO can also perform bit rot checks and healing using the :ref:`MinIO Scanner <minio-concepts-scanner>`.
+However, scanner bit rot checking is **off** by default.
+Active bit rot healing during scanner has a high performance impact in comparison to the low probability of bit rot affecting multiple object shards distributed across multiple drives and nodes.
+The automatic checks during normal operations is generally sufficient for bit rot, and MinIO does not recommend using the scanner for this type of health check.
 
 MinIO Distributes Data Across :ref:`Erasure Sets <minio-ec-erasure-set>` for High Availability and Resiliency
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/operations/concepts.rst
+++ b/source/operations/concepts.rst
@@ -166,7 +166,7 @@ MinIO Uses :ref:`Erasure Coding <minio-erasure-coding>` for Data Redundancy and 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 MinIO Erasure Coding is a data redundancy and availability feature that allows MinIO deployments with multiple drives to automatically reconstruct objects on-the-fly despite the loss of multiple drives or nodes in the cluster. 
-Erasure Coding provides object-level healing with significantly less overhead than adjacent technologies such as RAID or replication.
+Erasure Coding provides object-level :ref:`healing <minio-concepts-healing>` with significantly less overhead than adjacent technologies such as RAID or replication.
 
 
 MinIO Implements Bit Rot Healing to Protect Data At Rest
@@ -187,7 +187,7 @@ Some common reasons for bit rot include:
 
 MinIO uses a hashing algorithm to confirm the integrity of an object.
 This algorithm automatically applies at the time of any ``GET`` request that calls an object.
-If an object becomes corrupted by bit rot, MinIO can automatically heal the object depending on the availability of parity shards for the object.
+If an object becomes corrupted by bit rot, MinIO can automatically :ref:`heal <minio-concepts-healing>` the object depending on the availability of parity shards for the object.
 
 While bit rot healing can be done during the routine checks of the :ref:`MinIO Scanner <minio-concepts-scanner>`, this function is off by default.
 Bit rot is a low risk, especially if an object has multiple shards spread across multiple disks.
--- a/source/operations/concepts.rst
+++ b/source/operations/concepts.rst
@@ -186,7 +186,12 @@ Some common reasons for bit rot include:
 - accidental overwrites
 
 MinIO uses a hashing algorithm to confirm the integrity of an object.
-If an object becomes corrupted by bit rot, MinIO can automatically heals the object depending on the system topology and availability.
+This algorithm automatically applies at the time of any ``GET`` request that calls an object.
+If an object becomes corrupted by bit rot, MinIO can automatically heal the object depending on the availability of parity shards for the object.
+
+While bit rot healing can be done during the routine checks of the :ref:`MinIO Scanner <minio-concepts-scanner>`, this function is off by default.
+Bit rot is a low risk, especially if an object has multiple shards spread across multiple disks.
+Due to the relatively high performance cost compared to a relatively low benefit, MinIO generally recommends not activating bit rot checks during scanning.
 
 MinIO Distributes Data Across :ref:`Erasure Sets <minio-ec-erasure-set>` for High Availability and Resiliency
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -203,7 +208,7 @@ MinIO calculates the size and number of Erasure Sets in a Server Pool based on t
 MinIO Automatically Heals Corrupt or Missing Data On-the-fly
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Healing is MinIO's ability to restore data after some event causes data loss.
+:ref:`Healing <minio-concepts-healing>` is MinIO's ability to restore data after some event causes data loss.
 Data loss can come from bit rot, drive loss, or node loss.
 
 :ref:`Erasure coding <minio-erasure-coding>` provides continued read and write access if an object has been partially lost.
--- a/source/operations/concepts/healing.rst
+++ b/source/operations/concepts/healing.rst
@@ -32,7 +32,8 @@ The ability of MinIO to restore a damaged object relates directly to the followi
   For example, an erasure set may have eight total drives and use three drives during a write for parity.
   In this scenario, MinIO splits an object into 5 data shards and create 3 parity shards.
   MinIO distributes these eight shards across the drives in the erasure set.
-  No one drive contains only parity shards or only data shards as a protection.
+  No one drive contains only parity shards or only data shards.
+  Instead, MinIO writes shards for each object in a randomized way to distribute reads evenly across drives.
 
   When MinIO needs to provide the object, it looks for the data shards for the object.
   If any of the data shards are missing or damaged, MinIO uses one or more of the parity shards to restore the object.
@@ -67,7 +68,7 @@ One of these tasks checks the integrity of objects and, if found damaged or corr
 
 On each scanning pass, MinIO uses a hash of the object name to select one out of every 1,024 objects to check.
 
-If any object is found to have lost data shards, MinIO heals the object from available parity shards.
+If any object is found to have lost shards, MinIO heals the object from available shards.
 By default, MinIO does *not* check for :term:`bit rot` corruption using the scanner.
 This can be an expensive operation to perform and the risk of bit rot across multiple disks is low.
 
--- a/source/operations/concepts/healing.rst
+++ b/source/operations/concepts/healing.rst
@@ -14,7 +14,10 @@ What is healing?
 ----------------
 
 Healing is MinIO's ability to restore an object that has been damaged, corrupted, or partially lost.
-The loss can come from random corruptions in the form of :ref:`bitrot <minio-ec-bitrot>` or the loss of one or more drives due to failure or outages.
+The loss can come from random corruptions in the form of 
+- :ref:`bit rot <minio-ec-bitrot>`
+- the loss of one or more drives due to failure or outages
+- the corruption or loss of one or more shards of an object for other reasons
 
 Healing and Erasure Coding
 --------------------------
@@ -24,17 +27,17 @@ The ability of MinIO to restore a damaged object relates directly to the followi
 - total number of drives in the :ref:`erasure set <minio-erasure-coding>` where the object exists
 - :ref:`parity setting <minio-ec-parity>` for the erasure set
 
-  Parity refers to the number of extra shards MinIO create when writing the object to recover an object.
+  Parity refers to the number of extra shards MinIO create when writing the object.
   For example, an erasure set may have eight total drives and use three drives during a write for parity.
   In that scenario, MinIO splits the object into five parts written to five of the drives.
-  The remaining three drives in the erasure set are used for parity shards.
+  MinIO uses the remaining three drives in the erasure set for parity shards.
 
   When MinIO needs to provide the object, it looks for the data shards for the object.
-  If any of the data shards are missing or damaged, MinIO uses the parity shards to successfully restore the object.
+  If any of the data shards are missing or damaged, MinIO uses the parity shards to restore the object.
   For this scenario, up to three of data shard parts can be lost or damaged and MinIO can still successfully restore and serve the object. 
 - number of drives available with intact parts of the object
 
-  The number of drives available with intact parts of the object must meet or exceed the number of drives used for data shards in the erasure set.
+  The number of drives available with intact data or parity shards of the object must meet or exceed the number of drives used for data shards in the erasure set.
   In the scenario above, five drives with intact shards must be online and available for MinIO to successfully serve the object.
 
 When does MinIO heal an object?
@@ -42,33 +45,42 @@ When does MinIO heal an object?
 
 MinIO has a robust system for healing objects.
 
-Healing during ``GET`` request
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Healing during ``GET`` requests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 MinIO automatically checks the consistency of an object's data shards each time you request an object with a ``GET`` operation.
 
 If all data shards are found intact, MinIO takes no further action on the object and serves the object from the data shards.
 
 If MinIO finds missing or damaged data shards, MinIO utilizes the available parity shards to heal the object before serving it as part of the ``GET`` operation.
-There must be an available parity shard available for each lost or damaged data shard.
+There **must** be a parity shard available for each lost or damaged data shard.
 
 Healing with the object scanner
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 MinIO utilizes an object scanner to perform a number of tasks related to objects.
-One of the tasks is to check the integrity of objects and, if found damaged or corrupted, heal them.
+One of these tasks checks the integrity of objects and, if found damaged or corrupted, heals them.
 
-On each scanning pass, MinIO one out of every 1,024 objects to check.
-If any object is found to have corrupted or lost data shards, MinIO heals the object from available parity shards.
+On each scanning pass, MinIO selects one out of every 1,024 objects to check.
+MinIO chooses the objects based on a hash of the object name.
+
+If any object is found to have lost data shards, MinIO heals the object from available parity shards.
+By default, MinIO does *not* check for :term:`bit rot` corruption using the scanner, as it can be an expensive operation to perform while bit rot is a low risk across multiple disks.
 
 Healing by manual request
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
+Administrators can use :mc:`mc admin heal` to initiate a full system healing.
+The procedure is very resource intensive and not typically needed.
 
+Consult with MinIO Engineers before manually starting a healing process on a deployment.
 
 Healing metrics
 ---------------
 
+MinIO provides several `healing metrics <https://min.io/docs/minio/linux/operations/monitoring/metrics-and-alerts.html#healing-metrics>`__ to monitor the status of healing processes on a deployment.
+
+Refer to the :ref:`minio-metrics-and-alerts` for more information on available endpoints and configuration.
 
 .. _minio-concepts-healing-colors:
 
--- a/source/operations/concepts/healing.rst
+++ b/source/operations/concepts/healing.rst
@@ -15,6 +15,7 @@ What is healing?
 
 Healing is MinIO's ability to restore an object that has been damaged, corrupted, or partially lost.
 The loss can come from multiple types of corruptions or loss, such as but not limited to:
+
 - drive-level errors or failure
 - OS or filesystem errors or failure
 - :ref:`bit rot <minio-ec-bitrot>`
--- a/source/operations/concepts/healing.rst
+++ b/source/operations/concepts/healing.rst
@@ -14,10 +14,10 @@ What is healing?
 ----------------
 
 Healing is MinIO's ability to restore an object that has been damaged, corrupted, or partially lost.
-The loss can come from random corruptions in the form of 
+The loss can come from multiple types of corruptions or loss, such as but not limited to:
+- drive-level errors or failure
+- OS or filesystem errors or failure
 - :ref:`bit rot <minio-ec-bitrot>`
-- the loss of one or more drives due to failure or outages
-- the corruption or loss of one or more shards of an object for other reasons
 
 Healing and Erasure Coding
 --------------------------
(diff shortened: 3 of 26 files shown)


source/operations/concepts.rst
```diff
@@ -186,7 +186,12 @@ Some common reasons for bit rot include:
 - accidental overwrites
 
 MinIO uses a hashing algorithm to confirm the integrity of an object.
-If an object becomes corrupted by bit rot, MinIO can automatically heals the object depending on the system topology and availability.
+This algorithm automatically applies at the time of any ``GET`` request that calls an object.
+If an object becomes corrupted by bit rot, MinIO can automatically :ref:`heal <minio-concepts-healing>` the object depending on the availability of parity shards for the object.
+
+While bit rot healing can be done during the routine checks of the :ref:`MinIO Scanner <minio-concepts-scanner>`, this function is off by default.
```
ravindk89, line 192: SO I think we need to make this clearer.

We do a bitrot/health check on `GET` or `HEAD` (I think).

cc/ @harshavardhana @klauspost to confirm.

This clarity is important so we don't have users walking away thinking we never do bitrot checks - we just do not do them as part of the _scanner_, which itself _does_ do high level health checks.

source/operations/concepts.rst
```diff
@@ -186,7 +186,12 @@ Some common reasons for bit rot include:
...
+Bit rot is a low risk, especially if an object has multiple shards spread across multiple disks.
```
ravindk89, line 193: Hm. Somewhat subjective and begs the question of 'how many shards makes the risk low'.

I suggest dropping or pulling this back to

"Active bitrot healing during scanner has a high performance impact, in comparison to the low probability of bitrot affecting multiple object shards distributed across multiple drives and nodes.

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+   :depth: 1
+
+What is healing?
+----------------
+
+Healing is MinIO's ability to restore an object that has been damaged, corrupted, or partially lost.
+The loss can come from random corruptions in the form of 
+- :ref:`bit rot <minio-ec-bitrot>`
+- the loss of one or more drives due to failure or outages
+- the corruption or loss of one or more shards of an object for other reasons
```
ravindk89, line 20: I wonder if we should order these based on likelihood:

- Drive-level errors or failure
- OS/Filesystem errors or failure 
- :ref:`bit rot <minio-ec-bitrot>` of stored data

ravindk89: and use the 'including but not limited to' preface.

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+   :depth: 1
+
+Overview
+--------
+
+MinIO uses the built-in scanner to check objects for healing and to take any scheduled object actions.
+Such actions may include:
+
+- calculate data usage on drives
+- apply configured :ref:`lifecycle management <minio-lifecycle-management>` or :ref:`object retention <minio-object-retention>` rules
```
eco-minio, line 20: Evaluate and apply

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+- :ref:`bucket <minio-bucket-replication>` or :ref:`site <minio-site-replication-overview>` replication
```
eco-minio, line 21: Perform

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+- check objects for missing or corrupted data or parity shards and perform :ref:`healing <minio-concepts-healing>`
+
+The scanner performs these functions at two levels: cluster and bucket.
+At the cluster level, the scanner splits all buckets into sets and scans one set at a time.
```
eco-minio, line 25: Do we need to define set here?

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Healing and Erasure Coding
+--------------------------
+
+The ability of MinIO to restore a damaged object relates directly to the following:
+
+- total number of drives in the :ref:`erasure set <minio-erasure-coding>` where the object exists
+- :ref:`parity setting <minio-ec-parity>` for the erasure set
+
+  Parity refers to the number of extra shards MinIO create when writing the object.
```
ravindk89, line 30: I would not reference them as 'extra shards'

Parity refers to dedicated recovery shards MinIO creates when writing an object.
The number of parity shards created per object depends on the erasure parity configuration of the deployment.

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+The scanner starts with new buckets, then randomizes the scanning of other buckets.
```
eco-minio, line 26: Is this correct or is it correct to say it triggers automatically on new bucket creation

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+The scanner completes checks on all bucket sets before starting over with a new set of scans.
+
+At the bucket level, the scanner groups items in buckets and scans selected items from that bucket.
+The scanner selects objects for a scan based on a hash of the object name.
+Over a span of 16 scans, MinIO checks every object it knows about.
```
eco-minio, line 31:
```suggestion
Over a span of 16 scans, MinIO checks every object it in the namespace.
```

This is more correct, correct? ^

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Scan Length
+-----------
+
+Multiple factors impact the time it takes for a scan to complete.
+
+Some of these factors include:
+- Type of hard drives provided to MinIO
```
eco-minio, line 39:
```suggestion
- Type of  drives provided to MinIO
```

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+- Throughput available
```
eco-minio, line 40:
```suggestion
- Throughput and iops available
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+  For example, an erasure set may have eight total drives and use three drives during a write for parity.
+  In that scenario, MinIO splits the object into five parts written to five of the drives.
+  MinIO uses the remaining three drives in the erasure set for parity shards.
```
ravindk89, line 33: This *might* imply that we always use the same drives for parity or data. It's randomized, so maybe

```suggestion
For example, MinIO splits an object into 5 data and 3 parity shards and distributes those shards across 8 drives.
```

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+- Number and size of objects
+- Other activity on the MinIO Server
+
+For example, MinIO may pause the scanner to make I/O operations available for read and write requests.
```
eco-minio, line 44:
```suggestion
For example, by default MinIO will pause the scanner to make I/O operations available for read and write requests.
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+
+  When MinIO needs to provide the object, it looks for the data shards for the object.
```
ravindk89, line 35: We can use any combination of shards to assemble an object, up to the write quorum / data drives (5 shards).

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+  If any of the data shards are missing or damaged, MinIO uses the parity shards to restore the object.
```
ravindk89, line 36: This goes both ways - we can restore parity shards if missing.

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+This lengthens the time it takes for a scan to complete.
```
eco-minio, line 45:
```suggestion
This can lengthen the time it takes for a scan to complete.
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+
+  The number of drives available with intact data or parity shards of the object must meet or exceed the number of drives used for data shards in the erasure set.
+  In the scenario above, five drives with intact shards must be online and available for MinIO to successfully serve the object.
+
+When does MinIO heal an object?
+-------------------------------
+
+MinIO has a robust system for healing objects.
+
+Healing during ``GET`` requests
```
ravindk89, line 48: I think both GET and HEAD (possibly STAT?)

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO automatically checks the consistency of an object's data shards each time you request an object with a ``GET`` operation.
+
+If all data shards are found intact, MinIO takes no further action on the object and serves the object from the data shards.
```
ravindk89, line 53: We should be checking all shards, including parity

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+
+MinIO waits between each scan by a factor multiplication of the time it takes each scan operation to complete.
+By default, the value of this factor is ``10.0``, meaning MinIO waits 10x the length of an operation after one scan completes before starting the next scan.
+MinIO also applies a maximum wait time between operations, set to ``15s`` by default.
+
+Scanner Performance
+-------------------
+
+Scanner performance typically depends on the available node resources, the size of the cluster, the number of erasure sets compared to the number of drives, and the complexity of bucket hierarchy (objects and prefixes).
```
eco-minio, line 54:
```suggestion
Scanner performance is generally governed by the available node resources, the size of the cluster, the number of erasure sets compared to the number of drives, and the complexity of bucket hierarchy (objects and prefixes).
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+Healing with the object scanner
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO utilizes an :ref:`object scanner <minio-concepts-scanner>` to perform a number of tasks related to objects.
+One of these tasks checks the integrity of objects and, if found damaged or corrupted, heals them.
+
+On each scanning pass, MinIO selects one out of every 1,024 objects to check.
+MinIO chooses the objects based on a hash of the object name.
+
+If any object is found to have lost data shards, MinIO heals the object from available parity shards.
```
ravindk89, line 67: Same, we check for any missing shards

source/operations/data-recovery/recover-after-drive-failure.rst
```diff
@@ -12,7 +12,7 @@ Drive Failure Recovery
 
 MinIO supports hot-swapping failed drives with new healthy drives. MinIO detects
 and heals those drives without requiring any node or deployment-level restart.
-MinIO healing occurs only on the replaced drive(s) and does not typically impact
+:ref:`MinIO healing <minio-concepts-healing>` occurs only on the replaced drive(s) and does not typically impact
```
eco-minio, line 15:
```suggestion
:ref:`MinIO healing <minio-concepts-healing>` occurs only on the replaced drive(s) and in most cases has minimal or negligible  impact
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+If MinIO finds missing or damaged data shards, MinIO utilizes the available parity shards to heal the object before serving it as part of the ``GET`` operation.
+There **must** be a parity shard available for each lost or damaged data shard.
+
+Healing with the object scanner
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO utilizes an :ref:`object scanner <minio-concepts-scanner>` to perform a number of tasks related to objects.
+One of these tasks checks the integrity of objects and, if found damaged or corrupted, heals them.
+
+On each scanning pass, MinIO selects one out of every 1,024 objects to check.
```
ravindk89, line 64: This might imply we check 1 object per pass/cycle. Not sure how best to address that.

"MinIO checks a portion of the total namespace on each pass, using a filter ratio to determine which objects to check or skip."

source/operations/install-deploy-manage/multi-site-replication.rst
```diff
@@ -97,7 +97,7 @@ Any MinIO deployment in the site replication configuration can resynchronize dam
    Site replication operations retry up to three (3) times.
    
    MinIO dequeues replication operations that fail to replicate after three attempts.
-   The scanner can pick up those affected objects at a later time and requeue them for replication.
+   The :ref:`scanner <minio-concepts-scanner>` can pick up those affected objects at a later time and requeue them for replication.
```
eco-minio, line 100: can^does ?

ravindk89 (changes requested):
I think we need to tighten up the healing logic , but this is a good first pass

source/glossary.rst
```diff
@@ -40,7 +40,7 @@ Glossary
      MinIO combats bit rot with :term:`hashing` and :term:`erasure coding`.
 
    bit rot healing
-     Objects corrupted due to bit rot automatically restore to a healed state at time of read.
+     Objects corrupted due to bit rot automatically restore to a healed state at time of read during any ``GET`` operation.
```
feorlen, line 43: maybe?
```suggestion
     Objects corrupted due to bit rot are automatically healed the next time they are read by a ``GET`` operation.
```
could link to healing term as well

source/operations/concepts.rst
```diff
@@ -186,7 +186,12 @@ Some common reasons for bit rot include:
 - accidental overwrites
 
 MinIO uses a hashing algorithm to confirm the integrity of an object.
-If an object becomes corrupted by bit rot, MinIO can automatically heals the object depending on the system topology and availability.
+This algorithm automatically applies at the time of any ``GET`` request that calls an object.
+If an object becomes corrupted by bit rot, MinIO can automatically :ref:`heal <minio-concepts-healing>` the object depending on the availability of parity shards for the object.
```
feorlen, line 190: `can automatically heal` suggests this behavior is optional? Perhaps `automatically heals` would work.

feorlen: I am somewhat wondering about scanner healing vs `GET` read time healing. Are those the same thing, but done at different times?

source/operations/concepts/healing.rst
feorlen, line 30:
```suggestion
  Parity refers to the number of extra shards MinIO creates when writing the object.
```

source/operations/concepts/healing.rst
feorlen, line 33: My understanding is hazy, but maybe this is a thing?
```suggestion
  For example, an erasure set may have eight drives in total: five for data shards and three for parity shards.
  In that scenario, MinIO splits the object into five parts, writing each part to five different drives.
  MinIO uses the remaining three drives in the erasure set for parity shards.
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+
+Healing during ``GET`` requests
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO automatically checks the consistency of an object's data shards each time you request an object with a ``GET`` operation.
+
+If all data shards are found intact, MinIO takes no further action on the object and serves the object from the data shards.
+
+If MinIO finds missing or damaged data shards, MinIO utilizes the available parity shards to heal the object before serving it as part of the ``GET`` operation.
+There **must** be a parity shard available for each lost or damaged data shard.
```
feorlen, line 56: some more ideas that might work?
```suggestion
MinIO automatically checks the consistency of an object's data shards with each ``GET`` operation.

If all the data shards are intact, MinIO serves the object from the data shards without inspecting the corresponding parity shards

If the object has missing or damaged data shards, MinIO uses the available parity shards to heal the object before serving it as part of the ``GET`` operation.
There **must** be a parity shard available for each lost or damaged data shard, otherwise the object cannot be recovered.
```

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MinIO utilizes an :ref:`object scanner <minio-concepts-scanner>` to perform a number of tasks related to objects.
+One of these tasks checks the integrity of objects and, if found damaged or corrupted, heals them.
+
+On each scanning pass, MinIO selects one out of every 1,024 objects to check.
+MinIO chooses the objects based on a hash of the object name.
+
+If any object is found to have lost data shards, MinIO heals the object from available parity shards.
+By default, MinIO does *not* check for :term:`bit rot` corruption using the scanner, as it can be an expensive operation to perform while bit rot is a low risk across multiple disks.
```
feorlen, line 68: some ideas?
```suggestion
MinIO uses an :ref:`object scanner <minio-concepts-scanner>` to perform a number of tasks related to objects.
One of these tasks checks the integrity of objects and, if found damaged or corrupted, heals them.

On each scanning pass, MinIO uses a hash of the object name to select one out of every 1,024 objects to check.

If any object is found to have lost data shards, MinIO heals the object from available parity shards.
By default, MinIO does *not* check for :term:`bit rot` corruption using the scanner.
This can be an expensive operation to perform and the risk of bit rot across multiple disks is low.
```

Could even add something like `To enable scanner bit rot checking, see [ref]`

source/operations/concepts/healing.rst
djwfyi: It looks for and uses data shards first. Only if it cannot find the data shards does it go for parity shards. That's why I worded it this way.

source/operations/concepts/healing.rst
djwfyi: We don't do anything with parity shards if all data shards are intact.

source/operations/concepts/healing.rst
djwfyi: we don't, not if all data shards are intact by the scanner.

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+MinIO provides several `healing metrics <https://min.io/docs/minio/linux/operations/monitoring/metrics-and-alerts.html#healing-metrics>`__ to monitor the status of healing processes on a deployment.
+
+Refer to the :ref:`minio-metrics-and-alerts` for more information on available endpoints and configuration.
+
+.. _minio-concepts-healing-colors:
+
+Healing Output Color Key
+------------------------
+
+Originally, the healing mechanism output a table that used a Green-Yellow-Red-Gray color key to attempt to differentiate the status of objects in healing.
```
feorlen, line 90:
```suggestion
Earlier MinIO versions generated a table with green, yellow, red, and gray colors to indicate the healing status of individual objects.
```

Is there a specific version we could reference here? Does it even matter?

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,110 @@
...
+These colors have been deprecated in favor of more detailed :ref:`healing metrics available at the cluster level <minio-metrics-and-alerts-available-metrics>`.
+
+The following table describes the intent of each of the deprecated color keys.
```
feorlen, line 93: more ideas
```suggestion
The following table describes the meaning of the now deprecated status colors:
```

source/operations/concepts/scanner.rst
djwfyi: 🤔 
I may just remove `sets` here to avoid conflation with `erasure sets`. Since these are different things.

source/operations/concepts/scanner.rst
djwfyi: @klauspost?

source/operations/concepts/scanner.rst
djwfyi: Ooh, good catch. Thanks!

source/operations/concepts/scanner.rst
djwfyi: I have a thing agains `will` as almost always overpromising, but adjusting the phrasing to get it closer to your suggestion.

source/operations/concepts/scanner.rst
djwfyi: Converting to a list for better readability

source/operations/install-deploy-manage/multi-site-replication.rst
djwfyi: Sure.

source/operations/concepts.rst
djwfyi: Confirmed with Harsha and updated the wording here.

source/operations/concepts.rst
djwfyi: I'll take that wording. WIth `bit rot` instead of `bitrot`.

source/operations/concepts/healing.rst
djwfyi: Yes, good ideas on both counts

source/operations/concepts/healing.rst
djwfyi: Agree with this and updated.

source/operations/concepts/healing.rst
djwfyi: 🤔  Interesting thought. Taking your suggestion for the most part.

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,95 @@
...
+Scanner Metrics
+---------------
+
+MinIO provides a number of `metrics related to the scanner <https://min.io/docs/minio/linux/operations/monitoring/metrics-and-alerts.html#scanner-metrics>`__.
+
+The metrics updated by the scanner, including usage metrics, reflect the last completed scan. 
+``PUT`` or ``DELETE`` operations since the last scan do not update in the usage until the next scan of the affected bucket(s).
+
+Use ``mc admin scanner info`` to see the current status of the scanner and the time since the last full scan.
+This can help in understanding the metrics provided by the scanner operation.
```
feorlen, line 74: Maybe rearrange this a bit? In part to sidestep some technicalities around time series metrics that aren't, strictly speaking, "updated." (You get new values, at a new future point in time.)

```suggestion
Use ``mc admin scanner info`` to see the current status of the scanner and the time since the last full scan.
This can help in understanding the metrics provided by the scanner operation.

Scanner metrics, including usage metrics, reflect the last completed scan.
``PUT`` or ``DELETE`` operations since the last scan do not update in the usage until the next scan of the affected bucket(s).
```

source/operations/data-recovery/recover-after-site-failure.rst
```diff
@@ -132,7 +132,7 @@ Passive Bucket Replication Resynchronization
 
 As a passive process, bucket replication may not perform as quickly as desired for a site recovery scenario.
 
-Using bucket replication relies on the standard replication scanner queue, which does not take priority over other processes.
+Using bucket replication relies on the standard replication :ref:`scanner <minio-concepts-scanner>` queue, which does not take priority over other processes.
```
feorlen, line 135: I'm not sure what work "using" is doing here, perhaps drop?
```suggestion
Bucket replication relies on the standard replication :ref:`scanner <minio-concepts-scanner>` queue, which does not take priority over other processes.
```

source/reference/deprecated/mc-ilm-add.rst
```diff
@@ -122,7 +122,7 @@ Parameters
    For versioned buckets, the expiry rule applies only to the *current* object version. 
    Use the :mc-cmd:`~mc ilm add --noncurrent-expire-days` option to apply expiration behavior to noncurrent object versions.
 
-   MinIO uses a scanner process to check objects against all configured lifecycle management rules. 
+   MinIO uses a :ref:`scanner process <minio-concepts-scanner>` to check objects against all configured lifecycle management rules.
```
feorlen, line 125: Ok it is indeed `all` but `the` reads more smoothly for me here? Use if you like.
```suggestion
   MinIO uses a :ref:`scanner process <minio-concepts-scanner>` to check objects against the configured lifecycle management rules. 
```

feorlen: oh and there's a bunch of these

source/operations/concepts/healing.rst
djwfyi: Clarified

source/operations/concepts/healing.rst
djwfyi: Updated with the info from Harsha.

source/operations/concepts/healing.rst
djwfyi: Clarified the wording, but what I wrote is correct for how it works.
MinIO looks for the data shards and serves the object if all data shards are intact.
Only if any of the data shards are missing or damaged does it do anything with parity shards.

source/operations/concepts/healing.rst
djwfyi: Clarifying how this works internally to confirm my understanding.

source/operations/concepts/healing.rst
djwfyi: Tweaked your suggestion, but running with it.

source/operations/concepts.rst
djwfyi: The healing itself is similar. Bit rot healing is on that specific object.
The scanner checks a selection of objects with each pass, but doesn't check for bit rot unless that is specifically enabled (with a significant performance hit as a result).

source/operations/concepts/healing.rst
djwfyi: Ravind had also made a suggestion on this paragraph. I've reworked it.

source/operations/concepts/healing.rst
djwfyi: Taking these. Thanks.

source/operations/concepts/healing.rst
djwfyi: 👍

source/reference/deprecated/mc-ilm-add.rst
djwfyi: I'm not for or against this change. I'll take it, though.

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,114 @@
...
+
+- total number of drives in the :ref:`erasure set <minio-erasure-coding>` where the object exists
+- number of drives available with intact parts of the object
+- :ref:`parity setting <minio-ec-parity>` for the erasure set
+
+  :term:`Parity` refers to the number of dedicated recovery shards MinIO creates when writing the object.
+  For example, an erasure set may have eight total drives and use three drives during a write for parity.
+  In this scenario, MinIO splits an object into 5 data shards and create 3 parity shards.
+  MinIO distributes these eight shards across the drives in the erasure set.
+  No one drive contains only parity shards or only data shards as a protection.
```
klauspost, line 35:
```suggestion
  No one drive contains only parity shards or only data shards.
  Instead it is randomized per object to distribute reads evenly across disks.
```

source/operations/concepts/healing.rst
klauspost: I don't understand what the disagreement is about.

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,114 @@
...
+ 
+   * - **Yellow** 
+     - *Healing*, the object is still in the process of healing, and there are sufficient data or parity shards available to complete the healing
+
+   * - **Red** 
+     - *Unhealthy*, the object has lost one or more shards and requires healing
+
+   * - **Grey** 
+     -  *Unrecoverable*, the object has lost too many data and/or parity shards and cannot be healed or recovered
+
```
klauspost, line 114: Does all of this matter if it is deprecated?

source/operations/concepts/scanner.rst
klauspost: I don't understand what Eco is implying. The text seems correct.

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,102 @@
...
+
+The scanner performs these functions at two levels: cluster and bucket.
+At the cluster level, the scanner splits all buckets into groups and scans one group of buckets at a time.
+The scanner starts with new buckets, then randomizes the scanning of other buckets.
+The scanner completes checks on all bucket groups before starting over with a new set of scans.
+
+At the bucket level, the scanner groups items in buckets and scans selected items from that bucket.
+The scanner selects objects for a scan based on a hash of the object name.
+Over a span of 16 scans, MinIO checks every object in the namespace.
+
```
klauspost, line 32:
```suggestion
Over a span of 16 scans, MinIO checks every object in the namespace.
Any prefixes that are known to be new since the last scan will always be fully scanned.

```

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,102 @@
...
+- Throughput and :abbr:`iops <input/output operations per second>` available
+- Number and size of objects
+- Other activity on the MinIO Server
+
+For example, by default, MinIO pauses the scanner to make I/O operations available for read and write requests.
+This can lengthen the time it takes for a scan to complete.
+
+MinIO waits between each scan by a factor multiplication of the time it takes each scan operation to complete.
+By default, the value of this factor is ``10.0``, meaning MinIO waits 10x the length of an operation after one scan completes before starting the next scan.
+MinIO also applies a maximum wait time between operations, set to ``15s`` by default.
```
klauspost, line 49: Remove this. We use the `speed=slowest/slow/default/fast/fastest` which contains these values as presets.

source/operations/concepts/healing.rst
ravindk89: @klauspost my vague understanding is that we check all shards, data or parity, on scanner hit to ensure that the object is healthy. The above might imply that we do not check for parity shard health.

source/operations/concepts/healing.rst
ravindk89: I think it comes up often enough in support cases, and we have some UI/UX in Console/Enterprise Console that also uses color coding.

source/operations/concepts/scanner.rst
ravindk89: I think it's the nuance of the scanner hitting 'a bucket that is new' vs 'a bucket that was *just* created'

The timing of the former might be seconds, hours, days, etc. after creation depending on scanner load, while the latter is a trigger based on an event and should be more instant.

source/operations/concepts/healing.rst
klauspost: Ah. I see. Yeah ` lost data shards` -> ` lost shards` would be more accurate.

source/operations/concepts/scanner.rst
djwfyi: Clarified. Maybe even correctly.

source/operations/concepts/healing.rst
djwfyi: Changed!

source/operations/concepts/healing.rst
djwfyi: Mildly tweaked from the suggestion, but updated.

source/operations/concepts/healing.rst
djwfyi: As @ravindk89 says, this is mainly for those times when someone comes across it in an older version or in the Console and wants to know what the colors mean. This still happens.

source/operations/concepts/healing.rst
```diff
@@ -0,0 +1,115 @@
...
+.. contents:: Table of Contents
+   :local:
+   :depth: 1
+
+What is healing?
+----------------
+
+Healing is MinIO's ability to restore an object that has been damaged, corrupted, or partially lost.
+The loss can come from multiple types of corruptions or loss, such as but not limited to:
```
feorlen, line 17: Blank line to make list below work correctly
```suggestion
The loss can come from multiple types of corruptions or loss, such as but not limited to:

```

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,103 @@
...
+The scanner selects objects for a scan based on a hash of the object name.
+Over a span of 16 scans, MinIO checks every object in the namespace.
+MinIO fully scans any prefixes known to be new since the last scan.
+
+Scan Length
+-----------
+
+Multiple factors impact the time it takes for a scan to complete.
+
+Some of these factors include:
```
feorlen, line 39:
```suggestion
Some of these factors include:

```

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,103 @@
...
+- Type of drives provided to MinIO
+- Throughput and :abbr:`iops <input/output operations per second>` available
```
feorlen, line 41: This shows in the page with the angle brackets, is that intended?

source/operations/concepts/scanner.rst
```diff
@@ -0,0 +1,103 @@
...
+
+MinIO waits between each scan by a factor multiplication of the time it takes each scan operation to complete.
+By default, the value of this factor is ``10.0``, meaning MinIO waits 10x the length of an operation after one scan completes before starting the next scan.
+The value of this factor changes depending on the configured :ref:`scanner speed setting <minio-scanner-speed-options>`. 
+
+Scanner Performance
+-------------------
+
+Many factors impact the scanner performance.
+Some of these factors include:
```
feorlen, line 56:
```suggestion
Some of these factors include:

```

source/operations/concepts/scanner.rst
djwfyi: supposed to be `( )` instead of `< >`

feorlen (approved):
Noted a couple typos, otherwise LGTM
🎉 This is a ton of work, the results look great 🎉

ravindk89 (approved):
LGTM mod Andrea's edits

PR #1198, opened 2024-04-26
Replace mc admin console references with mc admin logs

furkansahin:
mc admin console command is depreciated, users should start using mc admin logs instead.


djwfyi:
Thanks for the PR!
We need to do a few other changes at the same time to note the deprecation. Will hold this while those other changes get made.

djwfyi (approved):
Thanks for this PR and catching this.

PR #1199, opened 2024-04-29
Note that mc support perf suspends other S3 API calls for testing

feorlen:
Add information about queuing requests for `drive` and `net` options.

See https://github.com/minio/minio/pull/19634

Staged:
http://192.241.195.202:9000/staging/mc-support-perf-s3-api/linux/reference/minio-mc/mc-support-perf.html#description

### DIFF
--- a/source/reference/minio-mc/mc-support-perf.rst
+++ b/source/reference/minio-mc/mc-support-perf.rst
@@ -28,8 +28,16 @@ Use the :mc:`mc support perf` command to review the performance of the S3 API (r
 .. end-mc-support-perf-desc
 
 The resulting tests can provide general guidance of deployment performance under S3 ``GET`` and ``PUT`` requests and identify any potential bottlenecks.
+
+.. admonition:: Other S3 API calls suspended during testing
+   :class: note
+
+   For S3 testing, :mc:`mc support perf` temporarily suspends S3 API calls during a test.
+   They are automatically restarted after the test is complete or if the command is cancelled.
+
 For more complete performance testing, consider using a combination of load-testing using your staging application environments and the MinIO `WARP <https://github.com/minio/warp>`_ S3 benchmarking tool.
-   
+
+
 :mc:`mc support perf` has the following subcommands
 
 #. :mc-cmd:`~mc support perf drive`
--- a/source/reference/minio-mc/mc-support-perf.rst
+++ b/source/reference/minio-mc/mc-support-perf.rst
@@ -29,12 +29,6 @@ Use the :mc:`mc support perf` command to review the performance of the S3 API (r
 
 The resulting tests can provide general guidance of deployment performance under S3 ``GET`` and ``PUT`` requests and identify any potential bottlenecks.
 
-.. admonition:: Other S3 API calls suspended during testing
-   :class: note
-
-   For S3 testing, :mc:`mc support perf` temporarily suspends S3 API calls during a test.
-   They are automatically restarted after the test is complete or if the command is cancelled.
-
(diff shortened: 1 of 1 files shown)


source/reference/minio-mc/mc-support-perf.rst
```diff
@@ -28,8 +28,16 @@ Use the :mc:`mc support perf` command to review the performance of the S3 API (r
 .. end-mc-support-perf-desc
 
 The resulting tests can provide general guidance of deployment performance under S3 ``GET`` and ``PUT`` requests and identify any potential bottlenecks.
+
+.. admonition:: Other S3 API calls suspended during testing
+   :class: note
+
+   For S3 testing, :mc:`mc support perf` temporarily suspends S3 API calls during a test.
+   They are automatically restarted after the test is complete or if the command is cancelled.
+
```
feorlen, line 37: @klauspost which tests do this? `mc support perf client`? What exactly happens during the test? Requests rejected?

feorlen: Also want to understand what "freeze" means. Maybe "suspend" isn't the right word here either. (And the other command is named `unfreeze`.) It would be better to give some additional context.

feorlen: I think this also applies to `mc support perf net`

klauspost: Requests are held back in and queued.

source/reference/minio-mc/mc-support-perf.rst
```diff
@@ -29,13 +29,16 @@ Use the :mc:`mc support perf` command to review the performance of the S3 API (r
...
 :mc:`mc support perf` has the following subcommands
 
 #. :mc-cmd:`~mc support perf drive`
 
    Measure the speed of drives in a MinIO deployment.
 
+   :mc-cmd:`mc support perf drive` temporarily suspends S3 API calls during the test.
+   Incoming requests are queued until the test is complete or the command is cancelled.
+
```
feorlen, line 41: 🤔  I debated making these admonitions, but they sure take up a lot of space in the page.

feorlen:
@klauspost moved the revised text to the individual `drive` and `net` items. Any more that needs to be said?

source/reference/minio-mc/mc-support-perf.rst
```diff
@@ -44,6 +47,9 @@ For more complete performance testing, consider using a combination of load-test
 
    Measure the network throughput of all nodes.
 
+   :mc-cmd:`mc support perf net` temporarily suspends S3 API calls during the test.
+   Incoming requests are queued until the test is complete or the command is cancelled.
```
djwfyi, line 51: Alternative for you to consider to avoid all of the passives.
If you use it here, use it in both places.

```suggestion
   Incoming requests are held in a queue while the command runs.
   When the command completes or ends, queued requests process as normal.
```

source/reference/minio-mc/mc-support-perf.rst
djwfyi: I think that's the right call. Admonitions can be obnoxious.

djwfyi (approved):
Alternative wording to consider. Mostly the three passives in the same sentence gave me pause. But up to you to use or not.

No admonitions is the Right Call™.

source/reference/minio-mc/mc-support-perf.rst
feorlen: Derp 🤦🏻

feorlen: used variant of your suggestion 💯

djwfyi (approved):
💯  LGTM

klauspost (approved):

PR #1205, opened 2024-05-07
Fix white-space behaviour of inline code blocks

rushenn:


rushenn:
@Ravind Apologise for the late reply, somehow I missed the notification for the Slack thread. 

I've added a potential solution for extremely long code blocks within table cells. Please let me know if this resolves the issue for you

ravindk89:
This is definitely _better_.

Granted for the very long metric names the break looks ugly, but I'm not sure we can do much about that unless we go through and _manually_ break the metrics, which is not something I want to deal with right now.

@djwfyi @feorlen can you confirm this builds/functions as expected on the OSX side of things?

ravindk89 (approved):

ravindk89:
Once one other docs team ensures this builds w/o issues on another arch, we're good to roll this out

feorlen:
Confirm macOS build is better 👍🏻

PR #1206, opened 2024-05-07
FIXUP: Existing object replication is on by default

ravindk89:
# Summary

We some time ago updated bucket replication to perform existing object replication by default. Not entirely sure when, but long enough that we can just make the change.

Also fixed one out-of-band syntax error.

# Staged

http://192.241.195.202:9000/staging/REPLFIX/linux/administration/bucket-replication.html#replication-of-existing-objects

### DIFF
--- a/source/administration/bucket-replication.rst
+++ b/source/administration/bucket-replication.rst
@@ -176,29 +176,21 @@ application of object expiration.
 Replication of Existing Objects
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-MinIO by default replicates existing objects in the source bucket to the configured remote. 
-Existing object replication marks all objects or object prefixes that satisfy the replication rules as eligible for synchronization to the source cluster, *even if* those objects were created prior to configuring or enabling replication. 
+MinIO by default replicates existing objects in the source bucket to the configured remote, similar to `AWS: Replicating existing objects between S3 buckets <https://aws.amazon.com/blogs/storage/replicating-existing-objects-between-s3-buckets/>`__ without the overhead of contacting technical support.
 
-You can disable existing object replication while configuring or modifying a replication rule.
+MinIO marks all objects or object prefixes that satisfy the replication rules as eligible for synchronization to the remote cluster and bucket.
+MinIO only excludes those objects without a version ID, such as those objects written before enabling versioning on the bucket.
+
+You can disable existing object replication while configuring or modifying the bucket replication rule.
 You must specify *all* desired replication features during creation or modification:
 
 - For new replication rules, exclude ``"existing-objects"`` from the list of replication features specified to :mc-cmd:`mc replicate add --replicate`.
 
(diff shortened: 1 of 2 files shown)


source/administration/bucket-replication.rst
```diff
@@ -176,45 +176,29 @@ application of object expiration.
...
-:ref:`replication scanner and queue system <minio-replication-process>` for
-detecting and synchronizing objects regardless of the enabled replication
-feature. The time required to fully synchronize a bucket depends on a number of
-factors, including but not limited to the current cluster replication load,
-overall cluster load, and the size of the namespace (all objects in the bucket).
+MinIO by default replicates existing objects in the source bucket to the configured remote. 
+Existing object replication marks all objects or object prefixes that satisfy the replication rules as eligible for synchronization to the source cluster, *even if* those objects were created prior to configuring or enabling replication. 
+
+You can disable existing object replication while configuring or modifying a replication rule.
+You must specify *all* desired replication features during creation or modification:
```
feorlen, line 183: Maybe also explicitly say that an updated configuration _replaces_ the previous? Just to make it as clear as possible that you can't incrementally edit a list of rules.

source/administration/bucket-replication.rst
```diff
@@ -176,45 +176,29 @@ application of object expiration.
...
-
-Enabling existing object replication does not increase the priority of objects
-pending replication. MinIO uses the same core 
-:ref:`replication scanner and queue system <minio-replication-process>` for
-detecting and synchronizing objects regardless of the enabled replication
-feature. The time required to fully synchronize a bucket depends on a number of
-factors, including but not limited to the current cluster replication load,
-overall cluster load, and the size of the namespace (all objects in the bucket).
+MinIO by default replicates existing objects in the source bucket to the configured remote. 
+Existing object replication marks all objects or object prefixes that satisfy the replication rules as eligible for synchronization to the source cluster, *even if* those objects were created prior to configuring or enabling replication.
```
feorlen, line 180: I keep thinking this should be two sentences, but not coming up with an alternative I like better. Something to contemplate? 🤔

djwfyi (approved):
LGTM

feorlen (approved):
LGTM + ideas if you feel like it

ravindk89:
@feorlen lemme chew on those, and if I can't think of anything before EOD I'll fire this out and add this to the bucket list

source/administration/bucket-replication.rst
djwfyi:
```suggestion
Default object replication marks all objects or object prefixes that satisfy the replication rule as eligible for synchronization to the target cluster.
This means that objects created prior to configuring or enabling bucket replication are marked for synchronization to the target cluster by default bucket replication configurations.
```

djwfyi: Also, we say "source cluster" when I think we mean "target cluster".

feorlen (approved):
👍🏻

Issue #1159, opened 2024-03-15
[RELEASE] Server RELEASE.2024-03-15T01-07-19Z


djwfyi:
Server [RELEASE.2024-03-15T01-07-19Z](https://github.com/minio/minio/releases/tag/RELEASE.2024-03-15T01-07-19Z) has changes that impact the docs:

- [ ] New v3 of metrics system combines metrics in smaller, different chunks; check how metrics are imported and update as needed [PR #19068](https://github.com/minio/minio/pull/19068)
- [ ] Changes how compression variables are handled, now accepting `*` as an valid setting [PR #19234](https://github.com/minio/minio/pull/19234)
- [ ] Allows configuration of the max objects per version limit [PR #19265](https://github.com/minio/minio/pull/19265)
- [ ] New "optimized" envvar for configuring parity settings when drives go offline; sets to maximize space rather than reducing data drives to retain parity [PR #19260](https://github.com/minio/minio/pull/19260)

ravindk89:
Note: Metrics V3 is being actively reworked, we can leave that off for now.

PR #1211, opened 2024-05-08
DOCS-1159: RELEASE.2024-03-15T01-07-19Z

ravindk89:
Partially addresses #1159 

Excluding Metrics v3 since that work is ongoing, and it's not fully baked as of yet.

All changes are effectively config level.

Also fixed up some format/hierarchy issues while I was at it.

Staged:

- http://192.241.195.202:9000/staging/DOCS-1159/linux/reference/minio-server/settings/core.html
- http://192.241.195.202:9000/staging/DOCS-1159/linux/reference/minio-server/settings/core.html#compression-extensions
- http://192.241.195.202:9000/staging/DOCS-1159/linux/reference/minio-server/settings/core.html#maximum-object-versions
- http://192.241.195.202:9000/staging/DOCS-1159/linux/reference/minio-server/settings/storage-class.html#parity-retention-optimization

### DIFF
--- a/source/reference/minio-server/settings/core.rst
+++ b/source/reference/minio-server/settings/core.rst
@@ -20,8 +20,8 @@ This page covers settings that control core behavior of the MinIO process.
    :start-after: start-minio-settings-test-before-prod
    :end-before: end-minio-settings-test-before-prod
 
-MinIO Options
-~~~~~~~~~~~~~
+MinIO Server CLI Options
+------------------------
 
 .. tab-set::
 
@@ -47,11 +47,8 @@ For example, to set up ftp access, you could set the variable to something like 
 
 On Unix-like systems, you can save a file with the environment variable to ``/etc/defaults/minio`` instead of setting the variable manually.
 
-Common Settings
+Storage Volumes
 ---------------
-
-Volumes
-~~~~~~~
 
 .. tab-set::
 
@@ -71,7 +68,7 @@ Volumes
          :end-before: end-minio-settings-no-config-option
 
 Environment Variable File Path
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+------------------------------
 
 .. tab-set::
 
@@ -90,7 +87,7 @@ Environment Variable File Path
          :end-before: end-minio-settings-no-config-option
       
 Workers for Expiration
-~~~~~~~~~~~~~~~~~~~~~~
+----------------------
 
 .. tab-set::
 
@@ -110,7 +107,7 @@ Workers for Expiration
          :end-before: end-minio-settings-no-config-option
 
 Domain
-~~~~~~
+------
 
 .. tab-set::
 
@@ -145,7 +142,7 @@ Domain
 .. _minio-scanner-speed-options:
 
 Scanner Speed
-~~~~~~~~~~~~~
+-------------
 
 .. tab-set::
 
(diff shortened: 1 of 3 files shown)


source/reference/minio-server/settings/core.rst
```diff
@@ -357,5 +353,30 @@ The selected stripe size is **immutable** after the cluster has been initialized
...
+         :delimiter: " "
+
+*Optional*
+
+Overrides the default maximum version per object limit of ``10000`` with the user specified value.
+
+.. important::
+
+   The default limit of 10,000 provides a safety valve against incorrect or inefficient application behavior in versioned buckets.
+   Lifting this limit without first ensuring application's are designed for versioned operations may result in a negative performance impact over time.
```
feorlen, line 379:
```suggestion
   Lifting this limit without first ensuring applications are designed for versioned operations may result in a negative performance impact over time.
```

also perhaps `your applications` even?

source/reference/minio-server/settings/storage-class.rst
```diff
@@ -23,80 +34,140 @@ This impacts how MinIO uses the space on the drive(s) and how MinIO can recover
...
-   
-   The maximum value depends on the erasure set size of the initial server pool in the deployment, where the upper bound is  :math:`\frac{\text{ERASURE_SET_SIZE}}{\text{2}}`.
-   For example, a deployment with erasure set stripe size of 16 has a maximum standard parity of 8.
+      .. mc-conf:: storage_class optimize
+         :delimiter: " "
 
-   You can change this value after startup to any value between ``0`` and the upper bound for the erasure set size.
-   MinIO only applies the changed parity to newly written objects.
-   Existing objects retain the parity value in place at the time of their creation.
+MinIO by default automatically "upgrades" parity for an object if the destination erasure set maintains write quorum *but* has one or more drives offline.
```
djwfyi, line 150:
```suggestion
MinIO by default automatically "upgrades" parity for an object if the destination erasure set maintains write quorum *but* has one or more drives offline.
The object written on the destination maintains the same number of data shards, but a reduced number of parity shards compared to the original.
```

I think we need an example to clarify how MinIO "upgrades" parity.
I might have gotten it backwards in my suggestion.

source/reference/minio-server/settings/core.rst
```diff
@@ -323,8 +300,27 @@ Defaults to ``"text/*, application/json, application/xml, binary/octet-stream"``
...
 
+Comments
+~~~~~~~~
+
+.. tab-set::
+
+   .. tab-item:: Environment Variable
+
+      This setting does not have an environment variable option.
+      Use the configuration variable instead.
```
djwfyi, line 311:
```suggestion
      Use the configuration setting instead.
```

source/reference/minio-server/settings/core.rst
```diff
@@ -357,5 +353,30 @@ The selected stripe size is **immutable** after the cluster has been initialized
...
+-----------------------
+
+.. tab-set::
+
+   .. tab-item:: Environment Variable
+      :sync: envvar
+
+      .. envvar:: MINIO_API_OBJECT_MAX_VERSIONS
+
+   .. tab-item:: Configuration Variable
```
djwfyi, line 366:
```suggestion
   .. tab-item:: Configuration Setting
```

feorlen (approved):
LGTM mod typo

djwfyi (approved):
Couple of suggestions for you.
Otherwise, looks good to me.

source/reference/minio-server/settings/storage-class.rst
ravindk89: I think so as well - @harshavardhana or perhaps @klauspost can confirm but:

in an erasure set of 16 drives and EC:4, if 1 drive goes online, we write with EC:5.

so the new object is 11:5 while other objects are 12:4. Since a drive is offline, we get  11 data blocks and 4 parity blocks, but the object is maintaining the 'quorum' of all other objects.

If you were to set this for capacity mode, we would write 12:4 and just leave off the last parity block, such that this object has reduced quorum (just like all the other objects on that set). 

AFAIK healing does not correct parity upgrades, though it would 'fix' the object written in capacity mode.

ravindk89: https://github.com/minio/minio/blob/f5e3eedf3458c9f5b580edac643292213c1322e7/cmd/erasure-object.go#L1293-L1320

Also implies we upgrade parity for each 'down' drive up to quorum.

I assume the issue is that the smaller number of data shards means those shards are larger, therefore there are some drives that end up hot-spotted if the erasure set operates in it's degraded state for a long time.

ravindk89: @djwfyi i might fire this off now and return to parity upgrade behavior in a dedicated PR, since it's likely nuanced.

Issue #1207, opened 2024-05-07
[IMPROVEMENT] Transition target object lock settings do not have to match the source


ravindk89:
https://min.io/docs/minio/linux/administration/object-management/object-lifecycle-management.html

A few things:

- We do *not* require the remote bucket implement Object Locking, even if the source has it
- We do *not* require the remote bucket to implement Versioning (and in fact that might be an antipattern)

We can likely add those as a single H2 and call this done.

That said we might also explore a light re-arch here as a stretch goal:

```
Object Management
-> Object Tiering
    -> Configure Hot/Cold Tiering
    -> Use GCS as Warm/Cold Tier
    -> Use Azure as Warm/Cold Tier
    -> Use S3 as Warm/Cold Tier
-> Object Expiration
    -> Expire Objects in Unversioned Buckets
    -> Expire Objects in Versioned Buckets

```

Or something - would slightly improve the findability of data. Discuss first.

PR #1220, opened 2024-05-21
Adds information about tiering rules and exclusive access

djwfyi:
Adds info to the object management page.
Does not restructure the docs at this time.

Partially addresses #1207

Staged: 
- http://192.241.195.202:9000/staging/target-lock/linux/administration/object-management.html#target-bucket-configuration-for-tiering

### DIFF
--- a/source/administration/object-management.rst
+++ b/source/administration/object-management.rst
@@ -220,13 +220,18 @@ Conversion to or from XML may be required for importing rules created on S3 or s
 
 See :ref:`minio-lifecycle-management` for more complete documentation.
 
-Target Bucket Configuration for Tiering
----------------------------------------
+Target Bucket Considerations
+----------------------------
 
 Keep in mind the following considerations when configuring the target bucket for tiering:
 
-- The target bucket *can* have its own set of object management rules different from the source.
+- The target bucket *can* have its own set of object management rules.
+  Tiering does *not* require the source and target buckets have the same rules.
+
   For example, the source bucket may have object locking defined while the target bucket does not.
+  
+  Target buckets should *not* have their own rules for expiration or additional tiering.
+  MinIO supports only one level of tiering.
 - While enabling tiering requires :ref:`versioning <minio-bucket-versioning>` on the source bucket, the target bucket does *not* require versioning.
   Enabling versioning on both the source and target buckets may lead to unexpected results.
 
--- a/source/administration/object-management.rst
+++ b/source/administration/object-management.rst
@@ -220,6 +220,33 @@ Conversion to or from XML may be required for importing rules created on S3 or s
 
 See :ref:`minio-lifecycle-management` for more complete documentation.
 
+Target Bucket Configuration for Tiering
+---------------------------------------
+
+Keep in mind the following considerations when configuring the target bucket for tiering:
+
+- The target bucket *can* have its own set of object management rules different from the source.
+  For example, the source bucket may have object locking defined while the target bucket does not.
+- While enabling tiering requires :ref:`versioning <minio-bucket-versioning>` on the source bucket, the target bucket does *not* require versioning.
+  Enabling versioning on both the source and target buckets may lead to unexpected results.
+
+Exclusive Access to Remote Data
+-------------------------------
+
+MinIO **must** have *exclusive* access to the target bucket.
+No other user, process, application, or resource should have any access to or perform any actions against the target bucket.
+
+All access to the transitioned objects *must* occur through MinIO via S3 API operations only. 
+Manually modifying a transitioned object - whether the metadata on the “hot” MinIO tier or the object data on the remote “warm/cold” tier - may result in loss of that object data.
+
+MinIO ignores any objects in the remote bucket or bucket prefix not explicitly managed by the MinIO deployment. Automatic transition and transparent object retrieval depend on the following assumptions:
+
+- No external mutation, migration, or deletion of objects on the remote storage.
+- No lifecycle management rules (such as transition or expiration) on the remote storage bucket.
+
+To facilitate this exclusive access, grant the lifecycle management user ``read``, ``write``, and ``delete`` access to the target bucket in its :ref:`policy <minio-policy>`.
+All other policies should ``deny`` access to the target bucket.
+
 .. toctree::
    :titlesonly:
    :hidden:
--- a/source/administration/object-management.rst
+++ b/source/administration/object-management.rst
@@ -231,19 +231,13 @@ Expiration rules can result in removal of tiered data still in use by the source
 Tiering to an additional remote creates an additional network hop between the hot tier and it's data while also increasing operational complexity.
 
 You *may* configure object locking or versioning on the remote bucket.
-Use caution, as mismatched rules between the source and the target buckets run a risk of data corruption and loss.
 
-Creating separate rules on the target bucket may have effects such as the following:
+Enabling versioning or object locking on the target bucket may have effects such as the following:
 
 - Object locking set on the target bucket may prevent desired ``delete`` operations from the source bucket from completing.
-
-- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is not required.
-
-  If you enable versioning on the target bucket, you may experience the following unexpected or undesired results:
-  
-  - Excess data usage on the target remote after a delete operation on the source
-  - Reduced storage efficiency on the target, as ``delete`` operations result in ``DeleteMarker`` rather than freeing space
-  - Duplicate delete markers on source and target buckets
+- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is redundant at best.
+- Reduced storage efficiency on the target, as ``delete`` operations result in creation of a ``DeleteMarker`` rather than freeing space.
+- Duplicate delete markers on source and target buckets.
 
 Exclusive Access to Remote Data
 -------------------------------
--- a/source/administration/object-management.rst
+++ b/source/administration/object-management.rst
@@ -223,22 +223,27 @@ See :ref:`minio-lifecycle-management` for more complete documentation.
 Target Bucket Considerations
 ----------------------------
 
-- MinIO does *not* require that the target bucket match object management or versioning configurations with the source bucket.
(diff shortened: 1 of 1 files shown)


source/administration/object-management.rst
```diff
@@ -220,6 +220,33 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
 
+Target Bucket Configuration for Tiering
+---------------------------------------
+
+Keep in mind the following considerations when configuring the target bucket for tiering:
+
+- The target bucket *can* have its own set of object management rules different from the source.
+  For example, the source bucket may have object locking defined while the target bucket does not.
+- While enabling tiering requires :ref:`versioning <minio-bucket-versioning>` on the source bucket, the target bucket does *not* require versioning.
+  Enabling versioning on both the source and target buckets may lead to unexpected results.
```
feorlen, line 231: Is there anything useful to say about _what sort_ of unexpected results?

source/administration/object-management.rst
```diff
@@ -220,6 +220,33 @@ Conversion to or from XML may be required for importing rules created on S3 or s
 
 See :ref:`minio-lifecycle-management` for more complete documentation.
 
+Target Bucket Configuration for Tiering
+---------------------------------------
+
+Keep in mind the following considerations when configuring the target bucket for tiering:
+
+- The target bucket *can* have its own set of object management rules different from the source.
```
feorlen, line 228: maybe something like
```suggestion
- The target bucket *can* have its own set of object management rules, tiering does not require the source and target buckets have the same rules.
```

feorlen (approved):
A couple ideas for your consideration. Otherwise LGTM.

source/administration/object-management.rst
```diff
@@ -220,6 +220,33 @@ Conversion to or from XML may be required for importing rules created on S3 or s
 
 See :ref:`minio-lifecycle-management` for more complete documentation.
 
+Target Bucket Configuration for Tiering
+---------------------------------------
+
+Keep in mind the following considerations when configuring the target bucket for tiering:
```
ravindk89, line 226: I would break this up into two H2s or defnlists. Or just drop this intro and go straight into

- The remote target does *not* require matching object management or versioning configurations to the source

- If you set object locking rules on the remote, these can prevent delete operations on the source from completing
- If you set object expiration rules on the remote, these may result in data loss
- If you enable versioning on the remote, this may result in excess data remaining after delete operations on the source

Something like that.

ravindk89: Mostly so that it's clearer what the cause-effect is to any given config.

source/administration/object-management.rst
ravindk89: Mostly that we already UUID on the remote, so versioning makes no sense.

instead when you delete a tiered object on the source, you get a deletemarker on the remote, so your storage efficiency drops.

At worst you end up with delete markers in both places. ILM eventually cleans up the old metadata on the source, but the remote is still versioned and w/o ILM it just sticks around.

The worst case is ILM on the remote that is not on the source, and then that tears apart your data.

ravindk89 (changes requested):
I think we should restructure the intro a bit. The good info is here, but I think we need to improve how we present it to make it absolutely clear what the consequences are.

source/administration/object-management.rst
djwfyi: Added Ravind's comments to a completely reworked section.

source/administration/object-management.rst
djwfyi: Incorporated this and the comment above on results of target versioning into a reworked section.

source/administration/object-management.rst
```diff
@@ -220,6 +220,43 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
 
+Target Bucket Considerations
+----------------------------
+
+- MinIO does *not* require that the target bucket match object management or versioning configurations with the source bucket.
+- Object locking set on the target bucket may prevent desired ``delete`` operations from the source bucket from completing.
+- If the target bucket has expiration rules configured, these may result in data loss.
+- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is not required.
+
+  Versioning on the target bucket may have the following unexpected or undesired results:
```
ravindk89, line 231:
```suggestion
Versioning on the target bucket may have the following unexpected or undesired results:
```

source/administration/object-management.rst
```diff
@@ -220,6 +220,43 @@ Conversion to or from XML may be required for importing rules created on S3 or s
 
 See :ref:`minio-lifecycle-management` for more complete documentation.
 
+Target Bucket Considerations
+----------------------------
+
+- MinIO does *not* require that the target bucket match object management or versioning configurations with the source bucket.
```
ravindk89, line 226:
```suggestion
MinIO does *not* require that the target bucket match object management or versioning configurations with the source bucket.

Configuring versioning, lifecycle management, or object locking on the remote may have effects such as the following:
```

source/administration/object-management.rst
ravindk89: Or crunch this section down into

- Versioning on the target bucket may result in reduced storage efficiency due to the creation of ``DeleteMarkers`` in response to delete operations on the source

source/administration/object-management.rst
```diff
@@ -220,6 +220,43 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
+- If the target bucket has expiration rules configured, these may result in data loss.
+- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is not required.
+
+  Versioning on the target bucket may have the following unexpected or undesired results:
+  
+  - Excess data usage on the target remote after a delete operation on the source
+  - Reduced storage efficiency on the target, as ``delete`` operations result in ``DeleteMarker``
+  - Duplicate delete markers on source and target buckets
+
+- The target bucket *can* have its own set of object management rules.
```
ravindk89, line 237:
```suggestion
```

source/administration/object-management.rst
```diff
@@ -220,6 +220,43 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
+
+  Target buckets should *not* have their own rules for expiration or additional tiering.
```
ravindk89, line 239: I'd move this up to be in line with 226.

So something like

The remote bucket should not have any ILM expiration or tiering configurations.
Expiration rules can result in removal of tiered data still in use by the source.
Tiering to an additional remote creates an additional network hop between the hot tier and it's data while also increasing operational complexity.

You *may* configure object locking or versioning on the remote bucket. However, doing so has the following consequences:

- The bullet points.

ravindk89 (commented):
Hm - making one more pass at structured guidance. Mostly moving a few things around.

source/administration/object-management.rst
djwfyi: Taking the second suggestion and simplifying it down to just that.

source/administration/object-management.rst
djwfyi: Okay, so trying a different structure to the section entirely.

source/administration/object-management.rst
```diff
@@ -220,6 +220,48 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
+
+MinIO does *not* require that the target bucket match object management or versioning configurations with the source bucket.
+The target bucket *may* have its own set of object management rules, if defined with care.
+
+Target buckets should *not* have their own rules for expiration or additional tiering.
+Expiration rules can result in removal of tiered data still in use by the source bucket.
+Tiering to an additional remote creates an additional network hop between the hot tier and it's data while also increasing operational complexity.
+
+You *may* configure object locking or versioning on the remote bucket.
+Use caution, as mismatched rules between the source and the target buckets run a risk of data corruption and loss.
```
ravindk89, line 234: :thinking: well not mismatched because the source *can* have either, but the remote does *not* require it. Spitballing

```suggestion
Enabling versioning or object locking on the remote is redundant at best and may result in unexpected behavior in response to delete operations.

File: source/administration/object-management.rst [hunk 9]
```diff
@@ -220,6 +220,48 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
+
+Creating separate rules on the target bucket may have effects such as the following:
```
ravindk89, line 236:
```suggestion
Enabling versioning or object locking on the target bucket may have effects such as the following:
```

source/administration/object-management.rst
ravindk89: Or

```suggestion
```

and go right into line 236

source/administration/object-management.rst
```diff
@@ -220,6 +220,48 @@ Conversion to or from XML may be required for importing rules created on S3 or s
...
+
+- Object locking set on the target bucket may prevent desired ``delete`` operations from the source bucket from completing.
+
+- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is not required.
+
+  If you enable versioning on the target bucket, you may experience the following unexpected or undesired results:
+  
+  - Excess data usage on the target remote after a delete operation on the source
+  - Reduced storage efficiency on the target, as ``delete`` operations result in ``DeleteMarker`` rather than freeing space
+  - Duplicate delete markers on source and target buckets
```
ravindk89, line 246:
```suggestion
- MinIO tiers objects with their own ``UUID``, so versioning on the target bucket is redundant at best.
- Reduced storage efficiency on the target, as ``delete`` operations result in creation of a ``DeleteMarker`` rather than freeing space
- Duplicate delete markers on source and target buckets
```

Might as well make these inline - object locking implies versioning anyways, so these are all true.

ravindk89 (approved):
LGTM mod changeups

source/administration/object-management.rst
djwfyi: I'll just cut that line entirely. 👍

source/administration/object-management.rst
djwfyi: 👍

source/administration/object-management.rst
djwfyi: 👍

PR #1231, opened 2024-05-30
Change scrape_interval in metrics scraping example

feorlen:
The examples show `15s`, which is too short for a typical deployment. `60s` is recommended. Also note that this value may need to be increased if there are too many metrics to collect in the specified amount of time. Each scraping operation should complete before the next begins.

Staged
http://192.241.195.202:9000/staging/scrape_interval/linux/operations/monitoring/collect-minio-metrics-using-prometheus.html#generate-the-scrape-configuration

### DIFF
--- a/source/operations/monitoring/collect-minio-metrics-using-prometheus.rst
+++ b/source/operations/monitoring/collect-minio-metrics-using-prometheus.rst
@@ -59,7 +59,7 @@ Use the :mc-cmd:`mc admin prometheus generate` command to generate the scrape co
          :class: copyable
       
          global:
-            scrape_interval: 15s
+            scrape_interval: 60s
          
          scrape_configs:
             - job_name: minio-job
@@ -84,7 +84,7 @@ Use the :mc-cmd:`mc admin prometheus generate` command to generate the scrape co
          :class: copyable
       
          global:
-            scrape_interval: 15s
+            scrape_interval: 60s
          
          scrape_configs:
             - job_name: minio-job-node
@@ -109,7 +109,7 @@ Use the :mc-cmd:`mc admin prometheus generate` command to generate the scrape co
          :class: copyable
       
          global:
-            scrape_interval: 15s
+            scrape_interval: 60s
          
          scrape_configs:
             - job_name: minio-job-bucket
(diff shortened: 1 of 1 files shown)


source/operations/monitoring/collect-minio-metrics-using-prometheus.rst
```diff
@@ -146,6 +146,9 @@ Use the :mc-cmd:`mc admin prometheus generate` command to generate the scrape co
               static_configs:
               - targets: [minio.example.net]
       
+- Set the ``scrape_interval`` to an appropriate duration for your environment.
+  If you are scraping a large number of metrics, you may need to increase the ``scrape_interval`` to ensure the scraping operation completes before the next time period.
+
```
feorlen, line 151: New text to explain why you might want to change the `scrape_interval`

ravindk89:
I think that was intended to show where in the file it should go, so maybe we can just drop those two lines entirely. We state it goes into the Prometheus configuration already.

source/operations/monitoring/collect-minio-metrics-using-prometheus.rst
ravindk89: Invites the question of what a 'large number' is. Hm. Spitballing

```suggestion
Increase the ``scrape_interval`` value to exceed the time required to complete a single scrape operation.
Some deployments may require higher ``scrape_intervals`` due to the number of metrics being scraped.
```

ravindk89 (approved):
LGTM mod nit

source/operations/monitoring/collect-minio-metrics-using-prometheus.rst
feorlen: @mjkota also provided some additional context (that GitHub ate somehow).

Considerations:
•	System Load: A 60-second interval significantly reduces the load on your MinIO server compared to shorter intervals.
•	Metric Granularity: It is sufficient for cluster and bucket-level metrics.
•	Monitoring Needs: Adjust the interval based on specific monitoring requirements and the capability of your Prometheus server.
By following this recommendation, you ensure efficient resource usage and maintain effective monitoring of your MinIO infrastructure.
This configuration and guideline will help you set up Prometheus to monitor MinIO with a focus on performance efficiency.

feorlen: Rewrote, incorporating the above information

PR #1219, opened 2024-05-16
attempt to remove plugin from docs

feorlen:
The Kubernetes plugin is gone, this PR replaces the procedures that use `kubectl minio` in all its various forms. The plugin was referenced on many pages and for many purposes so there is _a lot_ of restructuring involved.

Some procedures no longer have CLI instructions, which can be addressed in subsequent PRs. Everything should have at least one working method, even if it's to use Operator Console.

- Remove references to plugin, except for pre-4.5.8 upgrade paths
- Move pre-4.5.8 upgrade paths to new child page (currently hidden from TOC, linked in page)
- Fill in with new Kustomize, kubectl, and/or Operator Console steps.

A handful of old screen captures still to be updated

Staged:
- [Operator deploy](http://192.241.195.202:9000/staging/DOCS-1213-upstream/k8s/operations/installation.html)
- [Operator upgrade](http://192.241.195.202:9000/staging/DOCS-1213-upstream/k8s/operations/install-deploy-manage/upgrade-minio-operator.html)
- [Deploy and manage Tenants](http://192.241.195.202:9000/staging/DOCS-1213-upstream/k8s/operations/deploy-manage-tenants.html)

Fixes https://github.com/minio/docs/issues/1213

PR #1238, opened 2024-06-07
Fix build warnings for Operator pages

feorlen:
Fix build warnings from https://github.com/minio/docs/pull/1219 

Some are broken refs that need to point elsewhere.

Some are because `openshift`, `eks`, `gke`, `aks` can't see a target that's on a `k8s` page, apparently due to how `operations/installation.rst` is structured. These we are still investigating.


ravindk89 (approved):
I'll address the weird linking issues in a separate bit of work. Intersphinx really should auto-resolve those but I'm not surprised it's having issues.

feorlen:
Merging the partial fix while we figure out what intersphinx's problem is

PR #1239, opened 2024-06-07
fix additional build warnings

feorlen:
Fix more build warnings. This includes removing references to `MINIO_KMS_KES_ENCLAVE`


feorlen:
The only ones I have remaining are `sphinxcontrib.autoyaml` parallel reading (all platforms) and `could not lex` on linux.

ravindk89:
Those two are undefeated in our time and probably will be unless someone (not it) fixes up sphinxcontrib.autoyaml, or whatever pyment is upset with on the lexing

ravindk89 (approved):

PR #1244, opened 2024-06-11
MinIO removed the version limit in a prior version

djwfyi:
This removes references to version limits and suggests using object expiration rules to manage unneeded noncurrent versions.

There is no doc issue to track this.

Staged:
- [Envvar](http://192.241.195.202:9000/staging/version-max/linux/reference/minio-server/settings/core.html#maximum-object-versions)
- [Object versioning](http://192.241.195.202:9000/staging/version-max/linux/administration/object-management/object-versioning.html)
- [mc mb](http://192.241.195.202:9000/staging/version-max/linux/reference/minio-mc/mc-mb.html#mc.mb.-with-versioning)

And a few other files.

### DIFF
--- a/source/administration/object-management/object-versioning.rst
+++ b/source/administration/object-management/object-versioning.rst
@@ -29,12 +29,7 @@ For versioned buckets, a write operation that mutates an object results in a new
 MinIO marks the "latest" version of the object that clients retrieve by default. 
 Clients can then explicitly choose to list, retrieve, or remove a specific object version. 
 
-.. versionchanged:: 2023-08-04T17-40-21Z
-
-   MinIO restricts object versioning to no more than 10,000 versions of each object.
-
-   If a write operation would exceed the 10,000 object version limit, MinIO blocks the operation and returns an error.
-   :ref:`Delete one or more <minio-bucket-versioning-delete>` versions to create a new version of the object.
+Define :ref:`object expiration <minio-lifecycle-management-create-expiry-rule>` rules to remove versions of objects no longer needed, such as by the number of versions or the date of versions.
 
 Read Operations on Versioned Objects
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/reference/minio-server/settings/core.rst
+++ b/source/reference/minio-server/settings/core.rst
@@ -371,12 +371,7 @@ Maximum Object Versions
 
 *Optional*
 
-Overrides the default maximum version per object limit of ``10000`` with the user specified value.
-
-.. important::
-
-   The default limit of 10,000 provides a safety valve against incorrect or inefficient application behavior in versioned buckets.
-   Lifting this limit without first ensuring your applications are designed for versioned operations may result in a negative performance impact over time.
-
-
-   
+Defines the default maximum versions to allow per object.
+If not defined, MinIO allows a practically unlimited number of object versions.
+ 
+The real maximum is up to highest value of ``MaxInt64``, which is over 9.2 quintillion versions per object.
--- a/source/reference/minio-server/settings/core.rst
+++ b/source/reference/minio-server/settings/core.rst
@@ -383,3 +383,5 @@ By default, MinIO allows up to the maximum value of an Int64 versions per object
 Arbitrarily high versions per objects may cause performance degradation on some operations, such as ``LIST``.
 This is especially true on systems running budget hardware or spinning drives (HDD).
 Applications or workloads which produce thousands or more versions per object may require design or architecture review to mitigate potential performance degradations.
+
+Setting a limit of no more than ``100`` should provide enough versions for most typical use cases.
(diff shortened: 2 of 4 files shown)


source/reference/minio-server/settings/core.rst
```diff
@@ -371,12 +371,7 @@ Maximum Object Versions
...
-
-.. important::
-
-   The default limit of 10,000 provides a safety valve against incorrect or inefficient application behavior in versioned buckets.
-   Lifting this limit without first ensuring your applications are designed for versioned operations may result in a negative performance impact over time.
-
-
-   
+Defines the default maximum versions to allow per object.
+If not defined, MinIO allows a practically unlimited number of object versions.
```
feorlen, line 375: The other location for this uses "virtually," might be nice to use the same word in both locations.

Although personally I'd lean towards "nearly." Your call.

source/reference/minio-server/settings/core.rst
```diff
@@ -371,12 +371,7 @@ Maximum Object Versions
...
+ 
+The real maximum is up to highest value of ``MaxInt64``, which is over 9.2 quintillion versions per object.
```
feorlen, line 377: So there aren't performance considerations besides disk usage? We do say elsewhere that lots of versions take up lots of space, so this should not be surprising.

source/reference/minio-server/settings/core.rst
djwfyi: Yeah, I like `nearly`.
I also made two other sentences parallel.

feorlen (approved):
LGTM
Oops forgot to click the clicky thing

source/reference/minio-server/settings/core.rst
harshavardhana: we should advice that this is to avoid any breaking change upon upgrade, but you should consider setting this value to 100 or something as there is no practical reason to go beyond 100 versions per object.

source/administration/object-management/object-versioning.rst
```diff
@@ -29,12 +29,7 @@ For versioned buckets, a write operation that mutates an object results in a new
 MinIO marks the "latest" version of the object that clients retrieve by default. 
 Clients can then explicitly choose to list, retrieve, or remove a specific object version. 
 
-.. versionchanged:: 2023-08-04T17-40-21Z
-
-   MinIO restricts object versioning to no more than 10,000 versions of each object.
-
-   If a write operation would exceed the 10,000 object version limit, MinIO blocks the operation and returns an error.
-   :ref:`Delete one or more <minio-bucket-versioning-delete>` versions to create a new version of the object.
+Define :ref:`object expiration <minio-lifecycle-management-create-expiry-rule>` rules to remove versions of objects no longer needed, such as by the number of versions or the date of versions.
```
ravindk89, line 32: What version of MinIO did we drop the max object version in? We should if possible mention what period of time it was enforced.

source/reference/minio-server/settings/core.rst
```diff
@@ -371,12 +371,7 @@ Maximum Object Versions
...
-
-   The default limit of 10,000 provides a safety valve against incorrect or inefficient application behavior in versioned buckets.
-   Lifting this limit without first ensuring your applications are designed for versioned operations may result in a negative performance impact over time.
-
-
-   
+Defines the default maximum versions to allow per object.
+If not defined, MinIO keeps a nearly unlimited number of object versions.
+ 
+The real maximum is up to highest value of ``MaxInt64``, which is over 9.2 quintillion versions per object.
```
ravindk89, line 377:
```suggestion
MinIO default version limit is the maximum value of an Int64, or 9.2 quintillion versions per object.

Most application workloads should not require more than 100 versions per object.
In practice, workloads that produce thousands or more versions per object should initiate review of the access pattern to remove or reduce that behavior.

Arbitrarily high versions per objects may cause performance degradation on certain operations such as LIST, especially on systems running budget hardware or spinning drives (HDD).

ravindk89: Spitballing.

ravindk89 (changes requested):

source/administration/object-management/object-versioning.rst
harshavardhana: We don't have to, because enforcement was incorrect and it must have opt-in - like the way it is now.

source/reference/minio-server/settings/core.rst
djwfyi: Somewhat tweaked, but basically taking your suggestion.

source/administration/object-management/object-versioning.rst
djwfyi: @ravindk89 It was removed in [RELEASE.2024-03-30T09-41-56Z](https://github.com/minio/minio/releases/tag/RELEASE.2024-03-30T09-41-56Z). 
PR [19360](https://github.com/minio/minio/pull/19360).

djwfyi: I added a note to the max version setting.

source/reference/minio-server/settings/core.rst
djwfyi: Added a line about considering 100 versions as a limit.

PR #1246, opened 2024-06-12
Can't resize server pools

feorlen:
Explicitly note in the concepts page that you can't change the size of a server pool after it's created.

Not staged

### DIFF
--- a/source/operations/concepts.rst
+++ b/source/operations/concepts.rst
@@ -107,7 +107,7 @@ MinIO :ref:`distributed deployments <minio-mnmd>` support expansion and decommis
 
 Expansion consists of adding one or more :ref:`server pools <minio-intro-server-pool>` to an existing deployment.
 Each server pool consists of dedicated nodes and storage that contribute to the overall capacity of the deployment.
-The size of an individual server pool is fixed at the time it is created.
+Once you create a server pool you cannot change its size, but you can add or remove capacity at any time by adding or decommissioning pools.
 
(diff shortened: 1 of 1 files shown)


source/operations/concepts.rst
```diff
@@ -107,6 +107,7 @@ MinIO :ref:`distributed deployments <minio-mnmd>` support expansion and decommis
 
 Expansion consists of adding one or more :ref:`server pools <minio-intro-server-pool>` to an existing deployment.
 Each server pool consists of dedicated nodes and storage that contribute to the overall capacity of the deployment.
+The size of an individual server pool is fixed at the time it is created.
```
feorlen, line 110: Now that I look at this again, could be reworded.

djwfyi: Just trying to get it out of passive voice.

```suggestion
Once set, a server pool's size cannot change.
```

djwfyi (changes requested):
A wording suggestion to simplify and remove the passive voices.

source/operations/concepts.rst
feorlen: I'm very confused at what GitHub just did, your suggestion is referencing the previous version of the text.

djwfyi (approved):
LGTM!

source/operations/concepts.rst
djwfyi: Odd, very odd

PR #1251, opened 2024-06-20
Add `MINIO_SERVER_URL` to deprecated command list

ravindk89:
For reference by support w/ customers running older MinIO versions

Mostly need this because it sometimes comes up via support, and this lets us reference it while also pushing users to upgrade to versions where this kind of behavior is handled automatically.

### DIFF
--- a/source/reference/minio-server/settings/deprecated.rst
+++ b/source/reference/minio-server/settings/deprecated.rst
@@ -81,6 +81,8 @@ They are listed here for historical reference only.
 
    This setting may be required if:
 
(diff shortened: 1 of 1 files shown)


djwfyi (changes requested):
Not sure if we need the `or` at all, since it's a bulleted list. But should make those parallel.

ravindk89:
> Not sure if we need the `or` at all, since it's a bulleted list. But should make those parallel.

I'll take the changes, this was copy-pasted from the commit prior to removal anyways.

PR #1272, opened 2024-07-17
Add Docsearch facet attribute for search feature

rushenn:
We're exploring the possibility of adding a global search feature (including docs, blog, etc.) on the main min.io website. As part of this, we are adding a higher-level facet to filter results based on the given site. This is for experimental purposes. If it doesn't work as expected, we will send a PR in the future to remove this meta attribute.

Part of https://github.com/miniohq/min.io/issues/2136

### DIFF
--- a/source/_templates/head.html
+++ b/source/_templates/head.html
@@ -1,4 +1,5 @@
 <!-- Algolia meta attrs -->
+<meta name="docsearch:site" content="docs"/>
 <meta name="docsearch:platform" content="{{ doc_platform }}"/>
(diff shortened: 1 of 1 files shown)


kaankabalak (approved):

ravindk89 (approved):

ravindk89:
As a note @rushenn we're actively working on reducing the total number of platforms we're documenting from the current count (something like 8) down to 2 (Baremetal or Kubernetes), and we're hoping to get approval to drop back to a single site (min.io/docs/minio/) <- I'm not sure how this would affect faceting, but it might simplify our overall collection of algolia indexes

PR #1294, opened 2024-08-07
deprecate mc quota

feorlen:
Bucket quotas don't work the way people think they work, as they cannot enforce hard limits. Deprecate in docs while we sort out what to do with the commands. 

The future replacement is MinIO Enterprise Catalog.

Staged:
http://192.241.195.202:9000/staging/deprecate-bucket-quota/linux/reference/minio-mc-deprecated.html

See:
https://github.com/minio/mc/pull/5011
https://github.com/minio/mc/pull/5012


source/reference/minio-mc-deprecated.rst
```diff
@@ -51,6 +51,22 @@ Table of Deprecated Commands
      - :mc-cmd:`mc ilm rule rm`
      - mc RELEASE.2022-12-24T15-21-38Z
 
+   * - ``mc quota``
```
feorlen, line 54: Not entirely sure what to do about `mc quota` itself. There's important information there about how bucket quota works (and why it's not what people think it is). So I don't want to just drop its page.

Maybe put that information into the deprecated `mc quota set` page instead? 🤔

source/reference/deprecated/mc-quota.rst
```diff
@@ -0,0 +1,71 @@
...
+
+When a bucket with a quota configured reaches the specified limit, as determined by the MinIO object scanner, MinIO rejects further ``PUT`` requests for the bucket.
+
+Each time the MinIO :ref:`object scanner <minio-lifecycle-management-scanner>` scans a bucket for pending :ref:`object lifecycle transitions <minio-lifecycle-management>`, it also checks if the bucket has exceeded a configured quota.
+
+.. admonition:: Quota enforcement is not immediate
+   :class: note
+
+   Bucket quotas are not intended to enforce a strict hard limit on a bucket's size.
+   If a bucket exceeds its quota between scanner passes, MinIO continues to accept ``PUT`` requests for that bucket until _after_ the next scanner pass identifies the quota violation.
```
feorlen, line 34: New content here, which is hard to tell because GitHub thinks the page is "new"

source/reference/minio-mc-deprecated.rst
feorlen: Ok, did the above so there's no `mc quota` page under deprecated.

ravindk89 (approved):

PR #1306, opened 2024-08-23
Make port numbers consistent in the port assignment chapter

philipphenkel:


ravindk89 (approved):
I'm convinced we have gremlins that change these ports around randomly because each time we think we got all of them one more pops up.

Thanks!

PR #1346, opened 2024-10-08
Adds language about having write permission to use `mc admin update`

### LINKED ISSUE #1301: Clarify permissions required for `mc admin update`
On the [`mc admin update`](https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-update.html) reference page, this paragraph says the command may require sudo:
```
After running the command, a prompt displays to confirm the update.
Type ``y`` and ``[ENTER]`` to confirm and proceed with the update.
You may need to use ``sudo`` if your user does not have write permissions for the path where ``mc`` is installed.
```

`mc admin update` itself doesn't inherently require sudo. But the target location for the binary needs to be writable by the user executing the command. The user must be able to write to the path to stage a copy of the new binary before replacing the existing one. 

Presumably humans aren't logging in as the user that runs the `minio` process or owns the install location. Maybe make it writable for an appropriate group? We should clarify with engineering what makes sense to recommend here. 

`includes/linux/steps-upgrade-minio-deployment.rst` mentions the user running the `minio` process needs r/w on the binary, which isn't the same thing. Might be a good place to clarify, along with other text that references (or should reference) upgrading.

Related:
`includes/minio-server-installation.rst` mentions `mc admin update` (about source-based installations) but this include appears to not be used anywhere.

djwfyi:
Closes #1301

Staged:

- [mc admin update](http://192.241.195.202:9000/staging/admin-update-1301/linux/reference/minio-mc-admin/mc-admin-update.html)
- [linux upgrade](http://192.241.195.202:9000/staging/admin-update-1301/linux/operations/install-deploy-manage/upgrade-minio-deployment.html#update-using-mc-admin-update)

### DIFF
--- a/source/reference/minio-mc-admin/mc-admin-update.rst
+++ b/source/reference/minio-mc-admin/mc-admin-update.rst
@@ -48,7 +48,7 @@ Do not perform "rolling" (that is, one node at a time) upgrade procedures.
 Permissions
 ~~~~~~~~~~~
 
-The user running the command **must** have ``write`` permissions to the target path where MinIO Server binary installs.
+The user running the command **must** have ``write`` permissions to the target path where the MinIO Server binary installs.
 
 Examples
 --------
--- a/source/reference/minio-mc-admin/mc-admin-update.rst
+++ b/source/reference/minio-mc-admin/mc-admin-update.rst
@@ -22,7 +22,8 @@ The command also supports using a private mirror server for environments where t
 
 After running the command, a prompt displays to confirm the update.
 Type ``y`` and ``[ENTER]`` to confirm and proceed with the update.
-You may need to use ``sudo`` if your user does not have write permissions for the path where ``mc`` is installed.
+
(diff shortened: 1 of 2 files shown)


source/reference/minio-mc-admin/mc-admin-update.rst
```diff
@@ -44,6 +45,11 @@ MinIO operations are atomic and strictly consistent and as such the restart proc
 MinIO strongly recommends only performing simultaneous upgrade-and-restart procedures. 
 Do not perform "rolling" (that is, one node at a time) upgrade procedures.
 
+Permissions
+~~~~~~~~~~~
+
+The user running the command **must** have ``write`` permissions to the target path where MinIO Server binary installs.
```
feorlen, line 51: probably?
```suggestion
The user running the command **must** have ``write`` permissions to the target path where the MinIO Server binary installs.
```

feorlen (approved):
One thing for your consideration, otherwise LGTM

PR #1357, opened 2024-10-24
mention that tiny files are not compressed, also other typos

feorlen:
Came up in a customer issue. Specifically the threshold is 4 kb and there is no configuration option. (Nor does Klaus want one.)

Worth adding? With how much detail? The threshold choice is because of typical block size on disk. You will use at least that much anyway so why bother compressing.

Staged:

http://192.241.195.202:9000/staging/dont-compress-small-objects/linux/administration/object-management/data-compression.html#excluded-file-types

Drive-by fix tacked onto this PR: correct typos in some `mc anonymous` examples.

Staged:
http://192.241.195.202:9000/staging/dont-compress-small-objects/linux/reference/minio-mc/mc-anonymous.html

### DIFF
--- a/source/administration/object-management/data-compression.rst
+++ b/source/administration/object-management/data-compression.rst
@@ -48,7 +48,7 @@ You can control which objects are compressed by specifying the desired file exte
 Excluded File Types
 ~~~~~~~~~~~~~~~~~~~
 
-Some data cannot be effectively compressed, such as video or already compressed data.
+Some data cannot be effectively compressed, such as video, already compressed data, or very small files.
 MinIO does not compress common incompressible file types, even if they are specified in the compression configuration.
 
 Objects of these types are never compressed:
--- a/source/administration/object-management/data-compression.rst
+++ b/source/administration/object-management/data-compression.rst
@@ -48,7 +48,8 @@ You can control which objects are compressed by specifying the desired file exte
 Excluded File Types
 ~~~~~~~~~~~~~~~~~~~
 
-Some data cannot be effectively compressed, such as video, already compressed data, or very small files.
+Some data cannot be effectively compressed.
+For example: video, already compressed data, or very small files.
 MinIO does not compress common incompressible file types, even if they are specified in the compression configuration.
 
 Objects of these types are never compressed:
--- a/source/administration/object-management/data-compression.rst
+++ b/source/administration/object-management/data-compression.rst
@@ -49,7 +49,7 @@ Excluded File Types
 ~~~~~~~~~~~~~~~~~~~
 
 Some data cannot be effectively compressed.
-For example: video, already compressed data, or very small files.
+For example: video, already compressed data, or files less than 4KiB.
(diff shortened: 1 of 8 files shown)


ravindk89:
https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#data-compression makes sense to put here at the top if nowhere else.

feorlen:
> https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#data-compression makes sense to put here at the top if nowhere else.

🤔 to say that there _isn't_ a setting for this? Further down on that page `MINIO_COMPRESSION_EXTENSIONS` and `MINIO_COMPRESSION_MIME_TYPES` link to the same "Excluded File Types" section for more info.

ravindk89:
> > https://min.io/docs/minio/linux/reference/minio-server/settings/core.html#data-compression makes sense to put here at the top if nowhere else.
> 
> 🤔 to say that there _isn't_ a setting for this? Further down on that page `MINIO_COMPRESSION_EXTENSIONS` and `MINIO_COMPRESSION_MIME_TYPES` link to the same "Excluded File Types" section for more info.

The excluded file types section is good actually -  enough to state 'files under 4KiB' and call it done.

source/administration/object-management/data-compression.rst
```diff
@@ -48,7 +48,7 @@ You can control which objects are compressed by specifying the desired file exte
 Excluded File Types
 ~~~~~~~~~~~~~~~~~~~
 
-Some data cannot be effectively compressed, such as video or already compressed data.
+Some data cannot be effectively compressed, such as video, already compressed data, or very small files.
```
djwfyi, line 51:
```suggestion
Some data cannot be effectively compressed.
For example: video, already compressed data, or very small files.
```

A suggestion to consider to avoid a dizzying amount of commas.

djwfyi (approved):
One line split to consider, if you like. Otherwise LGTM.

source/administration/object-management/data-compression.rst
feorlen: oh yes, that is nicer 👍🏻

source/administration/object-management/data-compression.rst
```diff
@@ -48,7 +48,8 @@ You can control which objects are compressed by specifying the desired file exte
 Excluded File Types
 ~~~~~~~~~~~~~~~~~~~
 
-Some data cannot be effectively compressed, such as video or already compressed data.
+Some data cannot be effectively compressed.
+For example: video, already compressed data, or very small files.
```
ravindk89, line 52:
```suggestion
For example: video, already compressed data, or files less than 4KiB.
```

ravindk89 (approved):

source/administration/object-management/data-compression.rst
feorlen: 👍🏻 will add manually to reduce merge nonsense

PR #1364, opened 2024-11-09
experimenting with redirects

ravindk89:
@dvaldivia attempt to create nginx rewrites in a block.

Not sure about syntax since there are so many.

Goal is to support sometihng like:

```
server {
  listen 80;
  server_name example.com;
  […]
 location /docs/minio/ {
     include conf/location.conf;
 }
}
```

THis is assuming some things about the location block - I'd need @harshavardhana or something w/ access to the min.io webserver to confirm the current server + location config for the docs blocks before finalizing this.


harshavardhana:
We can not do these redirects in our min.io ?

ravindk89:
We have to do it on min.io, but for maintenance it would be good to keep
them in repo and periodically sync.

Pages change sometimes and I think this is easier than pulling you for
access when it happens.

The above also lets us test redirects in staging :)

ravindk89:
Ping @harshavardhana @dvaldivia @donatello 

Goal here is to have an in-code way of managing nginx redirects that just get pulled in by min.io server.

I have some ideas on how to validate in staging environment but its hard to 1:1 without also seeing our nginx location/server blocks. We can handle that offline as necessary

PR #1370, opened 2024-11-17
Fix console port number

gu1ll0me:


djwfyi (approved):
Thanks for catching these!
We had a blip there where we were thinking about 9090 for the console port, but that didn't pan out. Thought we had caught them all, but clearly not.

PR #1393, opened 2025-01-13
Adds info that `MC_HOST` envvar uses s3v4 api

### LINKED ISSUE #1379: [BUG] missing doc how to set api version and path via env vars
**Describe the bug**

the docs give no info on how to set an host alias via env var but with `--path="off" and api="S3v4"`. if it is not possible it should also say so (where to open a feature request for requesting this functionalilty?).

it could work like this (tried it but it does not work): 
```
export MC_HOST_myalias="https://access_key:secret_key@mystorage.com?path=off&api=S3v4
```

djwfyi:
Closes #1379

Not staged.

### DIFF
--- a/source/reference/minio-mc/minio-client-settings.rst
+++ b/source/reference/minio-mc/minio-client-settings.rst
@@ -25,7 +25,7 @@ Host Credentials
 Use this setting to add a temporary alias to use for `mc` commands.
 For example, for use with scripting.
 
-The temporary alias uses the ``s3v4`` API.
+The temporary alias uses the :aws-docs:`AWS s3v4 API <AmazonS3/latest/API/sig-v4-authenticating-requests.html>`.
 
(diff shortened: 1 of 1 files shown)


ravindk89 (approved):

source/reference/minio-mc/minio-client-settings.rst
```diff
@@ -25,6 +25,8 @@ Host Credentials
 Use this setting to add a temporary alias to use for `mc` commands.
 For example, for use with scripting.
 
+The temporary alias uses the ``s3v4`` API.
```
feorlen, line 28: I don't know what `s3v4` means, is there something useful to link to? Is this our API or one from AWS?

harshavardhana: It's more of a signature than API

PR #1406, opened 2025-02-05
Fixes command for upgrading a MinIO Tenant

### LINKED ISSUE #1403: [BUG] Incorrect kubectl command on upgrading a tenant
**Describe the bug**

This page: https://min.io/docs/minio/kubernetes/upstream/operations/install-deploy-manage/upgrade-minio-tenant.html

Has a step where we run `kubectl apply -f ./` on a kustomize directory.

The kubectl apply command with `-f` is for a file, however we are running it on a directory in which we just dropped a new `kustomization.yaml` file, so the correct command is `kubectl apply -k ./*` 

Reported and tested by a customer.

I'm not sure if the trailing `*` is necessary, I suspect it is not, however that's what was successfully run by the customer

djwfyi:
Closes #1403

### DIFF
--- a/source/operations/install-deploy-manage/upgrade-minio-tenant.rst
+++ b/source/operations/install-deploy-manage/upgrade-minio-tenant.rst
@@ -123,7 +123,7 @@ Refer to the :kube-docs:`Kustomize Documentation <tasks/manage-kubernetes-object
   .. code-block:: shell
      :class: copyable
 
-     kubectl apply -f ./
+     kubectl apply -k ./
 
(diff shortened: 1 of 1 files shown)


djwfyi:
Internet agreed with Mark's suggestion in the issue that a `*` is not required for the command.

I also did a pass to check for other occurrences, but other `kubectl apply` commands are all for individual files.

Bradamant3 (approved):

ravindk89 (approved):

PR #1415, opened 2025-02-11
Removing references to openshift container platform, bumping CRD

ravindk89:
# Summary

We removed support for deploying via OpenShift operatorhub / marketplace a few months back. The burden of maintaining the certifications/process required was sufficient to pull that feature set out moving forward and migrate it to AIStor as part of a more complete enterprise offering.

For now we need to remove references to the operator hub / marketplace. It is possible that the Openshift Kubernetes Engine may work w/ our stock kubernetes installation paths, but we do not have resources to test.

So for the immediate term, we're just going to pull whats outdated.

Also throwing in a fixup to the CRDs - I ran into a weird issue where I had to `pip3 install asciidoc` before the sync-operator-crd-docs job worked, am looking into that separately.

Staged: http://192.241.195.202:9000/staging/openshift-fixup/openshift/index.html


source/includes/k8s/operator-values.yaml
```diff
@@ -32,14 +28,14 @@ operator:
   ###
   # Specify the Operator container image to use for the deployment.
   # ``image.tag``
-  # For example, the following sets the image to the ``quay.io/minio/operator`` repo and the v6.0.3 tag.
+  # For example, the following sets the image to the ``quay.io/minio/operator`` repo and the v7.0.0 tag.
```
feorlen, line 31: I guess our automation can't directly replace a version number in files that are included this way?

ravindk89: We could potentially grep for it, but https://github.com/minio/operator/blob/master/helm/operator/values.yaml ultimately dictates the content. Ideally we would only update it there so we have minimal skew.

feorlen (approved):
LGTM

PR #1417, opened 2025-02-13
correct parent flag for mkdir in installation content

Bradamant3:
Flag is lower case, not upper case.

Caught while testing

Not (manually) staged.


djwfyi (approved):
:shipit: LGTM

PR #1422, opened 2025-02-19
noncurrent-trasition-days can accept 0

djwfyi:
If the target transition tier is another MinIO deployment, the days value can be `0`.

Not staged.


feorlen (approved):
LGTM

PR #1433, opened 2025-03-14
Missing part of command in Object Versioning documentation

Sinscerly:
Now there is standing: `mc version ALIAS/BUCKET` but it should be `mc version enable ALIAS/BUCKET`

the `enable` part in the command example is missing


djwfyi (approved):
Thanks for the find and the fix!

PR #1437, opened 2025-03-19
docs: fixed the format for --older-than parameter in mc-rm

moFahmi10:
The format for--older-than parameter in the docs was #d#hh#mm#ss which cannot be parsed when used, the correct format is #d#h#m#s


djwfyi (approved):
Thanks for the fix!

PR #1464, opened 2025-05-02
Fix UI issues

rushenn:
1. Toggle the docs menu conditionally to avoid showing empty white space when the menu is not present.

Screenshot 2025-05-02 at 10 41 07 AM

2. Fix padding and layout shift when pasting into the search input field.

Screenshot 2025-05-02 at 10 42 28 AM

### DIFF
--- a/source/_static/scss/includes/_search.scss
+++ b/source/_static/scss/includes/_search.scss
@@ -3,9 +3,9 @@
 		.header {
 			.search__dropdown {
 				position: absolute;
-				top: 100%;
-				left: 0;
-				width: 100%;
+				top: calc(100% - 1rem);
+				left: 0.1rem;
+				width: calc(100% - 0.2rem);
 			}
 	
 			.search--focused {
@@ -71,6 +71,11 @@
 	.search__reset {
 		display: grid;
 	}
+
+	#search-box {
+		position: relative;
+		z-index: 11;
+	}
 }
 
 .search__form {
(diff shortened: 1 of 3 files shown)


feorlen:
Thank you @rushenn ! Is there a way we can test these changes before they are merged?

djwfyi:
@feorlen 
> Thank you @rushenn ! Is there a way we can test these changes before they are merged?

You can check out the PR and then build and serve locally.

```
gh pr checkout 1464
```

Normal build thing.

```
python3 -m http.server --directory build/ui-fix/linux/html
```

(You may have to fiddle with the directory to serve from.)

djwfyi:
OH, uh ... I don't think we do have a way to turn off the menu. Duh.

ravindk89:
Well nothing breaks on build, so lets merge this and see what happens.

ravindk89 (approved):

djwfyi:
Tested locally and agree with Ravind.

rushenn:
@djwfyi 

> OH, uh ... I don't think we do have a way to turn off the menu. Duh.

Currently, the docs menu is populated based on the array defined in `default.conf.py`

```
html_context = {
   'doc_platform': platform.lower(),
   'docs': [
      # The first item has to be the current docs site #
      {
         'name': 'MinIO Server',
         'current': True
      },
      {
         'name': 'DirectPV',
         'url': 'https://min.io/docs/directpv',
         'external': True
      },
      {
         'name': 'KES',
         'url': 'https://min.io/docs/kes',
         'external': True
      }
   ]
}
```

The idea here is that if docs is empty or missing, the menu will not be displayed.

djwfyi:
Yeah, that's the section I commented out to test it. Thanks for the work!

PR #1470, opened 2025-05-28
Update TypeScript guidance

timglorioso:
The JavaScript Install instructions include a TypeScript section. I removed it entirely because types are now provided by `minio-js` and should not be installed separately via `@types/minio`.

Happy to consider alternatives, such as keeping the section with a note like "types are now provided by the `minio-js` package".

### DIFF
--- a/source/developers/minio-drivers.rst
+++ b/source/developers/minio-drivers.rst
@@ -165,13 +165,6 @@ Install
        npm install
        npm install -g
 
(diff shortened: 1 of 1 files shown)


ravindk89:
I'll actually defer the review here to one of our JS driver developers - @prakashsvmx can you confirm and approve?

prakashsvmx (approved):

Issue #1508, opened 2025-11-05
[BUG] Makefile doesn't account for GNU sed on Darwin


Airor4:
**Describe the bug**
I have gnu-sed installed view `brew install coreutils` which breaks the sed commands in the Makefile. GNU sed misinterprets the `-i` (in-place) argument since it doesn't expect the space, leading to the "No such file or directory" error.

```
--------------------------------------
         Building for MinIO
--------------------------------------
make[1]: Entering directory 'docs'
Synchronizing all external dependencies
make[2]: Entering directory 'docs'
Retrieving latest stable KES version
sed: can't read s|KESLATEST|2025-03-12T09-35-18Z|g: No such file or directory
make[2]: *** [Makefile:103: sync-kes-version] Error 2
```

**To Reproduce**
1. Use a darwin machine
2. `brew install coreutils`
3. add coreutils to your PATH
4. `make SYNC_SDK=true mindocs`

**Expected behavior**
`make SYNC_SDK=true mindocs` works regardless of sed installation

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS 26.0.1 25A362 arm64

**Smartphone (please complete the following information):**

**Additional context**
```
$ sed --version
sed (GNU sed) 4.9
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jay Fenlason, Tom Lord, Ken Pizzini,
Paolo Bonzini, Jim Meyering, and Assaf Gordon.

This sed program was built without SELinux support.

GNU sed home page: <https://www.gnu.org/software/sed/>.
General help using GNU software: <https://www.gnu.org/gethelp/>.
E-mail bug reports to: <bug-sed@gnu.org>.
```

ravindk89:
Hm - most of the team is on OSX so I'm surprised this is the first time this has come up. 

We do already have a `case` statement in `sync-kes-version` that has distinct syntax for `sed`, and that's worked for a while (at least up until now)

Just to check, if you `uname -s` what do you get back?

Airor4:
> Hm - most of the team is on OSX so I'm surprised this is the first time this has come up.
> 
> We do already have a `case` statement in `sync-kes-version` that has distinct syntax for `sed`, and that's worked for a while (at least up until now)
> 
[...]

PR #1509, opened 2025-11-05
decide which sed to use based on installed sed rather than os

Airor4:
#1508


ravindk89 (approved):

ravindk89:
@Airor4 thanks for taking the time! I recall us dealing with this a few years back but there's always another monkey with a wrench I suppose.

PR #1514, opened 2025-12-02
Fix SED_INPLACE for Darwin

jabenninghoff:
Fix building of docs on Darwin (macOS)


ravindk89:
Thanks for the fixup!

